PR #58, opened 2014-09-16
33513 pull replication by document id

tomblench:
Do pull replication with a subset of document IDs.

The user configures clientFilterDocIds on the CDTPullReplication object; this is an array of document IDs.

Internally, the sets clientFilterCurrentSetDocIds (doc IDs for inclusion during pull, for which we already have at least one rev in the database), and clientFilterNewDocIds (doc IDs for inclusion during pull, which we don't yet have in the database) are used.

Lots of things are missing right now, especially an updated remoteCheckpointDocID method whose input to the hash function should the doc ID list.

### DIFF
--- a/Classes/common/touchdb/TDPuller.m
+++ b/Classes/common/touchdb/TDPuller.m
@@ -29,6 +29,7 @@
 #import "ExceptionUtils.h"
 #import "TDJSON.h"
 #import "CDTLogging.h"
+#import "TDCanonicalJSON.h"
 
 // Maximum number of revisions to fetch simultaneously. (CFNetwork will only send about 5
 // simultaneous requests, but by keeping a larger number in its queue we ensure that it doesn't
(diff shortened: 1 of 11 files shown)


Classes/common/touchdb/TDPuller.m
```diff
@@ -347,35 +366,41 @@ - (void)pullRemoteRevision:(TD_Revision*)rev
...
+            } else {
+                TD_Revision* gotRev = [TD_Revision revisionWithProperties: dl.document];
+                gotRev.sequence = rev.sequence;
+                // Add to batcher ... eventually it will be fed to -insertRevisions:.
+                if (immediatelyInsert) {
+                    [strongSelf insertDownloads:@[gotRev]];
+                } else {
+                    [_downloadsToInsert queueObject: gotRev];
+                }
+                [strongSelf asyncTaskStarted];
```
mikerhodes: Sounds like this might not be required if you're using immediatelyInsert?

mikerhodes: Turned out immediatelyInsert decrements the asynctaskcount so we need to match here with an increment.
Though, it occurs to me we decrement asynctaskcount prior to incrementing it?

mikerhodes:
Suggest we close this one for now @tomblench ?

PR #99, opened 2014-12-11
42715 Removes the queue from the TDReplicator.

gadamc:
This was left over from NSThread code migration from the TDReplicatorManager to TDReplicator.


mikerhodes:
Looks better. +1 when squashed.

gadamc:
@mikerhodes With TDReplicator -queue now removed, a few of the TDReplicator methods that are called from external threads had to be modified. The guts of those methods were moved to new methods and then performSelector:@selector(newmethod) onThread:_replicatorThread ... was used to run that code on the _replicatorThread.  I just want to make sure you were okay with this before it was merged to master.

PR #100, opened 2014-12-12
42444 Updates the example app to use CDTReplicators properly.

gadamc:
Updates the example app to show best-practices
around using CDTReplicators. CDTReplicators must be retained
in order for their replciation to complete. In this example,
they are held in an array. The CDTTodoReplicator class
becomes the delegate for each of the CDTReplicator objects it
creates. It reports on the status of the replication via
the callbacks and releases the CDTReplicator objects when they
either 'complete' or 'error' (in both cases, the CDTReplicator
has stopped operations).

The polling block has been removed. Previously, this polling
block was keeping the strong reference to the CDTReplicator,
but is no longer needed and also not encouraged.

The CDTTodoReplicator now uses the new CDTPush/PullReplication
configuration and the CDTReplicatorFactor -oneWay:error method
to build CDTReplicator objects, as the CDTReplicatorFactory
-onewaySourceDatastore:targetURI and
-onewaySourceURI:targetDatastore are deprecated.

Additionally, the CDTAppDelegate now keeps a persistent
CDTTodoReplicator object, which the CDTViewController uses to
fire off synchronization. A long-lived object that manages
and maintains individual CDTReplicator objects is necessary.

### DIFF
--- a/Project/Project/CDTAppDelegate.h
+++ b/Project/Project/CDTAppDelegate.h
@@ -18,12 +18,14 @@
 @class CDTDatastore;
 @class CDTIndexManager;
 @class CDTReplicatorFactory;
+@class CDTTodoReplicator;
 
 @interface CDTAppDelegate : UIResponder <UIApplicationDelegate>
 
 @property (nonatomic, strong) CDTDatastore *datastore;
 @property (nonatomic, strong) CDTIndexManager *indexManager;
 @property (nonatomic, strong) CDTReplicatorFactory *replicatorFactory;
+@property (nonatomic, strong) CDTTodoReplicator *todoReplicator;
 
 @property (strong, nonatomic) UIWindow *window;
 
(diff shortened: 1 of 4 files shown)


Classes/common/CDTReplicator/CDTReplicator.m
```diff
@@ -100,6 +100,39 @@ - (id)initWithTDReplicatorManager:(TDReplicatorManager *)replicatorManager
...
+    }
+    
+    NSString *fullinfo = [NSString stringWithFormat: @"CDTReplicator %@, source: %@, target: %@ "
+                          @"filter name: %@, filter parameters %@, unique replication session "
+                          @"ID: %@", replicationType, source, target, self.tdReplicator.filterName,
+                          self.tdReplicator.filterParameters, self.tdReplicator.sessionID];
+    
+    return fullinfo;
+}
+
```
gadamc: I think all of this should be moved into https://github.com/cloudant/CDTDatastore/pull/104

mikerhodes:
I think this is fine now. Your proposed message in 384795c looks good to me.

PR #101, opened 2014-12-12
42711 multiple log messages

rhyshort:
- Bump Cocoalumberjack to version 2.0.0-rc
- Set up test logging inside dispatch_once block


mikerhodes:
I think this is fine for now, though I think we need to come back to this set of macros because we're got macros in macros in macros now, which is difficult to reason about. 

Merge 5242973 and 0d546bf into a single commit before merging to master.

PR #103, opened 2014-12-22
42812 crush when deleting db during push replication

gadamc:
(The FB ticket is 42884, with 42812 as the parent case.)


ReplicationAcceptance/ReplicationAcceptance/ReplicationAcceptance.m
```diff
@@ -48,7 +48,7 @@ @implementation ReplicationAcceptance
  as all these documents are read from both local and remote databases
  during the check phase.
  */
-static NSUInteger n_docs = 10000;
+static NSUInteger n_docs = 2000;
```
mikerhodes: Revert this line. Be careful what you commit.

ReplicationAcceptance/ReplicationAcceptance/ReplicationAcceptance.m
```diff
@@ -313,6 +331,154 @@ -(void) testPushErrorsWhenLocalDatabaseIsDeleted
...
+    STAssertEquals(delegateError.code, CDTReplicatorErrorLocalDatabaseDeleted,
+                   @"Wrong error code: %ld", delegateError.code);
+    
+}
+
+
+-(void) testSyncReplicationErrorsWhenLocalDatabaseDeleted_pullDelegateDeletes
+{
+    [self createLocalDocs:5000];
+    [self createRemoteDocs:5000 suffixFrom:5000];
```
mikerhodes: I think you need some notes here that you're creating 5k docs to try and guarantee that the deleted operation happens when you're still replicating. Could we have fewer docs?

mikerhodes:
I think that bar the line changing 2k for 10k test docs and the few notes you might want to add, this PR's basically good to be merged.

PR #105, opened 2015-01-06
42938 upgrade to xctest

rhyshort:
This PR updates all tests for CDTDatastore (Both replication acceptance and Unit tests) to use XCTest instead of SenTest. It also changes the schemes from Tests for both OS X and iOS to Tests for just iOS and Tests OS X for OS X. This also updates the RakeFile to allow OS X tests to be run on travis again.


mikerhodes:
Looks good in Travis. Rename the iOS-only scheme to Tests iOS (squashed into your Change build schemes commit) then go ahead and merge.

PR #106, opened 2015-01-12
Adds TDURLConnectionChangeTracker class.

gadamc:
Modifies TDPuller to propertly set headers so that the User-Agent
header isn't overwritten.

Adds a few new ChangeTracker tests.

TDChangeTracker init method returns TDURLConnectionChangeTracker object.

### DIFF
--- a/Classes/common/touchdb/ChangeTracker/TDURLConnectionChangeTracker.m
+++ b/Classes/common/touchdb/ChangeTracker/TDURLConnectionChangeTracker.m
@@ -87,7 +87,7 @@
 
     if (!self.connection) {
         CDTLogError(CDTREPLICATION_LOG_CONTEXT, @"%@: GET %@ failed to create NSURLConnection",
-                   self, url.resourceSpecifier);
+                   self, TDCleanURLtoString(url));
         return NO;
     }
     [self.connection scheduleInRunLoop:[NSRunLoop currentRunLoop] forMode:NSRunLoopCommonModes];
@@ -96,8 +96,7 @@
     self.inputBuffer = [NSMutableData dataWithCapacity:0];
         
     self.startTime = [NSDate date];
-    CDTLogInfo(CDTREPLICATION_LOG_CONTEXT, @"%@: Started... <%@>", self, self.changesFeedURL);
-    CDTLogVerbose(CDTREPLICATION_LOG_CONTEXT, @"%@: GET %@", self, url.resourceSpecifier);
+    CDTLogInfo(CDTREPLICATION_LOG_CONTEXT, @"%@: Started... <%@>", self, TDCleanURLtoString(url));
 
     return YES;
 }
@@ -315,7 +314,8 @@
     
     if (!match) {
         CDTLogError(CDTREPLICATION_LOG_CONTEXT, @"%@: Unparseable response from %@. Did not find "
-                    @"start of the expected response: %@", self, self.request.URL, prefixString);
+                    @"start of the expected response: %@", self,
+                    TDCleanURLtoString(self.request.URL), prefixString);
     }
     
     return match;
--- a/Classes/common/touchdb/ChangeTracker/TDURLConnectionChangeTracker.m
+++ b/Classes/common/touchdb/ChangeTracker/TDURLConnectionChangeTracker.m
@@ -1,0 +1,324 @@
+//
+//  TDURLConnectionChangeTracker.m
+//  
+//
+//  Created by Adam Cox on 1/5/15.
+//  Copyright (c) 2015 IBM.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+//  except in compliance with the License. You may obtain a copy of the License at
+//    http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software distributed under the
+//  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+//  either express or implied. See the License for the specific language governing permissions
+//  and limitations under the License.
+//
+// <http://wiki.apache.org/couchdb/HTTP_database_API#Changes>
+//
+
+#import "TDURLConnectionChangeTracker.h"
+#import "TDRemoteRequest.h"
+#import "TDAuthorizer.h"
+#import "TDStatus.h"
+#import "TDBase64.h"
+#import "MYBlockUtils.h"
+#import "MYURLUtils.h"
+#import <string.h>
+#import "TDJSON.h"
+#import "CDTLogging.h"
+#import "TDMisc.h"
+
+#define kMaxRetries 6
+#define kInitialRetryDelay 0.2
+
+@interface TDURLConnectionChangeTracker()
+@property (strong, nonatomic) NSMutableData* inputBuffer;
+@property (strong, nonatomic) NSURLConnection *connection;
+@property (strong, nonatomic) NSMutableURLRequest *request;
+@property (strong, nonatomic) NSDate* startTime;
+@property (nonatomic, readwrite) NSUInteger totalRetries;
+@end
+
+@implementation TDURLConnectionChangeTracker
+
+- (BOOL)start
+{
+    if (self.connection) return NO;
+    
+    CDTLogInfo(CDTREPLICATION_LOG_CONTEXT, @"%@: Starting...", [self class]);
+    [super start];
+    
+    NSURL* url = self.changesFeedURL;
+    self.request = [[NSMutableURLRequest alloc] initWithURL:url];
+    self.request.cachePolicy = NSURLRequestReloadIgnoringLocalCacheData;
+    self.request.HTTPMethod = @"GET";
+    
+    // Add headers from my .requestHeaders property:
+    for(NSString *key in self.requestHeaders) {
+        [self.request setValue:self.requestHeaders[key] forHTTPHeaderField:key];
+    }
+    
+    // Add cookie headers from the NSHTTPCookieStorage:
+    NSArray* cookies = [[NSHTTPCookieStorage sharedHTTPCookieStorage] cookiesForURL:url];
+    NSDictionary* cookieHeaders = [NSHTTPCookie requestHeaderFieldsWithCookies:cookies];
+    NSArray *requestHeadersKeys = [self.requestHeaders allKeys];
+    for (NSString* headerName in cookieHeaders) {
+        if ([requestHeadersKeys containsObject:headerName]) {
+            CDTLogWarn(CDTREPLICATION_LOG_CONTEXT, @"%@ Overwriting '%@' header with value %@",
+                       self, headerName, cookieHeaders[headerName]);
+        }
+        [self.request setValue:cookieHeaders[headerName] forHTTPHeaderField:headerName];
+    }
+
+    if (self.authorizer) {
+        NSString* authHeader = [self.authorizer authorizeURLRequest:self.request forRealm:nil];
+        if (authHeader) {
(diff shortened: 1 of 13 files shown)


Classes/common/touchdb/TDPuller.m
```diff
@@ -83,7 +83,10 @@ - (void)beginReplicating
 - (void)startChangeTracker
 {
     Assert(!_changeTracker);
-    TDChangeTrackerMode mode = (_continuous && _caughtUp) ? kLongPoll : kOneShot;
+    //continuous / longpoll modes are not supported or available at the CDT* level.
+    //As such, the new TDURLConnectionChangeTracker also only supports one-shot query
+    //to the _changes feed. 
+    TDChangeTrackerMode mode = kOneShot;
```
mikerhodes: I suggest removing the constants for long poll and continuous as we're not supporting them right  now,

Classes/common/touchdb/TDPuller.m
```diff
@@ -101,8 +104,21 @@ - (void)startChangeTracker
...
+    BOOL hasUserAgentHeader = NO;
+    for (NSString *key in self.requestHeaders) {
+        if ([[key lowercaseString] isEqualToString:@"user-agent"]) {
+            hasUserAgentHeader = YES;
+        }
+    }
+    NSMutableDictionary* headers;
+    if (!hasUserAgentHeader) {
+        headers = $mdict({ @"User-Agent", [TDRemoteRequest userAgentHeader] });
+        [headers addEntriesFromDictionary:_requestHeaders];
```
mikerhodes: I think here we could just `headers[@"User-Agent"] = [TDRemoteRequest userAgentHeader]`? If `-userAgentHeader` could be `nil`, add a check to the `!hasUserAgentHeader`:

```objc
NSString *userAgent = [TDRemoteRequest userAgentHeader];
if (!hasUserAgentHeader && userAgent) {
     headers[@"User-Agent"] = userAgent;
}
```

Classes/common/touchdb/ChangeTracker/TDURLConnectionChangeTracker.m
```diff
@@ -0,0 +1,298 @@
...
+    for (NSString* headerName in cookieHeaders) {
+        [self.request setValue:cookieHeaders[headerName] forHTTPHeaderField:headerName];
+    }
+
+    if (self.authorizer) {
+        NSString* authHeader = [self.authorizer authorizeURLRequest:self.request forRealm:nil];
+        if (authHeader) {
+            [self.request setValue: authHeader forHTTPHeaderField:@"Authorization"];
+        }
+    }
```
mikerhodes: For both the cookie headers and this authorization header addition, issue a warning log line if you are overwriting something already in the custom headers passed to the request. I know that at least Authorization shouldn't happen, but we might change the pre-conditions for this class in future.

Classes/common/touchdb/ChangeTracker/TDURLConnectionChangeTracker.m
```diff
@@ -0,0 +1,298 @@
...
+    
+    self.connection = [[NSURLConnection alloc] initWithRequest:self.request
+                                                      delegate:self
+                                              startImmediately:NO];
+
+    if (!self.connection) {
+        CDTLogError(CDTREPLICATION_LOG_CONTEXT, @"%@: GET %@ failed to create NSURLConnection",
+                   self, @"cleaned url.resourceSpecifier");
```
mikerhodes: Looks like the second argument needs replacing with the real variable?

Classes/common/touchdb/ChangeTracker/TDURLConnectionChangeTracker.m
```diff
@@ -0,0 +1,298 @@
...
+}
+
+- (void)retryOrError:(NSError*)error
+{
+    CDTLogInfo(CDTREPLICATION_LOG_CONTEXT, @"%@: retryOrError: %@", [self class], error);
+    if (++_retryCount <= kMaxRetries && TDMayBeTransientError(error)) {
+        self.totalRetries++;
+        [self clearConnection];
+        NSTimeInterval retryDelay = kInitialRetryDelay * (1 << (_retryCount - 1));
+        [self performSelector:@selector(start) withObject:nil afterDelay:retryDelay];
```
mikerhodes: Is `-performSelector:...` going to be be called on the replicator's thread? I'm assuming it will because from what I can see `-retryOrError` is called from the NSURLConnection callbacks which should be called in the replicator's thread due to the run queue for that thread being used. Can you confirm that?

Classes/common/touchdb/ChangeTracker/TDURLConnectionChangeTracker.m
```diff
@@ -0,0 +1,298 @@
...
+
+-(void) connection:(NSURLConnection *)connection didFailWithError:(NSError *)error
+{
+    [self retryOrError:error];
+}
+
+- (BOOL)receivedDataBeginsCorrectly:(NSData*)body
+{
+    // Convert to string:
+    NSString* bodyStr = [body my_UTF8ToString];
```
mikerhodes: I'm a bit worried about loading the entire response into a string -- if we're in an error state it could be larger than we expect. We know we should have UTF8 text here, we can just read the buffer itself:

```objc
	NSString* inputString = @"    {\"results\":";
	NSData* inputData = [inputString dataUsingEncoding:NSUTF8StringEncoding];
	NSUInteger inputLength = [inputData length];
	
	NSData *prefixData = [@"{\"results\":" dataUsingEncoding:NSUTF8StringEncoding];
	int prefixLength = [prefixData length];
	
	for (NSUInteger index = 0; index < inputLength; index++)
	{
		char currentChar;
		NSRange currentCharRange = NSMakeRange(index, 1);
		[inputData getBytes:&currentChar range:currentCharRange];
	
		// If it's the opening {, check for valid start JSON
	   	if (currentChar == '{') {
			NSRange r = NSMakeRange(index, prefixLength);
			char buf[prefixLength];
			
			if (inputLength >= (index + prefixLength)) {
				[inputData getBytes:buf range:r];
				BOOL match = (memcmp(buf, prefixData.bytes, prefixLength) == 0);
				if (match) {
					NSLog(@"found result start JSON");
				}
			}
			break;  // once we've seen a {, break always as can't succeed if we've not
					// already.
		}
	}
```

Classes/common/touchdb/ChangeTracker/TDURLConnectionChangeTracker.m
```diff
@@ -0,0 +1,298 @@
...
+    NSString* errorMessage = nil;
+    NSInteger numChanges = [self receivedPollResponse:self.inputBuffer errorMessage:&errorMessage];
+    
+    if (numChanges < 0) {
+        // unparseable response. See if it gets special handling:
+        if ([self receivedDataBeginsCorrectly:self.inputBuffer]) {
+            
+            // The response at least starts out as what we'd expect, so it looks like the connection
+            // was closed unexpectedly before the full response was sent.
+            NSTimeInterval elapsed = [self.startTime timeIntervalSinceNow] * -1000.0;
```
mikerhodes: Shouldn't this just be `* -1.0`?

Classes/common/touchdb/ChangeTracker/TDURLConnectionChangeTracker.m
```diff
@@ -0,0 +1,298 @@
...
+
+-(void) connection:(NSURLConnection *)connection didReceiveResponse:(NSURLResponse *)response
+{
+    NSHTTPURLResponse *httpresponse = (NSHTTPURLResponse *)response;
+    NSInteger status = httpresponse.statusCode;
+    CDTLogVerbose(CDTREPLICATION_LOG_CONTEXT, @"%@: didReceiveResponse, status %ld", [self class], (long)status);
+    
+    [self.inputBuffer setLength:0];
+
+    if (TDStatusIsError(status)) {
```
mikerhodes: Fix the conversion warnings between `status` and a `TDStatus`.

Classes/common/touchdb/ChangeTracker/TDURLConnectionChangeTracker.m
gadamc: ok

Classes/common/touchdb/ChangeTracker/TDURLConnectionChangeTracker.m
gadamc: woops, sorry. the cleaned URL code isn't merged in yet. chicken or egg?

Classes/common/touchdb/ChangeTracker/TDURLConnectionChangeTracker.m
gadamc: That's my understanding as well. But I don't have a test to confirm this -- are you looking for a test, or just written confirmation?

Classes/common/touchdb/ChangeTracker/TDURLConnectionChangeTracker.m
gadamc: ok

Classes/common/touchdb/ChangeTracker/TDURLConnectionChangeTracker.m
gadamc: yes

Classes/common/touchdb/ChangeTracker/TDURLConnectionChangeTracker.m
gadamc: good point. make sense.

Classes/common/touchdb/TDPuller.m
gadamc: there'll be a number of little tweaks in TDChangeTracker if we do this. Okay? i'll make this in a separate commit so we can easily change back.

Classes/common/touchdb/TDPuller.m
gadamc: ok

ReplicationAcceptance/ReplicationAcceptance/ChangeTrackerDelegate.m
```diff
@@ -0,0 +1,29 @@
...
+//
+//  Created by Adam Cox on 1/8/15.
+//
+//
+
+#import "ChangeTrackerDelegate.h"
+#import "TDURLConnectionChangeTracker.h"
+
+@implementation ChangeTrackerDelegate
+
```
gadamc: need to add an -init method that initializes the block properties to blocks that just return. otherwise, if those blocks are not defined and they are called below, there will be a crash.

Classes/common/touchdb/TDPuller.m
mikerhodes: Okay, let's leave the constants here for now. You're right, this PR should remain focussed

ReplicationAcceptance/ReplicationAcceptance/ChangeTrackerDelegate.m
mikerhodes: I'm not sure this comment belongs in a PR given it seems to be a note-to-self.

mikerhodes:
Looks like this will be ready after a few more tweaks.

gadamc:
I'll remove the changes in the commit that started to remove the longpoll/continuous stuff before I merge.. but I didn't want to rebase now since that seems to mess up the comments. 

Also, I added another commit which changes the response to an authentication challenge when that challengs is not 'Basic' or 'ServerTrust'. Previously, I called 'cancelAuthenticaionChalleng', but now it calls 'performDefaultHandlingForAuthenticationChallenge', which I think is the correct way to do it (and how TDRemoteRequest handles that situation. 

Other than that, I'm waiting for final review.

mikerhodes:
+1 to merge once squashed. I'll make a release after that.

PR #107, opened 2015-01-21
43494 last sequence not written to remote checkpoint doc

gadamc:
Fixes bug where a replicator was not writing the last sequence
value to the remote checkpoint document because it's thread was
quiting before the request was made.

Additionally, a change was made to TDReplciator -setError to prevent 
overwriting the error when it is for a local database 
being deleted. Otherwise, the error was being overwritten by a generic
"TDHTTP" Intenal Server error (code = 500), which was causing a 
previously written test to fail (testPullErrorsWhenLocalDatabaseIsDeleted)

Finally, the TDReplicator -clearDbRef method was removed. This method 
seemed to be appropriate when all replications were executed on a main 
replication thread. Now, it could interfere with proper synchronization 
of the last sequence value betweent the local and remote checkpoint documents.
    
Additionally, the pointer to the TD_Database is now set to nil after the 
replicator's thread has exited its runloop. Setting the pointer to nil is 
still necessary since it's the mechanism to make each TDReplicator 
object only useable once.

### DIFF
--- a/Classes/common/touchdb/TDReplicator.h
+++ b/Classes/common/touchdb/TDReplicator.h
@@ -122,4 +122,7 @@ extern NSString* TDReplicatorStoppedNotification;
 /** JSON-compatible array of status info about active remote HTTP requests. */
 @property (readonly) NSArray* activeRequestsStatus;
 
+/** Exposed for testing. Returns the doc ID for the checkpoint document. */
+- (NSString *)remoteCheckpointDocID;
+
 @end
--- a/Classes/common/touchdb/TDReplicator.m
+++ b/Classes/common/touchdb/TDReplicator.m
@@ -100,16 +100,6 @@ NSString* TDReplicatorStartedNotification = @"TDReplicatorStarted";
     [_host stop];
 }
 
-- (void)clearDbRef
-{
-    // If we're in the middle of saving the checkpoint and waiting for a response, by the time the
-    // response arrives _db will be nil, so there won't be any way to save the checkpoint locally.
-    // To avoid that, pre-emptively save the local checkpoint now.
-    if (_savingCheckpoint && _lastSequence)
-        [_db setLastSequence:_lastSequence withCheckpointID:self.remoteCheckpointDocID];
-    _db = nil;
-}
-
 - (void)databaseClosing
 {
     //this can be called from another thread, but we need to execute it the replicator's thread
@@ -120,9 +110,7 @@ NSString* TDReplicatorStartedNotification = @"TDReplicatorStarted";
 }
 - (void)databaseClosingOnMyThread
 {
-    [self saveLastSequence];
     [self stop];
-    [self clearDbRef];
 }
 
 - (NSString*) description {
@@ -290,6 +278,9 @@ NSString* TDReplicatorStartedNotification = @"TDReplicatorStarted";
                                         beforeDate: [NSDate dateWithTimeIntervalSinceNow:0.1]])
             ;
         
+        // clear the reference to the db
+        _db = nil;
+        
         CDTLogInfo(CDTREPLICATION_LOG_CONTEXT, @"TDReplicator thread exiting");
     }
 }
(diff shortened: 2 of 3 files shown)


mikerhodes:
> Additionally, the pointer to the TD_Database is now set to
nil after the replicator's thread has exited its runloop.
Setting the pointer to nil is still necessary since it's
the mechanism to make each TDReplicator object only
useable once.

This kind of implied logic is what makes maintaining code hard. It needs to be made explicit. If it's possible to do in a few lines, do it in this PR, otherwise file a FB case.

gadamc:
>> Additionally, the pointer to the TD_Database is now set to
nil after the replicator's thread has exited its runloop.
Setting the pointer to nil is still necessary since it's
the mechanism to make each TDReplicator object only
useable once.

>This kind of implied logic is what makes maintaining code hard. It needs to be made explicit. If it's possible to do in a few lines, do it in this PR, otherwise file a FB case.

So, setting the pointer to nil was part of -clearDbRef method. There's lots of logic in the TD* code base that is hard for me to follow and I'm trying not to alter it here. But getting rid of -clearDbRef is necessary for the lastSequence synchronization issue. I could have left in the `_db = nil` part of that method, but that too would impact the lastSequence synchronization as well because the callback block after the final write to the remote checkpoint doc needs the local TD_Database in order to write the local lastSequence value. So, we can't set `_db=nil` until after this is done. I'm assuming there is a reason why `_db` was set to nil by the original authors and I'm being cautious (after all it's a weak reference). So this is why I am setting this after the runloop exits.

mikerhodes:
>>> Additionally, the pointer to the TD_Database is now set to
        nil after the replicator's thread has exited its runloop.
        Setting the pointer to nil is still necessary since it's
        the mechanism to make each TDReplicator object only
        useable once.

>> This kind of implied logic is what makes maintaining code hard. It needs to be made explicit. If it's possible to do in a few lines, do it in this PR, otherwise file a FB case.

> So, setting the pointer to nil was part of -clearDbRef method. There's lots of logic in the TD* code base that is hard for me to follow and I'm trying not to alter it here. 

Okay, so the implied logic was already there; sensible not to try and alter it at this point.

> I'm assuming there is a reason why _db was set to nil by the original authors and I'm being cautious.

However, this appears to contradict that your assertion about "Setting the pointer to nil is still necessary since..." as it doesn't seem you're really sure if this is the case?

gadamc:
Setting the pointer to nil still seems necessary since there is an Assert(_db, ...) check at the beginning of `-startReplicatorTasks` and the message indicates that this is to prevent a TDReplicator from being used twice. That Assert has been there from the beginning and absolutely prevents somebody from using a TDReplicator twice. (However, there is also the check for the NSThread _replicatorThread at the beginning of -start, which was ported over from the old TDReplicatorManager, and seemed like a reasonable thing to ensure the object is just single use.)

But the Assert seems like a harsh response to me and I'm wondering if there's some other reason 

Also, it looks like there is a race condition between the TDReplicator task actually starting and the TDReplicator being set to receive notifications that the database is closing. It seems possible to me that the TD_Database can be deleted or closed after instantiation of TDReplicator but before the TDReplicator has registered with the NSNotificationCenter for TD_Database close/delete. I should be able to create a TD_Database, create a TDReplicator, delete the TD_Database and then try to start the TDReplicator, at which point it would hit the Assert(_db) and fail because the _db pointer in TDReplicator is a weak reference.  

Either way, an Assert for attempting to restart the TDReplicator or for protecting against this race condition occurs seems a bit harsh of a response to me. And maybe the original authors didn't even think about the possible race condition. The string message from the Assert doesn't seem to imply they did. The message is the same as it's always been: "Can't restart an already stopped TDReplicator".

So... it's still necessary due to the Assert(_db) because I've kept in that original code. But I'm also hedging here, assuming that there's a reason for something as strong as an Assert that I'm just not seeing.

mikerhodes:
Seems reasonable to Assert to me -- it'd be a bit bad to ship an app which tried to restart a replicator and the Assert is likely to warn an app dev about it.

To me I think just working out the right name for the state variable is the only thing stopping this PR being merged.

gadamc:
So, I'll change to `_replicationStopped` and add the commit hash to the last commit message as you suggested. 

Then, I'll rebase and merge. Okay to do this?

gadamc:
State variable has been changed and commit messages updated (I've rebased now...). Ready to merge if you are.

mikerhodes:
+1. I'm going to merge this so Tom can test his fixes against CouchDB 2.0.

For future reference, remember to add subjects to your commit messages, 26cfcda doesn't have one.

PR #109, opened 2015-02-03
Sequence number compatibility fix

tomblench:
In CouchDB 2.0/Cloudant Local, sequence numbers as used in checkpoints can be a number or an array.

Ensure that this is handled correctly and encoding/decoding issues are mitigated.


Classes/common/touchdb/TDInternal.h
```diff
@@ -89,8 +89,8 @@
 
 @interface TD_Database (Replication_Internal)
 - (void)stopAndForgetReplicator:(TDReplicator*)repl;
-- (NSString*)lastSequenceWithCheckpointID:(NSString*)checkpointID;
-- (BOOL)setLastSequence:(NSString*)lastSequence withCheckpointID:(NSString*)checkpointID;
+- (id)lastSequenceWithCheckpointID:(NSString*)checkpointID;
+- (BOOL)setLastSequence:(id)lastSequence withCheckpointID:(NSString*)checkpointID;
```
mikerhodes: Prefer `NSObject*` to `id`.

Classes/common/touchdb/TD_Database+Replication.m
```diff
@@ -66,25 +67,36 @@ - (void)replicatorDidStop:(NSNotification *)n
...
+        return nil;
+    }
 }
 
-- (BOOL)setLastSequence:(NSString *)lastSequence withCheckpointID:(NSString *)checkpointID
+- (BOOL)setLastSequence:(id)lastSequence withCheckpointID:(NSString *)checkpointID
 {
     __block BOOL result;
+    id lastSequenceJson;
+    // write the sequence as a json dict of {"seq": <data>}
```
mikerhodes: Worth making a note of this in the schema definition?

mikerhodes:
I think the remaining this here is to create a migration to take old entries in the replicator table to the new format?

Classes/common/touchdb/TD_Database.m
```diff
@@ -367,7 +367,33 @@ INSERT INTO INFO (key, value) VALUES ('publicUUID',  '%@')",
...
+                NSDictionary *lastSequenceJsonMaybe = [TDJSON JSONObjectWithData:lastSequence options:0 error:nil];
+                if (lastSequenceJsonMaybe == nil) {
+                    // sequence not in json format, so upgrade it
+                    NSString *lastSequenceString = [replicators stringForColumn:@"last_sequence"];
+                    // the sequence might have been a number, or a string "<number-base64>"
+                    NSDictionary *dict = @{@"seq": lastSequenceString};
+                    NSData *lastSequenceJson = [TDJSON dataWithJSONObject:dict options:0 error:nil];
+                    [db executeUpdate:@"UPDATE replicators SET last_sequence=? WHERE rowid=?", lastSequenceJson, @(rowid)];
+                }
+            }
```
mikerhodes: I think we might need to upgrade the hundreds place, [as detailed at the top of the method](https://github.com/cloudant/CDTDatastore/blob/43944-couchdb-20-clocal-compat/Classes/common/touchdb/TD_Database.m#L247), for breaking changes to the schema as we're changing the way `last_sequence` is stored to something that older versions of the library will unpack incorrectly?

mikerhodes:
Okay, this looks better now -- care to squash things up and then merge?

mikerhodes:
I squashed this up, waiting for Travis before merging.

PR #114, opened 2015-03-13
Bump cocoalumberjack version to 2.0.0-rc2

rhyshort:
Bump cocoalumber to version 2.0.0-rc2


tomblench:
@rhyshort why are we on RC versions of cocoalumberjack?

rhyshort:
We made the decision when adopting lumberjack to jump in with the 2.0 releases rather than use the older 1.x releases, to save on doing a migration twice, once from NSLog to CocoaLumberjack 1.x, then to Cocoalumbjack 2.0.0

mikerhodes:
+1 given the tests pass.

PR #117, opened 2015-03-25
44627 Add migration document, update other query docs

alfinkel:
- Add query-migration.md to detail the differences between the legacy
indexing/query and new query implementations.  The document uses the
differences detailed to showcase a migration path for customers that
are looking to change their existing indexing/query implementations
to use the new query logic.
- Remove index-query.md as this is the legacy document.
- Update query.md to no longer reference the legacy document and remove
the #import "CDTDatastore_Query.h" reference as it is no longer needed.
- Update README.md to provide a link to the migration document.
- Update CHANGELOG.md
- Ref: BugID: 44627

reviewer @mikerhodes
reviewer @tomblench


README.md
```diff
@@ -255,6 +255,10 @@ CDTQResultSet *result = [datastore find:query];
 
 See [Index and Querying Data](https://github.com/cloudant/CDTDatastore/blob/master/doc/query.md).
 
+For information about migrating your legacy indexing/query code to the new Cloudant Query - iOS implementation.
```
tomblench: These shouldn't be two sentences as they are closely related. I think this needs a more substantial introduction along the lines of "As of version xxx, indexing/query code has been re-written and has many more features than the previous implementation. For implementation about migrating existing indexing /query code see yyy"

alfinkel: Clarified migration message in README.  Done in 6a093ae.

mikerhodes:
+1

doc/query-migration.md
```diff
@@ -0,0 +1,99 @@
...
+                                 initWithDatastore:datastore error:nil];
+[indexManager ensureIndexedWithIndexName:@"default"
+                               fieldName:@"name" error:nil];
+[indexManager ensureIndexedWithIndexName:@"age"
+                               fieldName:@"age" error:nil];
+```
+
+You could now create a single index like this:
+
+```
```
tomblench: objc here so we get the right markup syntax

tomblench:
+1

doc/query-migration.md
alfinkel: Done

PR #118, opened 2015-03-30
Index documents with long deleted branches correctly [45667]

mikerhodes:
Previously the index updating code used the revision provided by the
changes feed from the local database. However, this feed always provides
the revision with the highest revID, regardless of whether it's the
winning revision or not.

An incorrect revision might occur when there is a document with two
branches, and the longer branch is tombstoned. The tombstone revision
has a higher revID so is returned by the changes feed. However, we
actually want to index the non-deleted revision -- i.e., the current
winner.

```
            winner (returned by getDocumentWithId:...)
              |
       /---- 2-4f
1-97 --
       \---- 2-33 ---- 3-bc
                        |
                      deleted (returned by changes feed)
```

This fix uses a brute force approach. It simply uses
`[CDTDatastore -getDocumentsWithIds:]` to get the real winning revision
for every entry in the changes feed. The winning revision is used to
update the index. The list from changes is still needed to cover any
deleted items and to get the sequence number the index is up to date
with.

reviewer @tomblench 
reviewer @rhyshort

### DIFF
--- a/Classes/common/Indexing/CDTIndexManager.m
+++ b/Classes/common/Indexing/CDTIndexManager.m
@@ -540,10 +540,11 @@ static const int VERSION = 1;
 {
     BOOL success = TRUE;
     TDChangesOptions options = {.limit = 10000,
-                                .contentOptions = 0,
-                                .includeDocs = TRUE,
-                                .includeConflicts = FALSE,
-                                .sortBySequence = TRUE};
+        .contentOptions = 0,                                
+        .includeDocs = NO,  // we only need the docIDs and sequences, body
+                            // is retrieved separately
+        .includeConflicts = FALSE,
+        .sortBySequence = TRUE};
 
     TD_RevisionList *changes;
     long lastSequence = [index lastSequence];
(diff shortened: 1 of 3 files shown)


Tests/Tests/IndexManagerTests.m
```diff
@@ -839,6 +843,117 @@ - (void)testNewIndexManagerPicksUpDocumentsAddedBeforeEnsureIndexedDuringQuery
...
+                         |
+                         deleted (returned by _changes)
+ 
+ This test ensures we index the winning revision rather than the one with the highest 
+ revision id.
+
+ The bug we are fixing (BugId: 45667) results from the indexing code updating the
+ index used the revision the _changes feed provides. It's possible the changed
+ feed gives a revision which isn't the winning revision. For example, it gives
+ 3-bc for the tree above, rather than 204f.
```
tomblench: s/204f/2-4f/

Tests/Tests/IndexManagerTests.m
```diff
@@ -839,6 +843,117 @@ - (void)testNewIndexManagerPicksUpDocumentsAddedBeforeEnsureIndexedDuringQuery
...
+        [db executeUpdate:sql withArgumentsInArray:args];
+    }];
+    XCTAssertNotNil([datastore getDocumentWithId:testDocId 
+                                             rev:@"3-bc3b310ab4ff877104b1aeb53d0b8457" 
+                                           error:nil]);
+    
+    // Should still have the same result.
+    CDTQueryResult *res2 = [im queryWithDictionary:@{@"description": @"desc2"} error:nil];
+    XCTAssertEqual(res2.documentIds.count, (NSUInteger)1);
+    XCTAssertEqualObjects(res2.documentIds[0], testDocId);
```
tomblench: Don't you need an assert on the rev id or document contents here? Asserting on the doc id doesn't ensure you have the right revision.

tomblench: OK I see that the query asserts on the document contents but i think it would make it more explicit if you added an assert on revid here

Tests/Tests/IndexManagerTests.m
```diff
@@ -839,6 +843,117 @@ - (void)testNewIndexManagerPicksUpDocumentsAddedBeforeEnsureIndexedDuringQuery
...
+ The bug we are fixing (BugId: 45667) results from the indexing code updating the
+ index used the revision the _changes feed provides. It's possible the changed
+ feed gives a revision which isn't the winning revision. For example, it gives
+ 3-bc for the tree above, rather than 204f.
+ */
+- (void)test_CanIndexConflictedDocWithLatestRevDeleted
+{
+    CDTDatastore *datastore;
+    
+    NSString *factoryPath = [self createTemporaryDirectoryAndReturnPath];
```
tomblench: Does this test need to do any cleanup (because it creates its own factory) or is this handled in teardown?

Classes/common/Indexing/CDTIndexManager.m
```diff
@@ -540,10 +540,11 @@ - (BOOL)updateIndex:(CDTIndex *)index error:(NSError *__autoreleasing *)error
 {
     BOOL success = TRUE;
     TDChangesOptions options = {.limit = 10000,
-                                .contentOptions = 0,
-                                .includeDocs = TRUE,
-                                .includeConflicts = FALSE,
-                                .sortBySequence = TRUE};
+        .contentOptions = 0,                                
+        .includeDocs = NO,  // we only need the docIDs and sequences, body
+                            // is retrived separately
```
tomblench: s/retrived/retrieved/

Tests/Tests/IndexManagerTests.m
mikerhodes: c855cb063f0fa0c424c0e9ff098f5d26580561af

Tests/Tests/IndexManagerTests.m
mikerhodes: Fixed in 81601ef9447cb305456675bef4b22d88841e1d86

Tests/Tests/IndexManagerTests.m
mikerhodes: Fixed in d2af7a4ca04b2265eaa1cb1080b964de84183e67

Classes/common/Indexing/CDTIndexManager.m
mikerhodes: Fixed in e6dc6c21dfb889901ed6570b10064504537ea6a8

Tests/Tests/IndexManagerTests.m
rhyshort: I agree with @tomblench asserting on the document revision will make it easier to read

tomblench:
OK +1

rhyshort:
+1

mikerhodes:
Travis is taking forever, but iOS tests have passed so I'm merging.

PR #120, opened 2015-04-01
Add CloudKit-inspired changes fetcher

mikerhodes:
CDTFetchChanges can be thought a more useful changes feed for
the outside world. It returns either the winning revision for
changes or new documents, or the IDs for deleted documents,
sparing callers from the plumbing required to work this out.

The CDTFetchChanges class allows users to find out what has
changed in a database. Inspired by CloudKit's fetch operations,
it uses a block-based method for notifying of the changes to
a database. The class uses sequence values, as one would expect,
to do this.

The class was build in response to our own issues reading
changes for our indexing code.

BugId: 45845

reviewer: @indisoluble 
reviewer: @gadamc

### DIFF
--- a/Classes/common/CDTFetchChanges.h
+++ b/Classes/common/CDTFetchChanges.h
@@ -1,0 +1,167 @@
+//
+//  CDTFetchChanges.h
+//  CloudantSync
+//
+//  Created by Michael Rhodes on 31/03/2015.
+//  Copyright (c) 2015 IBM Cloudant. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+//  except in compliance with the License. You may obtain a copy of the License at
+//    http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software distributed under the
+//  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+//  either express or implied. See the License for the specific language governing permissions
+//  and limitations under the License.
+
+#import <Foundation/Foundation.h>
+
+@class CDTDatastore;
+@class CDTDocumentRevision;
+
+/**
+ CDTFetchChanges objects report changes to a database. The changes include documents which
+ have been created, updated or deleted. Use this class to update your user interface efficiently
+ when notified a replication is complete, for example, by updating elements for only those
+ documents that have changed.
+
+ CDTFetchChanges provides callbacks for updated and created documents (documentChangedBlock)
+ and deleted documents (documentWithIDWasDeletedBlock). It also provides a completion block
+ for when all changes have been received and processed (fetchRecordChangesCompletionBlock).
+ The documentChangedBlock is passed the current revision of each document, the
+ documentWithIDWasDeletedBlock is just passed the id of the deleted document.
+
+ In addition, CDTFetchChanges is able to report on changes since a given point. The completion
+ block provides a sequence value which can be passed to future requests to have only changes
+ from that point reported. This allows your application to make incremental updates
+ in response to changes in a database.
+
+ The sequence value should be treated as an opaque value, but can be saved and used across
+ application sessions.
+
+ The blocks you assign to process the fetched documents are executed serially on an
+ internal queue managed by the operation. Your blocks must be capable of executing
+ on a background thread, so any tasks that require access to the main thread must
+ be redirected accordingly.
+ */
+@interface CDTFetchChanges : NSOperation
+
+#pragma mark Properties
+
+/**
+ The datastore whose changes should be reported.
+
+ Typically this is set with the initialiser.
+ */
+@property (nonatomic, strong) CDTDatastore *datastore;
+
+/**
+ The sequence value identifying the starting point for reading changes.
+
+ The value to use for this is returned by the completion block for a fetch request. This sequence
+ value can be used to
+ receive changes that have occured since the previous request.
+ Treat the sequence value as an opaque string; different implementations may provide
+ differently formatted values. A given sequence value should only be used with the
+ database that it was received from.
+
+ Typically this is set with the initialiser.
+ */
+@property (nonatomic, copy) NSString *startSequenceValue;
+
+#pragma mark Callbacks
+
+/**
+ The block to execute for each changed document.
+
+ The block returns no value and takes the following parameter:
+
+ <dl>
+ <dt>revision</dt>
+ <dd>The winning revision for the document that changed.</dd>
+ </dl>
+
+ The operation object executes this block once for each document in the database that changed
+ since the previous fetch request. Each time the block is executed, it is executed
+ serially with respect to the other progress blocks of the operation. If no documents
+ changed, the block is not executed.
+
+ If you intend to use this block to process results, set it before executing the
+ operation or submitting it to a queue.
+ */
+@property (nonatomic, copy) void (^documentChangedBlock)(CDTDocumentRevision *revision);
+
+
+/**
+ The block to execute for each deleted document.
+ 
+ The block returns no value and takes the following parameters:
+ 
+ <dl>
+ <dt>docId</dt>
+ <dd>The document id for the deleted document.</dd>
+ </dl>
+ 
+ The operation object executes this block once for each document in the database that
+ was deleted since the previous fetch request. Each time the block is executed, it 
+ is executed serially with respect to the other progress blocks of the operation. 
+ If no documents were deleted, the block is not executed.
+ 
+ If you intend to use this block to process results, set it before executing the 
+ operation or submitting it to a queue. 
+ */
+@property (nonatomic, copy) void (^documentWithIDWasDeletedBlock)(NSString *docId);
+
+/**
+ The block to execute when all changes have been reported.
+
+ The block returns no value and takes the following parameters:
+
+ <dl>
+ <dt>newSequenceValue</dt>
+ <dd>The new sequence value from the database. You can store this value locally and use it
+ during subsequent fetch operations to limit the results to documents that changed since
+ this operation executed. A sequence value is only valid for the database it was
+ originally retrieved from.</dd>
+
+ <dt>startSequenceValue</dt>
+ <dd>The sequence value you specified when you initialized the operation object.</dd>
+
+ <dt>fetchError</dt>
+ <dd>An error object containing information about a problem, or nil if the changes are
+ retrieved successfully.</dd>
+ </dl>
+
+ The operation object executes this block only once, at the conclusion of the operation. It
+ is executed after all individual change blocks.
+ The block is executed serially with respect to the other progress blocks of the operation.
+
+ If you intend to use this block to process results, set it before executing the operation or
+ submitting the operation object to a queue.
+ */
+@property (nonatomic, copy) void (^fetchRecordChangesCompletionBlock)
+    (NSString *newSequenceValue, NSString *startSequenceValue, NSError *fetchError);
+
+#pragma mark Initialisers
+
+/**
+ Initializes and returns an object configured to fetch changes in the specified database.
+ 
+ When initializing the fetch object, use the sequence value from a previous fetch request if 
+ you have one. You can archive sequence values and write them to disk for later use if needed.
+ 
+ After initializing the operation, associate at least one progress block with the operation 
+ object (excluding the completion block) to process the results. 
+ 
+ @param datastore The datastore containing the changes that should be fetched.
+ @param previousServerChangeToken The sequence value from a previous fetch. This is the value
+            passed to the completionHandler for this object. This value limits the changes
+            retrieved to those occuring after this sequence value. Pass `nil` to receive
+            all changes.
+ 
+ @return An initialised fetch object.
+ */
+- (instancetype)initWithDatastore:(CDTDatastore *)datastore
+               startSequenceValue:(NSString *)startSequenceValue;
+
+
+@end
--- a/Classes/common/CDTFetchChanges.m
+++ b/Classes/common/CDTFetchChanges.m
@@ -1,0 +1,115 @@
+//
+//  CDTFetchChanges.m
+//  CloudantSync
+//
+//  Created by Michael Rhodes on 31/03/2015.
+//  Copyright (c) 2015 IBM Cloudant. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+//  except in compliance with the License. You may obtain a copy of the License at
+//    http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software distributed under the
+//  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+//  either express or implied. See the License for the specific language governing permissions
+//  and limitations under the License.
+
+#import "CDTFetchChanges.h"
+
+#import "CDTDatastore.h"
+#import "CDTDocumentRevision.h"
+
+#import "TD_Database.h"
+
+@implementation CDTFetchChanges
+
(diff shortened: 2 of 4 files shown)


Classes/common/CDTFetchChanges.h
```diff
@@ -0,0 +1,163 @@
...
+@class CDTDocumentRevision;
+
+/**
+ Read changes to a database.
+ 
+ CDTFetchChanges reads the changes to a database from a given sequence value. It provides
+ callbacks for updated and created documents (documentChangedBlock) and deleted documents
+ (documentWithIDWasDeletedBlock). It also provides a completion block for when all changes
+ have been received and processed.
+
```
gadamc: Could be more verbose (with examples?) about how using this class differs from the standard changes feed?

I think I remember you saying that this would suppress non-winning revisions? I see that you've documented this down below for the callback block, but perhaps it's worth a mention in the Class description.

Classes/common/CDTFetchChanges.h
```diff
@@ -0,0 +1,163 @@
...
+/**
+ Read changes to a database.
+ 
+ CDTFetchChanges reads the changes to a database from a given sequence value. It provides
+ callbacks for updated and created documents (documentChangedBlock) and deleted documents
+ (documentWithIDWasDeletedBlock). It also provides a completion block for when all changes
+ have been received and processed.
+ 
+ The CDTFetchChanges object will always return _all_ changes from the given sequence value.
+
```
gadamc: Saying that it returns "_all_" changes may be confusing since this class filters out non-winning revisions. I think it might be especially confusing to a couchdb/cloudant veteran. 

I think your point is that each time you call ```-start``` with a particular value for ```previousServerSequenceValue```, the winning revision for all documents with changes since ```previousServerSequenceValue``` will be supplied to the appropriate callback block.

Classes/common/CDTFetchChanges.h
```diff
@@ -0,0 +1,163 @@
...
+/**
+ The datastore whose changes should be fetched.
+ 
+ Typically this is set with the initialiser.
+ */
+@property (nonatomic, strong) CDTDatastore *datastore;
+
+//@property (nonatomic, assign) NSUInteger resultsLimit;
+//@property (nonatomic, readonly) BOOL moreComing;
+
```
gadamc: should these be removed?

Classes/common/CDTFetchChanges.h
```diff
@@ -0,0 +1,163 @@
...
+
+/**
+ The sequence value identifying the starting point for reading changes.
+ 
+ Each changes request returns a sequence value in the completion block. This sequence value
+ can be used to receive changes that have occured since the previous request. Treat the
```
gadamc: Perhaps consider dropping these first two sentences here. Seems to be confusing the description of this property.

Classes/common/CDTFetchChanges.h
```diff
@@ -0,0 +1,163 @@
...
+ sequence value as an opaque string; different implementations may provide differently
+ formatted values. A given sequence value should only be used with the database that
+ it was received from.
+
```
gadamc: I guess I don't understand why you're using NSString here an not just SequenceNumber type.

Classes/common/CDTFetchChanges.h
```diff
@@ -0,0 +1,163 @@
...
+ 
+ Each changes request returns a sequence value in the completion block. This sequence value
+ can be used to receive changes that have occured since the previous request. Treat the
+ sequence value as an opaque string; different implementations may provide differently
+ formatted values. A given sequence value should only be used with the database that
+ it was received from.
+ 
+ Typically this is set with the initialiser.
+ */
+@property (nonatomic, copy) NSString *previousServerSequenceValue;
```
gadamc: Perhaps "initialSequnceValue"? I don't understand the "Server" part of the name here. It seems confusing and people will want to use this to pull changes from a remote (and then realize that they can't).

Classes/common/CDTFetchChanges.h
```diff
@@ -0,0 +1,163 @@
...
+ The block is executed serially with respect to the other progress blocks of the operation.
+ 
+ If you intend to use this block to process results, set it before executing the operation or 
+ submitting the operation object to a queue. 
+ */
+@property (nonatomic, copy) void (^fetchRecordChangesCompletionBlock)( 
+NSString *serverSequenceValue, 
+NSString *clientSequenceValue, 
+NSError *fetchError);
+
```
gadamc: I don't understand the "client/server" model here.  This is all about changes to a local CDTDatastore object, right? 

Perhaps
```
NSString *currentSequenceValue;
NSString *initialSequenceValue;
```

Classes/common/CDTFetchChanges.h
```diff
@@ -0,0 +1,163 @@
...
+#pragma mark Initialisers
+
+/**
+ Initializes and returns an object configured to fetch changes in the specified database.
+ 
+ When initializing the fetch object, use the sequence value from a previous fetch request if 
+ you have one. You can archive sequence values and write them to disk for later use if needed.
+
```
gadamc: One can also obtain the SequenceNumber value for a particular CDTDocumentRevision, which they could use in the initialization, in principle. However, that is not a NSString and they'd have to convert it first: ```[NSNumber numberWithLongLong:lastSequence] stringValue]```

Classes/common/CDTFetchChanges.m
```diff
@@ -0,0 +1,118 @@
...
+{
+    TDChangesOptions options = {.limit = 500,
+        .contentOptions = 0,                                
+        .includeDocs = NO,  // we only need the docIDs and sequences, body is retrieved separately
+        .includeConflicts = FALSE,
+        .sortBySequence = TRUE};
+    
+    TD_RevisionList *changes;
+    SequenceNumber lastSequence = [_previousServerSequenceValue longLongValue];
+
```
gadamc: perhaps should check to see if any of the changes blocks are set? If they are all null (obviously a programmer's error), this would just spin through the changes unnecessarily.  Maybe my suggestion is too cautious.

gadamc:
Overall, looks like it does what it's supposed to do. However, I felt that the documentation was confusing with the whole client/server paradigm.

Classes/common/CDTFetchChanges.m
gadamc: The suggestion here is to just quit if all callback block pointers are nil rather than running through the changes.

Classes/common/CDTFetchChanges.h
gadamc: Perhaps could also explain a bit what the 'sequence value' is. Or perhaps that explanation could be added to the introductory README.md.

mikerhodes: The standard TD_Datastore changes feed isn't publicly exposed, this is designed to be. So I think discussing the differences vs. the internal changes feed isn't applicable here.

I think you're right about expanding on sequence value and so on, however.

Classes/common/CDTFetchChanges.h
mikerhodes: The idea here is that this can be reused as an API on top of a remote changes feed eventually, to provide the same guarantees. This is also why the async API based on CloudKit is used.

Classes/common/CDTFetchChanges.h
mikerhodes: Exposing SequenceNumber from CDTDocumentRevisions would be a very bad idea, I think.

Classes/common/CDTFetchChanges.m
mikerhodes: I think it's reasonable to assume the user knows what they are doing?

Classes/common/CDTFetchChanges.h
mikerhodes: Addressed in 32a7cfa0d603190ac71a33d5102ca19b53b129bf

Classes/common/CDTFetchChanges.h
mikerhodes: I tried to address this in 32a7cfa0d603190ac71a33d5102ca19b53b129bf

Classes/common/CDTFetchChanges.h
mikerhodes: Yes, removed in 825569d3829b6058f9183fdcbf253636404f2109.

Classes/common/CDTFetchChanges.h
mikerhodes: I tweaked this slightly in 32a7cfa0d603190ac71a33d5102ca19b53b129bf. What do you think?

Classes/common/CDTFetchChanges.h
mikerhodes: Agreed, I should've corrected these names. See 0e7dc63cc887a074b315eb269b0b983f22e6a8f3.

Classes/common/CDTFetchChanges.h
mikerhodes: I've tried to improve this in 0e7dc63cc887a074b315eb269b0b983f22e6a8f3.

Classes/common/CDTFetchChanges.h
gadamc: It's been exposed for a while: https://github.com/cloudant/CDTDatastore/blob/45845-reusable-fetcher/Classes/common/CDTDocumentRevision.h#L36

Classes/common/CDTFetchChanges.h
```diff
@@ -0,0 +1,175 @@
...
+#import <Foundation/Foundation.h>
+
+@class CDTDatastore;
+@class CDTDocumentRevision;
+
+/**
+ CDTFetchChanges objects report changes to a database, that is, documents which have been
+ created, updated or deleted. This feature can be used to keep external datastores up to
+ date with changes in a database, for example to keep an external index of documents up
```
gadamc: It's unclear to me what this means... " This feature can be used to keep external datastores up to date with changes in a database". Is it to sync with remote Cloudant databases? 

I guess as an app developer, I'd like to see from the very first paragraph how it would be used in my app. The second sentence progresses quickly to a low level details. But maybe it's not envisioned to be used at a higher level. It's hard for me to say what I want you to write here, so I've no particular suggestion. I guess as an app developer, I'd like a better description of why I'd use this class.

Also, the second sentence could be broken up. "This feature can be used to.... .  For example, CDTQuery use CDTFetchChanges internally in the Query implementation to keep indexes up to date." (I suppose this will actually be implemented later?)

Classes/common/CDTFetchChanges.h
```diff
@@ -0,0 +1,175 @@
...
+ to date (we use CDTFetchChanges in the Query implementation, for example).
+
+ CDTFetchChanges provides callbacks for updated and created documents (documentChangedBlock)
+ and deleted documents (documentWithIDWasDeletedBlock). It also provides a completion block
+ for when all changes have been received and processed. The documentChangedBlock is passed the
```
gadamc: Since you mention the other blocks (which get a link in AppleDocs), perhaps add one here for the completed block. 

"It also provides a completion block for when all changes have been received and processed (fetchRecordChangesCompletionBlock)"

Classes/common/CDTFetchChanges.h
```diff
@@ -0,0 +1,175 @@
...
+
+ In addition, CDTFetchChanges is able to report on changes since a given point: the completion
+ block provides a sequence value which can be passed to future requests to have only changes
+ from that point to be reported. The sequence value should be treated as an opaque value, but
+ can be saved and used across application sessions. This allows for incremental updates in
+ response to changes in a database.
+
+ The CDTFetchChanges object reports all changed documents since the sequence value assigned
+ before -start is called.
+
```
gadamc: Suggested paragraph:

In addition, CDTFetchChanges is able to report on changes since a particular update to the database. Each change made to a CDTDatastore database is assigned a "sequence value". By setting the startSequenceValue, the CDTFetchChanges object will only report all changed documents since that update. This allows you to obtain changes to the database from a more recent update rather than receiving all changes to the database since its creation. Thus, your application can more easily respond to incremental changes in a database.

A sequence value can be obtained from the completion block, fetchRecordChangesCompletionBlock. When the completion block is called, the sequence value associated with the most recent update for the database is passed in via an argument (see below: newSequenceValue). This value can be used for future requests by assigning that value to startSequenceValue before calling -start. Then only changes from that point be reported. 

The sequence value should be treated as an opaque value, but can be saved and used across application sessions.

gadamc:
Any chance that a developer using this class will need to have the CDTDatastore* object passed in as an argument to the callback blocks? 

That is, would 
```
@property (nonatomic, copy) void ( ^ ) ( CDTDatastore* datastore, CDTDocumentRevision *revision ) documentChangedBlock
```
be useful?  (same for the deletion and completion blocks).

Might it allow for CDTFetchChanges objects to be passed to other objects within a developer's application? Or for other classes to construct the callback blocks?

Classes/common/CDTFetchChanges.h
mikerhodes: Fair enough, I wonder why we did that. I might deprecate and remove it if I can't see that many uses.

mikerhodes:
> Any chance that a developer using this class will need to have the CDTDatastore* object passed in as an argument to the callback blocks? 

The block probably already captures the datastore within its scope if the developer needs to use it.

Classes/common/CDTFetchChanges.h
mikerhodes: I tried to provide a concrete example:

>  CDTFetchChanges objects report changes to a database. The changes include documents which have been created, updated or deleted. Use this class to update your user interface efficiently when notified a replication is complete, for example, by updating elements for only those documents that have changed.

Is this better?

Classes/common/CDTFetchChanges.h
mikerhodes: I changed this slightly to try and make things clearer:

>  In addition, CDTFetchChanges is able to report on changes since a given point. The completion block provides a sequence value which can be passed to future requests to have only changes from that point reported. This allows your application to make incremental updates  in response to changes in a database.
 
> The sequence value should be treated as an opaque value, but can be saved and used across  application sessions.

However, I think going into more detail is too much for an introduction, where I think we should mostly be describing the features. I'd like to keep the sequence value as an opaque value, which is why I chose not to describe its derivation.

Do you think the details in the method and property descriptions need expanding to describe this adequately?

Classes/common/CDTFetchChanges.h
mikerhodes: Done in 8e0f703.

Classes/common/CDTFetchChanges.h
gadamc: Looks like it's used in a couple of tests, but that's all I could find.

Classes/common/CDTFetchChanges.h
gadamc: Much better. Now I can immediately see why I'd use this class -- after replication is a great moment to find out all the docs that have changed.

Classes/common/CDTFetchChanges.h
gadamc: What's the harm in exposing it? Maybe convert it to an NSString representation? 

Should we at least expose the last sequence? Perhaps ```CDTDatastore -getLastSequence``` would be useful. I am thinking about the app that has a pre-seeded datasbase. Otherwise, the developer would need to run CDTFetchObjects on the app's first launch to scroll through to the end to get the last sequence.

Classes/common/CDTFetchChanges.h
gadamc: Yeah, it looks fine. I think it's clear.

indisoluble:
+1

One last comment, I think we should create a task in the bug tracker to remember that we have to provide an easier way to launch this operation.

gadamc:
+1

mikerhodes:
Thanks @indisoluble and @gadamc for your comments; this thing's much better for them.

Merging.

mikerhodes:
Travis tests pass locally.

PR #121, opened 2015-04-08
Update indexes using CDTFetchChanges

mikerhodes:
This PR updates both old and new indexing code to use CDTFetchChanges, one commit per indexing implementation

The main things to look for are whether the logic I've created appears the same as the logic I'm replacing.

Some things to note:

1. The old code handled deletes and updates using the same code path and loop. This code separates that out as the changes fetcher blocks are executed separately for deletes and updates. However, the logic should be the same for a deleted or an updated document.
2. There are no extra tests in this PR as I'm assuming existing indexing tests cover this code satisfactorily.
3. Note a change in style prefering shadowing `self` to using `strongSelf`. I think this makes it less likely we'll accidentally create a retain cycle, particularly when copy/pasting code which uses `self`, as it's easy to miss changing one to `strongSelf`.

This PR is easier to review in Split diff mode.

reviewer @indisoluble 
reviewer @alfinkel 

BugId: 45846

### DIFF
--- a/Classes/common/Indexing/CDTIndexManager.m
+++ b/Classes/common/Indexing/CDTIndexManager.m
@@ -22,8 +22,7 @@
 #import "CDTDatastore.h"
 #import "CDTQueryBuilder.h"
 
-#import "TD_Database.h"
-#import "TD_Body.h"
+#import "CDTFetchChanges.h"
 
 #import "FMResultSet.h"
 #import "FMDatabase.h"
@@ -52,9 +51,7 @@ static const int VERSION = 1;
 
 - (BOOL)updateIndex:(CDTIndex *)index error:(NSError *__autoreleasing *)error;
 
-- (BOOL)updateIndex:(CDTIndex *)index
-            changes:(TD_RevisionList *)changes
-       lastSequence:(long *)lastSequence;
+- (BOOL)updateIndex:(CDTIndex *)index;
 
 - (BOOL)updateSchema:(int)currentVersion;
 
@@ -538,24 +535,7 @@ static const int VERSION = 1;
 
 - (BOOL)updateIndex:(CDTIndex *)index error:(NSError *__autoreleasing *)error
 {
-    BOOL success = TRUE;
-    TDChangesOptions options = {.limit = 10000,
-        .contentOptions = 0,                                
-        .includeDocs = NO,  // we only need the docIDs and sequences, body
-                            // is retrieved separately
-        .includeConflicts = FALSE,
-        .sortBySequence = TRUE};
-
-    TD_RevisionList *changes;
-    long lastSequence = [index lastSequence];
-
-    do {
-        changes = [[_datastore database] changesSinceSequence:lastSequence
-                                                      options:&options
-                                                       filter:nil
-                                                       params:nil];
-        success = success && [self updateIndex:index changes:changes lastSequence:&lastSequence];
-    } while (success && [changes count] > 0);
+    BOOL success = [self updateIndex:index];
 
     // raise error
     if (!success) {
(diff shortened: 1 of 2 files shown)


Classes/common/Indexing/CDTIndexManager.m
```diff
@@ -572,83 +552,157 @@ - (BOOL)updateIndex:(CDTIndex *)index error:(NSError *__autoreleasing *)error
...
+    fetcher.documentChangedBlock = ^(CDTDocumentRevision *revision) {
+        
+        CDTLogVerbose(CDTINDEX_LOG_CONTEXT, @"documentChangedBlock: <%@,%@>", 
+                      index.indexName, revision.docId);
+        
+        [updateBatch addObject:revision];
+        
+        if (updateBatch.count > 500) {
+            CDTIndexManager *self = weakSelf;
+            if (self) {
```
indisoluble: I thought that you were going to use `strongSelf` instead of shadowing `self`

Classes/common/query/CDTQIndexUpdater.m
```diff
@@ -138,71 +117,149 @@ - (BOOL)updateIndex:(NSString *)indexName
...
+    fetcher.documentChangedBlock = ^(CDTDocumentRevision *revision) {
+
+        CDTLogVerbose(CDTQ_LOGGING_CONTEXT, @"documentChangedBlock: <%@,%@>", indexName,
+                      revision.docId);
+
+        [updateBatch addObject:revision];
+
+        if (updateBatch.count > 500) {
+            CDTQIndexUpdater *self = weakSelf;
+            if (self) {
```
indisoluble: As before, I thought you were going to use `strongSelf`

indisoluble:
I'm concerned about when the [last sequence is updated](https://github.com/cloudant/CDTDatastore/pull/121/files#diff-4bdd68085cb14108b87d69a7602c063dR619). This is done only if all the previous batches are processed completely and successfully. This means that every batch of 500 records has to be [committed](https://github.com/cloudant/CDTDatastore/pull/121/files#diff-4bdd68085cb14108b87d69a7602c063dR667) to the index database; now imagine that the last batch fails, in that case we do not update the last sequence in the index but all the other batches are already committed to the database. I think that if we execute again this code, it is not going to fail; however until we do that the last sequence stored in the index is not synchronised with the real data in it.

Classes/common/query/CDTQIndexUpdater.m
mikerhodes: Sorry, my comment was meant to imply using `self`, but it was ambiguous. I changed it.

mikerhodes:
@indisoluble I think your concerns are valid, and in fact this is a change from the previous behaviour which did update the stored sequence number for each batch of changes. This isn't solvable using the current implementation of CDTFetchChanges, I think we'd need to implement `resultsLimit` and `moreComing` (with the same semantics as https://developer.apple.com/library/prerelease/ios/documentation/CloudKit/Reference/CKFetchRecordChangesOperation_class/index.html#//apple_ref/occ/instp/CKFetchRecordChangesOperation/resultsLimit) in order to do things this way.

Do you feel this should block this PR?

indisoluble:
[processUpdateBatch:](https://github.com/cloudant/CDTDatastore/pull/121/files#diff-4bdd68085cb14108b87d69a7602c063dR646), as well as [processDeleteBatch:](https://github.com/cloudant/CDTDatastore/pull/121/files#diff-4bdd68085cb14108b87d69a7602c063dR678), inits a transaction in a FMDBDatabaseQueue; would it be possible to pass to these methods a FMDBDatabase so they do not have to decide to commit the changes or rollback them?

mikerhodes:
@indisoluble 
> would it be possible to pass to these methods a FMDBDatabase so they do not have to decide to commit the changes or rollback them?

I purposely execute the transaction over the batch so we don't build up a huge transaction to commit. I'm not sure whether 500 docs is right here, but I'm loath to leave this open-ended.

alfinkel:
I agree with the concern @indisoluble has about the sequence number but I don't think it should block this PR.  Can we just create an FB ticket for this and we can handle it then?

indisoluble:
It is ok for me too.

alfinkel:
It might be worth noting in the FB ticket that the proposed change to the way the sequence number is handled needs to happen for both iOS and Android implementations of Query.

Other than that +1 for me.

indisoluble:
+1

mikerhodes:
Ace, thanks for the critical eyes @indisoluble and @alfinkel

PR #116, opened 2015-03-24
44626 remove legacy indexing code

alfinkel:
This PR removes all legacy indexing/query code as well as any references to the legacy code throughout CDTDatastore.  Details follow:

- Remove legacy indexing tests
- Update `SetUpDatastore` test to use new query logic
- Update the sample app project to use new query logic. `CDTAppDelegate` and `CDTViewController` were changed.  `CDTCompletedIndexer` was removed.
- Remove `indexing` group and all classes included
- Remove legacy references from `CloudantSync.h` 

BugID: 44626
reviewer @mikerhodes
reviewer @tomblench

PR #122, opened 2015-04-08
Port index encryption code to Query

indisoluble:
*What*
Encrypt the SQLite database used in a query index

*Why*
As the rest of the databases in the library, this one has to be encrypted too.

*How*
Use the same mechanism that we use for the other databases, i.e. we provide a encryption key to FMDB through a CDTEncryptionKeyProvider.

*Tests*
* Query index + SQLCipher
  1. Create an non-encrypted db with a non-encrypted datastore
  2. Create an encrypted db with an encrypted datastore
  3. Do not open an encrypted db with a non-encrypted datastore
  4. Do not open a non-encrypted db with an encrypted datastore
  5. Do not open an encrypted db with an encrypted datastore if they do not share the same key

* Query index + SQLite
  1. Create an non-encrypted db with a non-encrypted datastore
  2. Do not open an encrypted db with a non-encrypted datastore

reviewer @mikerhodes
reviewer @alfinkel

### DIFF
--- a/EncryptionTests/Encryption Tests/CDTQIndexManagerEncryptionTests.m
+++ b/EncryptionTests/Encryption Tests/CDTQIndexManagerEncryptionTests.m
@@ -1,0 +1,216 @@
+//
+//  CDTQIndexManagerEncryptionTests.m
+//  EncryptionTests
+//
+//  Created by Enrique de la Torre Fernandez on 08/04/2015.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+//  except in compliance with the License. You may obtain a copy of the License at
+//  http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software distributed under the
+//  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+//  either express or implied. See the License for the specific language governing permissions
+//  and limitations under the License.
+//
+
+#import <XCTest/XCTest.h>
+
+#import "CloudantSyncTests.h"
+#import "CloudantTests+EncryptionTests.h"
+#import "CDTDatastoreManager+EncryptionKey.h"
+#import "CDTEncryptionKeyNilProvider.h"
+#import "CDTHelperFixedKeyProvider.h"
+#import "FMDatabase+SQLCipher.h"
+
+#import "CDTQIndexManager.h"
+
+@interface CDTQIndexManagerEncryptionTests : CloudantSyncTests
+
+@end
+
+@implementation CDTQIndexManagerEncryptionTests
+
+- (void)setUp
+{
+    [super setUp];
+
+    // Put setup code here. This method is called before the invocation of each test method in the
+    // class.
+}
+
+- (void)tearDown
+{
+    // Put teardown code here. This method is called after the invocation of each test method in the
+    // class.
+
+    [super tearDown];
+}
+
+- (void)testCreateQueryIndexManagerWithEncryptionKeyNilProvider
+{
+    CDTEncryptionKeyNilProvider *provider = [CDTEncryptionKeyNilProvider provider];
+    CDTDatastore *datastore = [self.factory datastoreNamed:@"create_query_index_tests_nilprovider"
+                                 withEncryptionKeyProvider:provider
+                                                     error:nil];
+
+    NSError *err = nil;
+    CDTQIndexManager *im = [[CDTQIndexManager alloc] initUsingDatastore:datastore error:&err];
+
+    XCTAssertNotNil(im, @"indexManager is not nil");
+    XCTAssertNil(err, @"error has to be nil");
+}
+
+- (void)testCreateQueryIndexManagerWithEncryptionKeyNilProviderDoesNotCipherIndex
+{
+    // Create index
+    CDTEncryptionKeyNilProvider *provider = [CDTEncryptionKeyNilProvider provider];
+    CDTDatastore *datastore =
+        [self.factory datastoreNamed:@"create_query_index_tests_nilprovider_notcipher"
+            withEncryptionKeyProvider:provider
+                                error:nil];
+
+    __unused CDTQIndexManager *im = [[CDTQIndexManager alloc] initUsingDatastore:datastore error:nil];
+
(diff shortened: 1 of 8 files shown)


EncryptionTests/Encryption Tests/CDTQIndexManagerEncryptionTests.m
```diff
@@ -0,0 +1,213 @@
...
+    // Replace index
+    NSFileManager *defaultManager = [NSFileManager defaultManager];
+    NSString *indexPath = [CloudantTests pathForQueryIndexInDatastore:datastore];
+
+    [defaultManager removeItemAtPath:indexPath error:nil];
+
+    NSBundle *bundle = [NSBundle bundleForClass:[self class]];
+    NSString *assetPath = [bundle pathForResource:@"emptyencryptedindex" ofType:@"sqlite"];
+
+    [[NSFileManager defaultManager] copyItemAtPath:assetPath toPath:indexPath error:nil];
```
mikerhodes: I'm not really understanding why this test requires that we create the index database first before replacing it. Can we just create the database, then put in the canned database before initialising the IndexManager object?

Tests/Tests/CDTQIndexManagerEncryptionTests.m
```diff
@@ -0,0 +1,111 @@
...
+
+#import "CloudantSyncTests.h"
+#import "CloudantTests+EncryptionTests.h"
+#import "CDTDatastoreManager+EncryptionKey.h"
+#import "CDTEncryptionKeyNilProvider.h"
+#import "FMDatabase+SQLCipher.h"
+
+#import "CDTQIndexManager.h"
+
+@interface CDTQIndexManagerEncryptionTests : CloudantSyncTests
```
mikerhodes: Worth noting here that these tests are checking that the encryption code works when the test app is compiled w/out SQLCipher support? I was confused as to why tests appeared repeated, a comment would've hopefully helped me avoid that mistake.

Tests/Tests/CDTQIndexManagerEncryptionTests.m
```diff
@@ -0,0 +1,111 @@
...
+
+    [[NSFileManager defaultManager] copyItemAtPath:assetPath toPath:indexPath error:nil];
+
+    // Test
+    NSError *err = nil;
+    im = [[CDTQIndexManager alloc] initUsingDatastore:datastore error:&err];
+
+    XCTAssertNil(im, @"indexManager is nil");
+    XCTAssertNotNil(err, @"There is an error");
+}
```
mikerhodes: I'd like to see a test showing what should happen if someone passes a non-nil key provider when there's no SQLCipher available. What is this behaviour supposed to be?

EncryptionTests/Encryption Tests/CDTQIndexManagerEncryptionTests.m
indisoluble: I did that this way to avoid to have to create any non-existing folder. `NSFileManager:copyItemAtPath:toPath:error:` does not create a folder if it does not exists and it returns an error.

Tests/Tests/CDTQIndexManagerEncryptionTests.m
indisoluble: Yeah, the tests are repeated but in different projects. I might not be that clear here but I documented this in the previous [PR 112](https://github.com/cloudant/CDTDatastore/commit/8ee69974ee935f7dbe3113b110de837916edf805#diff-6a3371457528722a734f3c51d9238c13R148)

Tests/Tests/CDTQIndexManagerEncryptionTests.m
indisoluble: It is not possible to implement that test.

Check [this code](https://github.com/cloudant/CDTDatastore/commit/8ee69974ee935f7dbe3113b110de837916edf805#diff-6f72c4ee8250b96772fe8bc450cae5cbR72):
- A TD_Database can not be opened with a provider that returns a key
- A Datastore can not be created if a TD_Database can not be opened
- An index manager can not be created without a datastore.

Tests/Tests/CDTQIndexManagerEncryptionTests.m
mikerhodes: Okay.

EncryptionTests/Encryption Tests/CDTQIndexManagerEncryptionTests.m
mikerhodes: At some point we might unique CDTQIndexManager objects. If we do that you'll end up pulling the rug out from under someone as the index manager will probably have a database connection to the file you're removing, or perhaps you won't be able to do the copy at all, or perhaps the new index manager will keep using the old file as it's got an open handle so the test won't do what you expect.

Overall, I think creating the folder (that is, assuming the folder structure doesn't change), is a lower risk here.

Tests/Tests/CDTQIndexManagerEncryptionTests.m
mikerhodes: So your facts are:

a. IndexManager gets its key provider from the datastore.
b. A datastore won't accept a non-nil key provider if SQLCipher support isn't available.

_Therefore_: it's impossible to be in a state where the index manager receives a non-nil key provider if SQLCipher isn't available.

Accepted.

EncryptionTests/Encryption Tests/CDTQIndexManagerEncryptionTests.m
indisoluble: Fixed in f643c96

indisoluble:
Tests in `CDTQIndexManagerEncryptionTests` are identical to tests in `IndexManagerEncryptionTests`. In my last [commit](https://github.com/cloudant/CDTDatastore/commit/f643c96e6e6ca0b0e28e2f017ede0bf0cd8e063d), I only modified the tests in `CDTQIndexManagerEncryptionTests` because `IndexManagerEncryptionTests` is going to be deleted once [PR 116](https://github.com/cloudant/CDTDatastore/pull/116) is merged into `master` and `feature-encryption` is rebased on `master`.

mikerhodes:
Okay, looks fine, +1 from me.

alfinkel:
+1

PR #124, opened 2015-04-14
Add CDTDatastore tests to build script

rhyshort:
Added ability to run tests from the CDTDatastore workspace, in addition
to replication acceptance tests.

ReplicationAcceptance tests are only run when -nightly true is passed
into the build script. if -nightly false is passed in or nothing is
passed in the build script will only run the tests from CDTDatastore

reviewer: @tomblench 
reviewer: @mikerhodes

### DIFF
--- a/build.rb
+++ b/build.rb
@@ -6,8 +6,10 @@
 #  -platform-version iOS version to test agasint (only supported for iOS)
 #  -hardware the simultated hardware to test against eg iPhone 4S (only supported for iOS)
 #  -D* gets passed into build.
-#  
+#  -ra true | false  default is false flags if true it runs the replication acceptance tests
+#                    in addition to the normal unit tests
 #
+require 'fileutils'
 
 params = {}
 arg_is_value = false 
@@ -34,6 +36,7 @@ params["platform"] = "OSX" unless params["platform"]
 params["couch"] = "couchdb1.6" unless params["couch"]
 params["platform-version"] = "latest" unless params["platform-version"]
 params["hardware"] = "iPhone 4S" unless params["hardware"]
+params["ra"] = "false" unless params["ra"]
 
 #kill any docker container that may be running on the machine.
 #we don't want to effected by another failing build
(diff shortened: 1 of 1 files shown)


mikerhodes:
Seems like the option should be "-runReplicationAcceptance" rather than "-nightly". We could, I guess, have -nightly imply -runReplicationAcceptance, but they should be separate concepts.

rhyshort:
Maybe have both? -nightly as a shorthand for -runReplicationAcceptance? The short -nightly makes the jenkins config a little easier to read on the nightly builds

build.rb
```diff
@@ -86,9 +88,20 @@
 end
 
 if params["platform"] == "OSX"
-	system("xcodebuild -workspace ./ReplicationAcceptance/ReplicationAcceptance.xcworkspace -scheme 'RA_Tests_OSX' -destination 'platform=OS X' #{replication_options.join(" ")} test | xcpretty -r junit; exit ${PIPESTATUS[0]}")
+
+	system("xcodebuild -workspace CDTDatastore.xcworkspace -scheme 'Tests OSX' -destination 'platform=OS X' #{replication_options.join(" ")} test | xcpretty -r junit ; exit ${PIPESTATUS[0]}")
+
+	if(params["nightly"] == "true")
+		FileUtils.mv("build/reports/junit.xml","build/reports/osx.xml")
```
tomblench: I think you should have a FileUtils.mv after each system() call which does a test run.

Maybe the names could be $params["platform"]_ra.xml and $params["platform"]_unit.xml to distinguish replication acceptance results from unit test results.

tomblench:
I agree wtih @mikerhodes that RA or whatever makes more sense to me than nightly. I wouldn't like to see multiple synonyms for the same thing because that just gets more confusing.

Is there any use case where you would want to run RA but not unit tests?

build.rb
rhyshort: fixed in 78bfd89

rhyshort:
okay, I've changed to option to from `nightly` to `ra`

build.rb
```diff
@@ -6,8 +6,9 @@
 #  -platform-version iOS version to test agasint (only supported for iOS)
 #  -hardware the simultated hardware to test against eg iPhone 4S (only supported for iOS)
 #  -D* gets passed into build.
-#  
+#  -ra true | false  default is false flags if
```
tomblench: fix this comment - it doesn't tell us what the flag does

rhyshort: fixed in 9d3c00b

build.rb
```diff
@@ -6,8 +6,10 @@
 #  -platform-version iOS version to test agasint (only supported for iOS)
 #  -hardware the simultated hardware to test against eg iPhone 4S (only supported for iOS)
 #  -D* gets passed into build.
-#  
+#  -ra true | false  default is false flags if true it runs the replication acceptance tests
+# 					 in addition to the normal unit tests
```
tomblench: swap hard tabs for spaces

rhyshort: done 7b3caeb

tomblench:
OK +1

mikerhodes:
+1 looks good. Are there docs for the options somewhere?

rhyshort:
Not yet, my current plan is to have them both in the cloudant-sync repo, it'll include set up instructions for CI nodes too

rhyshort:
Travis is having intermittent errors, one of the builds through it was fine, going to merge anyway.

PR #125, opened 2015-04-14
Support cancelling of CDTFetchChanges

mikerhodes:
_What_

We now periodically check for the NSOperation's cancelled
property and bail early rather than always running the whole
changes fetch process.

_How_

Returning early from methods if the cancel state is set.

reviewer: @indisoluble 
reviewer: @gadamc

BugId: 46038

### DIFF
--- a/Classes/common/CDTFetchChanges.m
+++ b/Classes/common/CDTFetchChanges.m
@@ -20,6 +20,21 @@
 
 #import "TD_Database.h"
 
+@interface CDTFetchChanges ()
+
+@property (nonatomic, copy) void (^mDocumentChangedBlock)(CDTDocumentRevision *revision);
+
+@property (nonatomic, copy) void (^mDocumentWithIDWasDeletedBlock)(NSString *docId);
+
+@property (nonatomic, copy) void (^mFetchRecordChangesCompletionBlock)
+    (NSString *newSequenceValue, NSString *startSequenceValue, NSError *fetchError);
+
+@property (nonatomic, strong) CDTDatastore *mDatastore;
+
+@property (nonatomic, copy) NSString *mStartSequenceValue;
+
+@end
+
 @implementation CDTFetchChanges
 
 #pragma mark Initialisers
@@ -39,6 +54,14 @@
 
 - (void)main
 {
+    // Ensure our settings are not changed during execution
+    self.mDocumentChangedBlock = [self.documentChangedBlock copy];
+    self.mDocumentWithIDWasDeletedBlock = [self.documentWithIDWasDeletedBlock copy];
+    self.mFetchRecordChangesCompletionBlock = [self.fetchRecordChangesCompletionBlock copy];
+
+    self.mDatastore = self.datastore;
+    self.mStartSequenceValue = self.startSequenceValue;
+
     TDChangesOptions options = {.limit = 500,
         .contentOptions = 0,                                
         .includeDocs = NO,  // we only need the docIDs and sequences, body is retrieved separately
@@ -46,27 +69,28 @@
         .sortBySequence = TRUE};
     
     TD_RevisionList *changes;
-    SequenceNumber lastSequence = [_startSequenceValue longLongValue];
+    SequenceNumber lastSequence = [self.mStartSequenceValue longLongValue];
 
     do {
         if (self.cancelled) {
             return;  // Bail early, don't call completion block
         }
 
-        changes = [[_datastore database] changesSinceSequence:lastSequence
-                                                      options:&options
-                                                       filter:nil
-                                                       params:nil];
+        changes = [[self.mDatastore database] changesSinceSequence:lastSequence
+                                                           options:&options
+                                                            filter:nil
+                                                            params:nil];
         lastSequence = [self notifyChanges:changes startingSequence:lastSequence];
     } while (changes.count > 0);
 
-    void (^f)(NSString *nsv, NSString *ssv, NSError *fe) = self.fetchRecordChangesCompletionBlock;
+    void (^f)(NSString *nsv, NSString *ssv, NSError *fe) = self.mFetchRecordChangesCompletionBlock;
     if (f) {
         // Try our best to avoid calling the completion block if cancelled is set:
         //  - after the check in the loop above 
         //  - where there are no remaining changes, so we fall through to here
         if (!self.cancelled) {
-            f([[NSNumber numberWithLongLong:lastSequence] stringValue], _startSequenceValue, nil);
+            f([[NSNumber numberWithLongLong:lastSequence] stringValue], self.mStartSequenceValue,
+              nil);
         }
     }
 }
@@ -97,7 +121,7 @@
     // that are updated rather than deleted, we need to be sure we index the
     // winning revision. This loop gets those revisions.
     NSMutableDictionary *updatedRevisions = [NSMutableDictionary dictionary];
-    for (CDTDocumentRevision *rev in [_datastore getDocumentsWithIds:[changes allDocIDs]]) {
+    for (CDTDocumentRevision *rev in [self.mDatastore getDocumentsWithIds:[changes allDocIDs]]) {
         if (rev != nil && !rev.deleted) {
             updatedRevisions[rev.docId] = rev;
         }
@@ -110,14 +134,12 @@
 
         CDTDocumentRevision *updatedRevision;
         if ((updatedRevision = updatedRevisions[change.docID]) != nil) {
-            void (^dcb)(CDTDocumentRevision *r) = self.documentChangedBlock;
-            if (dcb) {
-                dcb(updatedRevision);
+            if (self.mDocumentChangedBlock) {
+                self.mDocumentChangedBlock(updatedRevision);
             }
         } else {
-            void (^ddb)(NSString *docId) = self.documentWithIDWasDeletedBlock;
-            if (ddb) {
-                ddb(change.docID);
+            if (self.mDocumentWithIDWasDeletedBlock) {
+                self.mDocumentWithIDWasDeletedBlock(change.docID);
             }
         }
         
--- a/Classes/common/CDTFetchChanges.m
+++ b/Classes/common/CDTFetchChanges.m
@@ -62,7 +62,12 @@
 
     void (^f)(NSString *nsv, NSString *ssv, NSError *fe) = self.fetchRecordChangesCompletionBlock;
     if (f) {
-        f([[NSNumber numberWithLongLong:lastSequence] stringValue], _startSequenceValue, nil);
+        // Try our best to avoid calling the completion block if cancelled is set:
+        //  - after the check in the loop above 
+        //  - where there are no remaining changes, so we fall through to here
+        if (!self.cancelled) {
+            f([[NSNumber numberWithLongLong:lastSequence] stringValue], _startSequenceValue, nil);
+        }
     }
 }
 
--- a/Classes/common/CDTFetchChanges.m
+++ b/Classes/common/CDTFetchChanges.m
@@ -55,10 +55,9 @@
 - (void)main
 {
     // Ensure our settings are not changed during execution
-    self.mDocumentChangedBlock = [self.documentChangedBlock copy];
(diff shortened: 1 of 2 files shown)


gadamc:
@indisoluble probably has more NSOperations experience than me, but it looks good to me. 
When will direct tests be added? (CDTFetchChanges is tested indirectly via the Index tests, I presume.)

gadamc:
After reading a bit more on NSOperations, I have a question. 

Is CDTFetchChanges re-usable? From what I understand, after a CDTFetchChanges object completes the `-main` function, the `isFinished` property will be set to `YES` and will no longer call `-main` on subsequent calls to `-start`. From the documentation (https://developer.apple.com/library/ios/documentation/Cocoa/Reference/NSOperation_class/index.html#//apple_ref/occ/instm/NSOperation/start)

> The default implementation of this method updates the execution state of the operation and calls the receiver’s main method. This method also performs several checks to ensure that the operation can actually run. For example, if the receiver was cancelled or is already finished, this method simply returns without calling main. 

Additionally... the next few lines in this documentation describe exceptions thrown by NSOperation. Do we handle those properly? Our `-main` method doesn't catch anything. (Am I reading the documentation correctly?)

> (In OS X v10.5, this method throws an exception if the operation is already finished.) If the operation is currently executing or is not ready to execute, this method throws an NSInvalidArgumentException exception. In OS X v10.5, this method catches and ignores any exceptions thrown by your main method automatically. In OS X v10.6 and later, exceptions are allowed to propagate beyond this method. You should never allow exceptions to propagate out of your main method.

mikerhodes:
@gadamc Yes, you're right about exceptions. This isn't related to this PR, but I should go re-read the NSOperation docs.

I also intend to add tests but have been pulled in a bunch of different directions so haven't got to this as quickly as I'd like after getting the merge in for the changes reading bug fix.

indisoluble:
@gadamc You are absolutely right, a NSOperation  can only be used once.

I've re-read the documentation and I haven't found any reference to set `startSequenceValue` and call again `-start`. I think this was my mistake, I said that `CDTFetchChanges` (as well as its Apple counterpart `CKFetchRecordsOperation`) could be re-used in [this comment](https://github.com/cloudant/CDTDatastore/commit/902d9f99405722c3739047188bc38f5bddcf77e9#commitcomment-10545514).

Classes/common/CDTFetchChanges.m
```diff
@@ -91,7 +100,10 @@ - (SequenceNumber)notifyChanges:(TD_RevisionList *)changes
     }
```
indisoluble: This comment is about line [96](https://github.com/cloudant/CDTDatastore/pull/125/files#diff-bac15dfde454859cd8e3a8dc04fcf29fR96) but I can not make the comment right below it.

Let imagine this case:
1. The operation is not cancelled and pass the first validation(line [52](https://github.com/cloudant/CDTDatastore/pull/125/files#diff-bac15dfde454859cd8e3a8dc04fcf29fR52))
2. It gets all the changes (line [56](https://github.com/cloudant/CDTDatastore/pull/125/files#diff-bac15dfde454859cd8e3a8dc04fcf29fR56)). To do that, it reads from **disk** through the **serial queue** managed by the FMDatabase, therefore it takes a while.
3. The operation is cancelled
4. In line [96](https://github.com/cloudant/CDTDatastore/pull/125/files#diff-bac15dfde454859cd8e3a8dc04fcf29fR96), we get the documents with the specified IDs. As before, to do that we have to read from **disk** through the same **serial queue**, this is slow and costly.
5. Finally, in line [103](https://github.com/cloudant/CDTDatastore/pull/125/files#diff-bac15dfde454859cd8e3a8dc04fcf29fR103), we check if the NSOperation is cancelled.

I think it would be a good idea to check before and after if the NSOperation is cancelled, for example:

``` objc
    if (self.cancelled) {
        return lastSequence;
    }

    NSArray *docs = [_datastore getDocumentsWithIds:[changes allDocIDs]];

    NSMutableDictionary *updatedRevisions = [NSMutableDictionary dictionary];
    for (CDTDocumentRevision *rev in docs) {
        if (self.cancelled) {
            return lastSequence;
        }

        if (rev != nil && !rev.deleted) {
            updatedRevisions[rev.docId] = rev;
        }
    }
```

Or it might be clearer to add this validation right after line [56](https://github.com/cloudant/CDTDatastore/pull/125/files#diff-bac15dfde454859cd8e3a8dc04fcf29fR56). However I recommend to keep the validation inside the loop that iterates through the changed docs (we can not know how long it is going to take):

``` objc
    for (CDTDocumentRevision *rev in docs) {
        if (self.cancelled) {
            return lastSequence;
        }
        ...
    }
```

gadamc:
@indisoluble I don't think there's an explicit instruction in the documentation to set `startSequenceValue` and then call `-start`, but I think it's definitely implied by lines:
https://github.com/cloudant/CDTDatastore/blob/master/Classes/common/CDTFetchChanges.h#L34

https://github.com/cloudant/CDTDatastore/blob/master/Classes/common/CDTFetchChanges.h#L60-L62

https://github.com/cloudant/CDTDatastore/blob/master/Classes/common/CDTFetchChanges.h#L156.

These lines reference a "previous fetch". I think it would be more clear if they said something like "obtained from the completion block of a previous CDTFetchChanges object", or "previous CDTFetchChanges object". Also, I think we should put at the top of the documentation for the class, as a reminder, that this is a single-use object because it inherits from NSOperation. (Not everybody is going to remember that NSOperations are single-use could be a pain point for somebody. I don't see why we just wouldn't remind them here in the class description.)

I've opened a new FB ticket (46342)

indisoluble:
I agree, the reminder would be a good idea ... I totally forgot it :)

Classes/common/CDTFetchChanges.m
mikerhodes: Addressed in f5222af. See note in commit as to why I didn't add a cancelled check within the loop over the doc array returned from `getDocumentsWithIds`.

mikerhodes:
@gadamc I updated the docs to replace "request" and "fetch" with "operation" which implies to me a new operation object.

Classes/common/CDTFetchChanges.m
```diff
@@ -39,6 +54,14 @@ - (instancetype)initWithDatastore:(CDTDatastore *)datastore
 
 - (void)main
 {
+    // Ensure our settings are not changed during execution
+    self.mDocumentChangedBlock = [self.documentChangedBlock copy];
+    self.mDocumentWithIDWasDeletedBlock = [self.documentWithIDWasDeletedBlock copy];
+    self.mFetchRecordChangesCompletionBlock = [self.fetchRecordChangesCompletionBlock copy];
```
indisoluble: The new properties specify the attribute `copy`, I do not think you need to copy the original block before assigning it (check [apple doc](https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/EncapsulatingData/EncapsulatingData.html), under title *Copy Properties Maintain Their Own Copies*)

Classes/common/CDTFetchChanges.m
```diff
@@ -46,19 +69,29 @@ - (void)main
...
+
+        changes = [[self.mDatastore database] changesSinceSequence:lastSequence
+                                                           options:&options
+                                                            filter:nil
+                                                            params:nil];
         lastSequence = [self notifyChanges:changes startingSequence:lastSequence];
     } while (changes.count > 0);
 
-    void (^f)(NSString *nsv, NSString *ssv, NSError *fe) = self.fetchRecordChangesCompletionBlock;
+    void (^f)(NSString *nsv, NSString *ssv, NSError *fe) = self.mFetchRecordChangesCompletionBlock;
```
indisoluble: This assignment does not look like necessary anymore

Classes/common/CDTFetchChanges.m
indisoluble: My concern about the loop was that the number of revisions could be really high so looping through all of them would take a while but if that ever happens, the replication will also take ages and that is a totally different problem.

Classes/common/CDTFetchChanges.m
mikerhodes: I kept this one because I thought it make the if a bit clearer, but I think actually it doesn't now I come back to it.

Classes/common/CDTFetchChanges.m
mikerhodes: Fixed in 629c0b6

Classes/common/CDTFetchChanges.m
mikerhodes: Fixed in 629c0b6

Classes/common/CDTFetchChanges.m
```diff
@@ -46,19 +68,29 @@ - (void)main
...
-    void (^f)(NSString *nsv, NSString *ssv, NSError *fe) = self.fetchRecordChangesCompletionBlock;
-    if (f) {
-        f([[NSNumber numberWithLongLong:lastSequence] stringValue], _startSequenceValue, nil);
+    if (self.mFetchRecordChangesCompletionBlock) {
+        // Try our best to avoid calling the completion block if cancelled is set:
+        //  - after the check in the loop above 
+        //  - where there are no remaining changes, so we fall through to here
+        if (!self.cancelled) {
+            self.mFetchRecordChangesCompletionBlock(
+                [[NSNumber numberWithLongLong:lastSequence] stringValue], self.mStartSequenceValue,
```
indisoluble: There is a strong reference cycle here: `self.mStartSequenceValue`

mikerhodes: This is a method call not a block definition.

mikerhodes:
@indisoluble and @gadamc please check this one again.

Classes/common/CDTFetchChanges.m
mikerhodes: This loops only gets a maximum of 500 documents as the call to get the list of database changes uses blocks of 500, which are then passed to this method.

As a second point, this code isn't used for replications.

indisoluble: Right. Thanks.

indisoluble:
+1

Classes/common/CDTFetchChanges.m
```diff
@@ -39,6 +54,13 @@ - (instancetype)initWithDatastore:(CDTDatastore *)datastore
 
 - (void)main
 {
+    // Ensure our settings are not changed during execution
+    self.mDocumentChangedBlock = self.documentChangedBlock;
+    self.mDocumentWithIDWasDeletedBlock = self.documentWithIDWasDeletedBlock;
+    self.mFetchRecordChangesCompletionBlock = self.fetchRecordChangesCompletionBlock;
+    self.mDatastore = self.datastore;
+    self.mStartSequenceValue = self.startSequenceValue;
```
gadamc: Maybe this is too much of an edge case scenario... 

But this would prevent somebody from setting the "completion" block based upon the docs it gets in the "changed" and "deleted" blocks. 

Perhaps we could only set `self.mFetchRecordChangesCompletionBlock = self.fetchRecordChangesCompletionBlock` just before the completion block is called. (Or not set it at all...)

Then somebody could have a "changed" block that actually sets the "completion" block depending upon the data it receives. 

Like I said... might be an edge case, but potentially an interesting application of CDTFetchChanges.

gadamc: On the other hand, I now realize one could write a "completed" block that simply reacts differently based upon the accumulated knowledge received during the 'changes' and 'deleted' blocks... which means there wouldn't be a need to set a different "completion" block. The users's completion block would contain the logic.

gadamc: so... +1

mikerhodes: Docs also [state](https://github.com/cloudant/CDTDatastore/blob/46038-cancellable-fetchchanges/Classes/common/CDTFetchChanges.h#L136-L137):

> If you intend to use this block to process results, set it before executing the operation or submitting the operation object to a queue.

mikerhodes:
Thanks @indisoluble @gadamc

PR #126, opened 2015-04-21
45570 Generate, retrieve & store a key: AES & PBKDF2

indisoluble:
*What:*
Add classes to apply AES to a NSData and generate a key with PBKDF2.

*Why:*
This PR is part of a bigger development to provide a mechanism to generate a key and safely store it.

*How:*
These classes are essentially a wrapper around the methods offered by Security framework.

*Tests:*
Only to be sure that we get what we expect, using some hardcoded values we check that the results we get are as expected.

reviewer @mikerhodes 
reviewer @rhyshort


Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils+AES.h
```diff
@@ -0,0 +1,47 @@
...
+ Encrypts a buffer by using a key and an Initialization Vector (IV).
+
+ @param data The data to encrypt
+ @param key The key used for encryption
+ @param iv The IV used for encryption
+
+ @return The encrypted data
+ */
++ (NSData *)doEncrypt:(NSData *)data withKey:(NSData *)key iv:(NSData *)iv;
+
```
rhyshort: I think `doEncrypt:withKey:iv:` and `doDecrypt:withKey:iv:`  should be called `encrypt:withKey:iv:` and `decrypt:withKey:iv:` respectively

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils+AES.m
```diff
@@ -0,0 +1,80 @@
...
+    NSAssert((cryptorStatus == kCCSuccess) && cryptor, @"Cryptographic context not created");
+
+    // Encrypt
+    size_t dataOutSize = CCCryptorGetOutputLength(cryptor, (size_t)[data length], true);
+    void *dataOut = malloc(dataOutSize);
+
+    size_t dataOutPartialSize = 0;
+    cryptorStatus = CCCryptorUpdate(cryptor, [data bytes], (size_t)[data length], dataOut,
+                                    dataOutSize, &dataOutPartialSize);
+    NSAssert(cryptorStatus == kCCSuccess, @"Data not encrypted (update)");
```
rhyshort: We probably want to fail gracefully here, rather than using assert

rhyshort:
General comment, they are all missing Cloudant/IBM copyright eg `Copyright IBM Inc. 2015` See some of the other files for the exact wording

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils+AES.h
rhyshort: Now that I've seen `decrypt` etc is used for this class already being used, so ignore the comment above

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils.m
```diff
@@ -0,0 +1,151 @@
...
+@end
+
+@implementation CDTEncryptionKeychainUtils
+
+#pragma mark - Public class methods
++ (NSData *)generateRandomBytesInBufferWithLength:(NSUInteger)length
+{
+    uint8_t randBytes[length];
+    
+    int rc = SecRandomCopyBytes(kSecRandomDefault, (size_t)length, randBytes);
```
mikerhodes: Do we need to check `length` doesn't exceed the max for `size_t`?

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils.m
```diff
@@ -0,0 +1,151 @@
...
+
+    if (![pass isKindOfClass:[NSString class]] || (pass.length < 1)) {
+        [NSException raise:CDTENCRYPTION_KEYCHAIN_UTILS_ERROR_KEYGEN_LABEL
+                    format:CDTENCRYPTION_KEYCHAIN_UTILS_ERROR_KEYGEN_MSG_EMPTY_PASSWORD];
+    }
+
+    if (![salt isKindOfClass:[NSData class]] || (salt.length < 1)) {
+        [NSException raise:CDTENCRYPTION_KEYCHAIN_UTILS_ERROR_KEYGEN_LABEL
+                    format:CDTENCRYPTION_KEYCHAIN_UTILS_ERROR_KEYGEN_MSG_EMPTY_SALT];
+    }
```
mikerhodes: Reorder these checks to be in the same order as the arguments.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils+PBKDF2.h
```diff
@@ -0,0 +1,43 @@
...
+ @param salt The salt that is used to generate the key
+ @param iterations The number of iterations that is passed to the key generation algorithm
+ @param length Size of the key
+
+ @return The generated key
+ */
++ (NSData *)derivePassword:(NSString *)password
+                  withSalt:(NSData *)salt
+                iterations:(NSUInteger)iterations
+                    length:(NSUInteger)length;
```
mikerhodes: `+deriveKeyOfLength:fromPassword:usingSalt:iterations:`

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils+PBKDF2.m
```diff
@@ -0,0 +1,48 @@
...
+                iterations:(NSUInteger)iterations
+                    length:(NSUInteger)length
+{
+    NSData *passData = [password dataUsingEncoding:NSUTF8StringEncoding];
+    
+    NSMutableData *derivedKey = [NSMutableData dataWithLength:length];
+
+    int retVal =
+        CCKeyDerivationPBKDF(kCCPBKDF2, passData.bytes, passData.length, salt.bytes, salt.length,
+                             kCCPRFHmacAlgSHA1, (uint)iterations, derivedKey.mutableBytes, length);
```
mikerhodes: Is SHA1 good enough here? Why not `kCCPRFHmacAlgSHA512`?

mikerhodes: SHA1 looks fine in this particular use.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils+AES.m
mikerhodes: Failures here are memory failures or bad params. One's a pretty fundamental failure condition the other programmer error. @indisoluble it'd be good to understand why you chose asserts for the CC* return values.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils+AES.m
```diff
@@ -0,0 +1,80 @@
...
+                 (key.length == kCCKeySizeAES256),
+             @"Key length must be appropriate for the selected operation and algorithm.");
+    NSAssert(iv.length == kCCBlockSizeAES128,
+             @"IV length must be the same length as the selected algorithm's block size");
+
+    // Generate context
+    CCCryptorRef cryptor = NULL;
+    CCCryptorStatus cryptorStatus =
+        CCCryptorCreate(operation, kCCAlgorithmAES, kCCOptionPKCS7Padding, key.bytes, key.length,
+                        iv.bytes, &cryptor);
```
mikerhodes: Describe where the choice for `kCCOptionPKCS7Padding` came from.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils+AES.m
```diff
@@ -0,0 +1,80 @@
...
+    NSAssert((cryptorStatus == kCCSuccess) && cryptor, @"Cryptographic context not created");
+
+    // Encrypt
+    size_t dataOutSize = CCCryptorGetOutputLength(cryptor, (size_t)[data length], true);
+    void *dataOut = malloc(dataOutSize);
```
mikerhodes: Do the bytes need clearing?

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils+AES.m
```diff
@@ -0,0 +1,80 @@
...
+    dataOutTotalSize += dataOutPartialSize;
+
+    // Free context
+    CCCryptorRelease(cryptor);
+
+    // Return
+    NSData *processedData = [NSData dataWithBytesNoCopy:dataOut length:dataOutTotalSize];
+
+    return processedData;
+}
```
mikerhodes: Looks right to me. My only worry is all the stuff in the docs about bad alignment during decryption -- do we know that won't happen here?

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils+AES.m
rhyshort: oh right okay

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils+AES.h
```diff
@@ -0,0 +1,47 @@
...
+/**
+ Decrypts a buffer by using a key and an Initialization Vector (IV).
+
+ @param data The encrypted data to decrypt
+ @param key The key used for decryption
+ @param iv The IV used for decryption
+
+ @return The decrypted data
+ */
++ (NSData *)doDecrypt:(NSData *)data withKey:(NSData *)key iv:(NSData *)iv;
```
mikerhodes: Why did you choose not to include the cipher name (e.g., AES or PBKDF2) in the function names which use specific ciphers?

Tests/Tests/Encryption/Keychain/CDTEncryptionKeychainUtilsAESTests.m
```diff
@@ -0,0 +1,181 @@
...
+    self.key = nil;
+    self.iv = nil;
+
+    [super tearDown];
+}
+
+- (void)testDoEncrypt
+{
+    NSString *txt = @"1234567890";
+    NSString *expectedResult = @"ExUL75zXTBiQKHAqYeR3Glt+EXMR25qmNTdeToHdA40=";
```
mikerhodes: Name with more meaningful names:

- `txt` -> `plainText`
- `expectedResult` -> `expectedCiphertext`

Helps because below you reuse the name `txt` for the input ciphertext rather than the plain text.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils.h
```diff
@@ -0,0 +1,76 @@
...
+ @param salt The salt that is used to generate the key
+ @param iterations The number of iterations that is passed to the key generation algorithm
+ @param length Size of the key
+
+ @return The generated key
+ */
++ (NSData *)generateKeyWithPassword:(NSString *)pass
+                               salt:(NSData *)salt
+                         iterations:(NSInteger)iterations
+                             length:(NSUInteger)length;
```
mikerhodes: `+generateKeyOfLength:usingPassword:withSalt:iterations:`

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils+AES.h
mikerhodes: Is it worth noting in the docs for the category methods that they assume correct inputs (and refer people to use the ones from the main class file instead which do the checks?).

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils+AES.m
indisoluble: Well, I do not expect anyone to use these methods directly, instead they will use them through the `CDTEncryptionKeychainManager`. To say in another way, it is unlikely any of the failures described before to happen but just to be sure, I checked the results with NSAssert given that is only an extra line.

To say in another way, I wanted to check the results but

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils.m
indisoluble: Like this:

```objc
NSAssert((size_t)length <= SIZE_T_MAX, @"length %lu out of bound", (unsigned long)length);
```

This method is only used by `CDTEncryptionKeychainManager`, I do not think we need a more complicate validation.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils.m
indisoluble: Fixed in f255986

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils+PBKDF2.h
indisoluble: Fixed in 39da160

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils+AES.m
indisoluble: The other option available is `kCCOptionECBMode`. If I choose this one, the IV is ignored and I want to use it: `This parameter is ignored if ECB mode is used or if a stream cipher algorithm is selected`.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils+AES.m
indisoluble: I'm going to use only the bytes [I did write](https://github.com/cloudant/CDTDatastore/pull/126/files#diff-29e72c1605b8d898b1f99861bf3a4bb2R75) but clearing the buffer it is a good practice.

indisoluble: Fixed in 5c42d33

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils+AES.m
indisoluble: I do not think we have to worry about that because in our case we only call once to `CCCryptorUpdate`. We do not have to worry about aligning the input buffer because we use the entire buffer.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils+AES.h
indisoluble: I did not include the cipher name in the function names because it is already in the category.

indisoluble: Fixed in 6d0c190 ... I think, have a look to be sure.

Tests/Tests/Encryption/Keychain/CDTEncryptionKeychainUtilsAESTests.m
indisoluble: Fixed in 22a134b

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils.h
indisoluble: Fixed in bb1ba54

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils.m
indisoluble: Fixed in 7bb8c2a

indisoluble:
Added copyright to new classes in cdda2a8

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils+AES.m
mikerhodes: Sounds like we need not worry here. Passing incorrect input data to decrypt is probably a programmer error (either not sending the right buffer, or not checking a file has been read correctly), so we can assert failure on alignment issues.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils+AES.h
mikerhodes: Yup, that's fine.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainUtils+AES.m
indisoluble: @mikerhodes: Thanks for pointing out that using kCCOptionPKCS7Padding we do not need to worry about aligning the input buffer to a block size.

mikerhodes: Looking further, we figured that this call enables AES in CBC mode with padding automatically appended if needed. This is appropriate for our needs.

mikerhodes:
Okay I think we are good now. +1.

rhyshort:
looks fine to me +1

PR #128, opened 2015-04-21
45570 Generate, retrieve & store a key: Manager to generate & retrieve a key

indisoluble:
*What:*
Define a manager able to create a DPK (Data Protection Key) and store and retrieve this DPK from keychain.

*Why:*
This PR is the last one in a bigger development to provide a mechanism to generate a key and safely store it.

*How:*
The manager will be based on the classes implemented before to apply AES, PBKDF2 and read from/write to the keychain.

*Tests:*
* Clear the data stored in the keychain
* Check there is data in the keychain
* Get data from the keychain
* Check that the number of iterations stored in the keychain is right
* Check that the IV in the keychain has the correct size
* Validate the steps followed to retrieve a key
* Check that is not possible to generate the key twice
* Validate the steps followed to generate a key

reviewer @mikerhodes 
reviewer @rhyshort 

**NOTE:** Wait to review this PR until PR #134  is closed. This branch will be rebased and there will be only one commit to review.

### DIFF
--- a/Classes/common/Encryption/Keychain/CDTEncryptionKeychainConstants.h
+++ b/Classes/common/Encryption/Keychain/CDTEncryptionKeychainConstants.h
@@ -1,0 +1,37 @@
+//
+//  CDTEncryptionKeychainConstants.h
+//  CloudantSync
+//
+//  Created by Enrique de la Torre Fernandez on 09/04/2015.
+//  Copyright (c) 2015 IBM Cloudant. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+//  except in compliance with the License. You may obtain a copy of the License at
+//  http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software distributed under the
+//  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+//  either express or implied. See the License for the specific language governing permissions
+//  and limitations under the License.
+//
+
+#ifndef _CDTEncryptionKeychainConstants_h
+#define _CDTEncryptionKeychainConstants_h
+
+#import <CommonCrypto/CommonCryptor.h>
+
+// Define the version of the implementation used to cipher/decipher DPKs
+#define CDTENCRYPTION_KEYCHAIN_VERSION @"1.0"
+
+// Size (bytes) of the DPK (Data Protection Key)
+#define CDTENCRYPTION_KEYCHAIN_ENCRYPTIONKEY_SIZE 32
+
+// PBKDF2: Size (bytes) of the salt value used to derive a user-provided password
+#define CDTENCRYPTION_KEYCHAIN_PBKDF2_SALT_SIZE 32
+
+// PBKDF2: Number of times the derivation process is applied to a user-provided password
+#define CDTENCRYPTION_KEYCHAIN_PBKDF2_ITERATIONS (NSInteger)10000
+
+// AES: Size (bytes) of the key used to encrypt/decrypt a DPK
+#define CDTENCRYPTION_KEYCHAIN_AES_KEY_SIZE kCCKeySizeAES256
+
+#endif
--- a/Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager+Internal.h
+++ b/Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager+Internal.h
@@ -1,0 +1,43 @@
+//
+//  CDTEncryptionKeychainManager+Internal.h
+//  CloudantSync
+//
+//  Created by Enrique de la Torre Fernandez on 15/04/2015.
+//  Copyright (c) 2015 IBM Cloudant. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+//  except in compliance with the License. You may obtain a copy of the License at
+//  http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software distributed under the
+//  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+//  either express or implied. See the License for the specific language governing permissions
+//  and limitations under the License.
+//
+
+#import "CDTEncryptionKeychainManager.h"
+
+/**
+ This category is only for testing purposes. It presents some of the private methods used by
+ CDTEncryptionKeychainManager to complete its job.
+ */
+@interface CDTEncryptionKeychainManager (Internal)
+
+- (BOOL)validateEncryptionKeyData:(CDTEncryptionKeychainData *)data;
+
+- (NSData *)generateDpk;
+
+- (CDTEncryptionKeychainData *)keychainDataToStoreDpk:(NSData *)dpk
+                                encryptedWithPassword:(NSString *)password;
+
+- (NSData *)generatePBKDF2Salt;
+- (NSData *)pbkdf2DerivedKeyForPassword:(NSString *)pass
+                                   salt:(NSData *)salt
+                             iterations:(NSInteger)iterations
+                                 length:(NSUInteger)length;
+
+- (NSData *)generateAESIv;
+
+- (NSData *)encryptDpk:(NSData *)dpk usingAESWithKey:(NSData *)key iv:(NSData *)iv;
+- (NSData *)decryptDpk:(NSData *)cipheredDpk usingAESWithKey:(NSData *)key iv:(NSData *)iv;
+
+@end
--- a/Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.h
+++ b/Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.h
@@ -1,0 +1,85 @@
+//
+//  CDTEncryptionKeychainManager.h
+//  CloudantSync
+//
+//  Created by Enrique de la Torre Fernandez on 09/04/2015.
+//  Copyright (c) 2015 IBM Cloudant. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+//  except in compliance with the License. You may obtain a copy of the License at
+//  http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software distributed under the
+//  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+//  either express or implied. See the License for the specific language governing permissions
+//  and limitations under the License.
+//
+
+#import <Foundation/Foundation.h>
+
+#import "CDTEncryptionKeychainStorage.h"
+
+/**
+ Use this class to generate a Data Protection Key (DPK), i.e. a strong password that can be used
+ later on for other purposes like encrypting a database.
+ 
+ The generated DPK is automatically encrypted and saved to the keychain, for this reason, it is
+ neccesary to provide a password to generate and retrieve the DPK. On a high level, this is done as
+ follow:
+ - Generate a DPK as a 32 bytes buffer with secure random values.
+ - Generate a salt as a 32 bytes buffer with secure random values.
+ - Use PBKDF2 to derive a key based on the user-provided password and the salt.
+ - Generate an initialization vector (IV) as a 16 bytes buffer with secure random values.
+ - Use AES to cipher the DPK with the key and the IV.
+ - Return the DPK and save the encrypted version to the keychain.
+ */
+@interface CDTEncryptionKeychainManager : NSObject
+
+- (instancetype)init UNAVAILABLE_ATTRIBUTE;
+
+/**
+ Initialise a manager with a CDTEncryptionKeychainStorage instance.
+ 
+ A CDTEncryptionKeychainStorage binds an entry in the keychain to an identifier. The data
+ protection key (DPK) saved to the keychain by this class will therefore be bound to the storage's
+ identifier. To save different DPKs (say for different users of your application), create multiple
+ managers using storages initialised with different identifiers.
+ 
+ @param storage Storage instance to save DPKs to the keychain
+ 
+ @see CDTEncryptionKeychainStorage
+ */
+- (instancetype)initWithStorage:(CDTEncryptionKeychainStorage *)storage;
+
+/**
+ Returns the decrypted Data Protection Key (DPK) from the keychain.
+ 
+ @param password Password used to decrypt the DPK
+ 
+ @return The DPK
+ */
+- (NSData *)loadKeyUsingPassword:(NSString *)password;
+
+/**
+ Generates a Data Protection Key (DPK), encrypts it, and stores it inside the keychain.
+ 
+ @param password Password used to encrypt the DPK
+ 
+ @return The DPK
+ */
+- (NSData *)generateAndSaveKeyProtectedByPassword:(NSString *)password;
+
+/**
+ Checks if the encrypted Data Protection Key (DPK) is inside the keychain.
+ 
+ @return YES if the encrypted DPK is inside the keychain, NO otherwise
+ */
+- (BOOL)keyExists;
+
+/**
+ Clears security metadata from the keychain.
+ 
+ @return Success (true) or failure (false)
+ */
+- (BOOL)clearKey;
+
+@end
--- a/Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
+++ b/Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
@@ -1,0 +1,184 @@
+//
+//  CDTEncryptionKeychainManager.m
+//  CloudantSync
+//
+//  Created by Enrique de la Torre Fernandez on 09/04/2015.
+//  Copyright (c) 2015 IBM Cloudant. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+//  except in compliance with the License. You may obtain a copy of the License at
+//  http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software distributed under the
+//  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+//  either express or implied. See the License for the specific language governing permissions
+//  and limitations under the License.
+//
+
+#import <CommonCrypto/CommonCryptor.h>
+
+#import "CDTEncryptionKeychainManager.h"
+
+#import "CDTEncryptionKeychainUtils.h"
+#import "CDTEncryptionKeychainConstants.h"
+
+#import "CDTLogging.h"
+
+#define CDTENCRYPTIONKEYCHAINMANAGER_AES_IV_SIZE kCCBlockSizeAES128
+
+@interface CDTEncryptionKeychainManager ()
+
+@property (strong, nonatomic, readonly) CDTEncryptionKeychainStorage *storage;
+
+@end
+
+@implementation CDTEncryptionKeychainManager
+
+#pragma mark - Init object
+- (instancetype)init { return [self initWithStorage:nil]; }
+
+- (instancetype)initWithStorage:(CDTEncryptionKeychainStorage *)storage
+{
+    self = [super init];
+    if (self) {
+        if (storage) {
+            _storage = storage;
+        } else {
+            CDTLogError(CDTDATASTORE_LOG_CONTEXT, @"Storage is mandatory");
+
+            self = nil;
+        }
+    }
+
+    return self;
+}
+
+#pragma mark - Public methods
+- (NSData *)loadKeyUsingPassword:(NSString *)password
+{
+    CDTEncryptionKeychainData *data = [self.storage encryptionKeyData];
+    if (!data || ![self validateEncryptionKeyData:data]) {
+        return nil;
+    }
+
+    NSData *aesKey = [self pbkdf2DerivedKeyForPassword:password
+                                                  salt:data.salt
+                                            iterations:data.iterations
+                                                length:CDTENCRYPTION_KEYCHAIN_AES_KEY_SIZE];
+
+    NSData *dpk = [self decryptDpk:data.encryptedDPK usingAESWithKey:aesKey iv:data.iv];
+
+    return dpk;
+}
+
+- (NSData *)generateAndSaveKeyProtectedByPassword:(NSString *)password
+{
+    NSData *dpk = nil;
+
+    if (![self keyExists]) {
+        dpk = [self generateDpk];
+
+        CDTEncryptionKeychainData *keychainData =
+            [self keychainDataToStoreDpk:dpk encryptedWithPassword:password];
+
+        if (![self.storage saveEncryptionKeyData:keychainData]) {
+            dpk = nil;
+        }
+    }
+
+    return dpk;
+}
+
+- (BOOL)keyExists { return [self.storage encryptionKeyDataExists]; }
+
+- (BOOL)clearKey { return [self.storage clearEncryptionKeyData]; }
+
+#pragma mark - CDTEncryptionKeychainManager+Internal methods
+- (BOOL)validateEncryptionKeyData:(CDTEncryptionKeychainData *)data
+{
+    // Ensure IV has the correct length
+    if (data.iv.length != CDTENCRYPTIONKEYCHAINMANAGER_AES_IV_SIZE) {
+        CDTLogWarn(CDTDATASTORE_LOG_CONTEXT, @"IV does not have the expected size: %i bytes",
+                   CDTENCRYPTIONKEYCHAINMANAGER_AES_IV_SIZE);
+
+        return NO;
+    }
+
+    return YES;
+}
+
+- (NSData *)generateDpk
+{
+    NSData *dpk = [CDTEncryptionKeychainUtils
+        generateSecureRandomBytesWithLength:CDTENCRYPTION_KEYCHAIN_ENCRYPTIONKEY_SIZE];
+
+    return dpk;
+}
+
+- (CDTEncryptionKeychainData *)keychainDataToStoreDpk:(NSData *)dpk
+                                encryptedWithPassword:(NSString *)password
+{
+    NSData *salt = [self generatePBKDF2Salt];
+
+    NSData *key = [self pbkdf2DerivedKeyForPassword:password
+                                               salt:salt
+                                         iterations:CDTENCRYPTION_KEYCHAIN_PBKDF2_ITERATIONS
+                                             length:CDTENCRYPTION_KEYCHAIN_AES_KEY_SIZE];
+
+    NSData *iv = [self generateAESIv];
+
+    NSData *encryptedDpk = [self encryptDpk:dpk usingAESWithKey:key iv:iv];
+
+    CDTEncryptionKeychainData *keychainData =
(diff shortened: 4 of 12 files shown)


Classes/common/Encryption/Keychain/CDTEncryptionKeychainConstants.h
```diff
@@ -0,0 +1,40 @@
...
+// PBKDF2: Number of times the derivation process is applied to a user-provided password
+#define CDTENCRYPTION_KEYCHAIN_PBKDF2_ITERATIONS (NSInteger)10000
+
+// AES: Size (bytes) of the key used to encrypt/decrypt a DPK
+#define CDTENCRYPTION_KEYCHAIN_AES_KEY_SIZE kCCKeySizeAES256
+
+// AES: Size (bytes) of the IV (Initialization Vector) used to encrypt/decrypt a DPK
+#define CDTENCRYPTION_KEYCHAIN_AES_IV_SIZE kCCBlockSizeAES128
+
+#endif
```
rhyshort: New line here

rhyshort:
Copyright statements should mention IBM Cloudant/ IBM

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager+Internal.h
```diff
@@ -0,0 +1,47 @@
...
+ This category is only for testing purposes. It presents some of the private methods used by
+ CDTEncryptionKeychainManager to complete its job.
+ 
+ Notice that although all these methods are instance methods, none of them modifies the state of a
+ CDTEncryptionKeychainManager instance or writes to disk, however they are easier to override
+ this way.
+ */
+@interface CDTEncryptionKeychainManager (Internal)
+
+- (BOOL)validatedEncryptionKeyData:(CDTEncryptionKeychainData *)data;
```
rhyshort: I think this should be called `validEncryptionKeyData:`

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
```diff
@@ -0,0 +1,193 @@
...
+@interface CDTEncryptionKeychainManager ()
+
+@property (strong, nonatomic, readonly) CDTEncryptionKeychainStorage *storage;
+
+@end
+
+@implementation CDTEncryptionKeychainManager
+
+#pragma mark - Init object
+- (instancetype)init { return [self initWithStorage:nil]; }
```
rhyshort: Since storage is mandatory, I don't think this init method should exist

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
```diff
@@ -0,0 +1,193 @@
...
+
+        if (![self.storage saveEncryptionKeyData:keychainData]) {
+            dpk = nil;
+        }
+    }
+
+    return dpk;
+}
+
+- (BOOL)encryptionKeyDataAlreadyGenerated { return [self.storage encryptionKeyDataExists]; }
```
rhyshort: I think this should be named the same as the underlying method it is calling, so `encryptionKeyDataExists`

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
```diff
@@ -0,0 +1,193 @@
...
+    NSData *dpk = [CDTEncryptionKeychainUtils
+        generateRandomBytesInBufferWithLength:CDTENCRYPTION_KEYCHAIN_ENCRYPTIONKEY_SIZE];
+
+    return dpk;
+}
+
+- (CDTEncryptionKeychainData *)keychainDataToStoreDpk:(NSData *)dpk
+                                encryptedWithPassword:(NSString *)password
+{
+    NSData *pbkdf2Salt = [self generatePBKDF2Salt];
```
rhyshort: I think the variable should be called `salt` makes it easier to read and if we can see from the generation method's name what kind of salt it is.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
```diff
@@ -0,0 +1,193 @@
...
+
+    NSData *aesKey =
```
rhyshort: Like above, I think this should just be called `key`

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
indisoluble: But it always exists. `CDTEncryptionKeychainManager` is a subclass of `NSObject`, I have to override the designated initialiser in the parent and redirect it to the designated initialiser in the child.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainConstants.h
indisoluble: Fixed in 1711de8

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager+Internal.h
indisoluble: Fixed in b029beb

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
indisoluble: But there is another method called 'generateEncryptionKeyDataUsingPassword:', this way both method are related.

@mikerhodes What do you think? Any suggestion?

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
indisoluble: Fixed in 56e26d5

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
indisoluble: Fixed in 56e26d5

indisoluble:
@rhyshort Copyright added in 34feabb

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager+Internal.h
mikerhodes: Or possibly, to be a pain, `validateEncryptionKeyData:` as the object is performing a task on `data`, c.f., [For methods that represent actions an object takes, start the name with a verb][1]. `is` should be reserved for accessor methods.

It looks like this method could also be a class method rather than an instance method?

[1]: https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/NamingMethods.html

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
mikerhodes: I think the method naming of "generate" is misleading because the method also stores the keychain data, in addition to generating it. Is there a reason not to mirror the underlying methods like so:

```
-loadEncryptionKeyDataUsingPassword:
-saveEncryptionKeyDataUsingPassword:
-encryptionKeyDataExists
```

I guess the generate speaks to the purpose of the class, "Use this class to generate a Data Protection Key (DPK)"?

mikerhodes: Given the object is a CDTEncryptionKeychainManager, perhaps we can drop some of the names:

```
-loadKeyUsingPassword:
-generateAndSaveKeyProtectedByPassword:
-keyExists:
```

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager+Internal.h
indisoluble: Yes, it could be a class method. I explained why it is not in the [comment above](https://github.com/cloudant/CDTDatastore/pull/128/files#diff-c0d6fc1b018f904068993be320db61c6R23).

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager+Internal.h
```diff
@@ -0,0 +1,47 @@
...
+
+#import "CDTEncryptionKeychainManager.h"
+
+/**
+ This category is only for testing purposes. It presents some of the private methods used by
+ CDTEncryptionKeychainManager to complete its job.
+ 
+ Notice that although all these methods are instance methods, none of them modifies the state of a
+ CDTEncryptionKeychainManager instance or writes to disk, however they are easier to override
+ this way.
```
mikerhodes: It's a bit too easy for this comment to become out of date w.r.t. the implementation.

Do we think they will be overridden? If not, marking them as class methods would imply the stateless nature without the risk of going out of date.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager+Internal.h
```diff
@@ -0,0 +1,47 @@
...
+- (NSData *)generatePBKDF2Salt;
+- (NSData *)generateAESKeyUsingPBKDF2ToDerivePassword:(NSString *)password
+                                             withSalt:(NSData *)salt
+                                           iterations:(NSInteger)iterations
+                                               length:(NSUInteger)length;
+
+- (NSData *)generateAESIv;
+
+- (NSData *)encryptDpk:(NSData *)dpk usingAESWithKey:(NSData *)key iv:(NSData *)iv;
+- (NSData *)decryptCipheredDpk:(NSData *)cipheredDpk usingAESWithKey:(NSData *)key iv:(NSData *)iv;
```
mikerhodes: `decrypt` implies encrypted, remove `Ciphered`.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.h
```diff
@@ -0,0 +1,76 @@
...
+//  and limitations under the License.
+//
+
+#import <Foundation/Foundation.h>
+
+#import "CDTEncryptionKeychainStorage.h"
+
+/**
+ Use this class to generate a Data Protection Key (DPK), i.e. a strong password that can be used
+ later on for other purposes like ciphering a database.
```
mikerhodes: I'm not sure "ciphered" is in common use in this context, though is probably technically accurate. I'd use "encrypted", "encrypting" etc.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.h
```diff
@@ -0,0 +1,76 @@
...
+ */
+@interface CDTEncryptionKeychainManager : NSObject
+
+/**
+ Initialise a manager with a CDTEncryptionKeychainStorage instance.
+ 
+ A CDTEncryptionKeychainStorage gives access to the keychain throught an identifier, this means
+ that the DPK saved to the keychain is bound to this identifier. If a new manager is created with
+ a storage pointing to another identifier, the previous DPK will not be accesible. So you can
+ create as many DPKs as you want as long as you provide different identifiers.
```
mikerhodes: I think this is a bit confusing. Perhaps:

"""
A CDTEncryptionKeychainStorage binds an entry in the keychain to an identifier. The data protection key (DPK) saved to the keychain by this class will therefore be bound to the storage's identifier. To save different DPKs (say for different users of your application), create multiple managers using storages initialised with different identifiers.
"""

Is that correct? I think "the previous DPK will not be accessible" is a bit fear-inducing -- it's still accessible, you just need to use a storage with the right identifier.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager+Internal.h
indisoluble: Fixed in a6a9c22

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.h
```diff
@@ -0,0 +1,76 @@
...
+
+#import "CDTEncryptionKeychainStorage.h"
+
+/**
+ Use this class to generate a Data Protection Key (DPK), i.e. a strong password that can be used
+ later on for other purposes like ciphering a database.
+ 
+ The generated DPK is automatically encrypted and saved to the keychain, for this reason, it is
+ neccesary to provide a password to generate and retrieve the DPK.
+ */
```
mikerhodes: I would add the details here about the method of saving and encrypting the DPK.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
```diff
@@ -0,0 +1,193 @@
...
+        return NO;
+    }
+
+    // Ensure IV has the correct length
+    if (data.iv.length != CDTENCRYPTION_KEYCHAIN_AES_IV_SIZE) {
+        CDTLogWarn(CDTDATASTORE_LOG_CONTEXT, @"IV does not have the expected size: %i bytes",
+                   CDTENCRYPTION_KEYCHAIN_AES_IV_SIZE);
+
+        return NO;
+    }
```
mikerhodes: I'm not sure these checks are useful. It'd be easy to break old stored data if we change the values of the constants and forgot to update these checks in some way. In addition, given this is used at the decryption stage, we're just going to end up with an invalid key if these have been changed to values which are not suitable for decrypting the stored key.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
```diff
@@ -0,0 +1,193 @@
...
+
+    return YES;
+}
+
+- (NSData *)generateDpk
+{
+    NSData *dpk = [CDTEncryptionKeychainUtils
+        generateRandomBytesInBufferWithLength:CDTENCRYPTION_KEYCHAIN_ENCRYPTIONKEY_SIZE];
```
mikerhodes: If this method is being used to generate a key, it needs to be securely random bytes in a buffer, and the method needs to make that clear.

Looking at the method, the bytes are secure, so rename the method `-generateSecureRandomBytesInBufferWithLength`. I don't know we need `InBuffer` as the returned NSData isn't a buffer but does contain the bytes.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
```diff
@@ -0,0 +1,193 @@
...
+    NSData *salt = [CDTEncryptionKeychainUtils
+        generateRandomBytesInBufferWithLength:CDTENCRYPTION_KEYCHAIN_PBKDF2_SALT_SIZE];
+
+    return salt;
+}
+
+- (NSData *)generateAESKeyUsingPBKDF2ToDerivePassword:(NSString *)password
+                                             withSalt:(NSData *)salt
+                                           iterations:(NSInteger)iterations
+                                               length:(NSUInteger)length
```
mikerhodes: You are not deriving a password, but instead a key. In addition it's not necessarily an AES key, so far as I can see. So I'd suggest `-pkbdf2DerivedKeyForPassword:salt:iterations:length:`.

mikerhodes: Or just use the Utils class's name -- there's no guarantee given in that name that PKBDF2 is used, so it's possible someone might change the alg used without updating the name of this method.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
```diff
@@ -0,0 +1,193 @@
...
+
+- (NSData *)generateAESIv
+{
+    NSData *iv = [CDTEncryptionKeychainUtils
+        generateRandomBytesInBufferWithLength:CDTENCRYPTION_KEYCHAIN_AES_IV_SIZE];
+
+    return iv;
+}
+
+- (NSData *)encryptDpk:(NSData *)dpk usingAESWithKey:(NSData *)key iv:(NSData *)iv
```
mikerhodes: Again, using AES in the name here and not in the Utils method risks the name here becoming incorrect.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
```diff
@@ -0,0 +1,193 @@
...
+    
+    NSData *aesKey =
+        [self generateAESKeyUsingPBKDF2ToDerivePassword:password
+                                               withSalt:data.salt
+                                             iterations:data.iterations
+                                                 length:CDTENCRYPTION_KEYCHAIN_AES_KEY_SIZE];
+
+    NSData *dpk = [self decryptCipheredDpk:data.encryptedDPK usingAESWithKey:aesKey iv:data.iv];
+
+    return dpk;
```
mikerhodes: `key` would be a better name than `dpk` simply because the name is `retrieve...Key`.

Tests/Tests/Encryption/Keychain/CDTEncryptionKeychainManagerTests.m
```diff
@@ -0,0 +1,337 @@
...
+#import <XCTest/XCTest.h>
+
+#import "CDTEncryptionKeychainManager+Internal.h"
+
+#import "CDTEncryptionKeychainUtils.h"
+#import "CDTEncryptionKeychainConstants.h"
+
+#import "CDTMockEncryptionKeychainStorage.h"
+
+#define CDTENCRYPTIONKEYCHAINMANAGERTESTS_PASSWORD_DEFAULT @"password"
```
mikerhodes: Suggest making this more complicated -- unicode chars, spaces, special chars etc.

Tests/Tests/Encryption/Keychain/CDTEncryptionKeychainManagerTests.m
```diff
@@ -0,0 +1,337 @@
...
+NSString *const kCDTEncryptionKeychainManagerTestsEncryptDpk =
+    @"kCDTEncryptionKeychainManagerTestsEncryptDpk";
+NSString *const kCDTEncryptionKeychainManagerTestsDecryptCipheredDpk =
+    @"kCDTEncryptionKeychainManagerTestsDecryptCipheredDpk";
+
+@interface CDTEncryptionCustomKeychainManager : CDTEncryptionKeychainManager
+
+@property (strong, nonatomic) NSMutableArray *lastSteps;
+
+@end
```
mikerhodes: Explain what this is for in a doc comment.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
indisoluble: Fixed in 880b63b

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager+Internal.h
indisoluble: Fixed in 738fc82

Tests/Tests/Encryption/Keychain/CDTEncryptionKeychainManagerTests.m
```diff
@@ -0,0 +1,337 @@
...
+}
+
+- (NSData *)decryptCipheredDpk:(NSData *)cipheredDpk usingAESWithKey:(NSData *)key iv:(NSData *)iv
+{
+    [self.lastSteps addObject:kCDTEncryptionKeychainManagerTestsDecryptCipheredDpk];
+
+    return [super decryptCipheredDpk:cipheredDpk usingAESWithKey:key iv:iv];
+}
+
+@end
```
mikerhodes: Getting quite close here to requiring a proper mock object framework to keep our lines of code down. Think we're just about okay.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager+Internal.h
indisoluble: CDTEncryptionCustomKeychainManager [overrides all these methods](https://github.com/cloudant/CDTDatastore/pull/128/files#diff-91343d9161d85d0c76aa178e5fd9f785R273)

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.h
indisoluble: Fixed in 10b8ab7

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.h
indisoluble: Fixed in 5ac3125

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.h
indisoluble: Fixed in 2e29e6c

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
indisoluble: The methods in `CDTEncryptionKeychainUtils` does not mention in their names the algorithm used to generate the different values, however the comments above them do. Other comments along the library also mention the algorithms used and the constants in `CDTEncryptionKeychainConstants.h` are prefixed with AES or PBKDF2. If I delete the reference to PBKDF2 here, I should the same in all the other places. Instead of that, I'd rather change the names of the methods in `CDTEncryptionKeychainUtils`.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
indisoluble: As mentioned in the [previous comment](https://github.com/cloudant/CDTDatastore/pull/128#discussion_r29869733), I'd rather change the name of the methods in `CDTEncryptionKeychainUtils `.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
indisoluble: But I also use this name in [generateAndSaveKeyProtectedByPassword:](https://github.com/cloudant/CDTDatastore/pull/128/files#diff-1aa2c80639dcad5df86af652bc515982R77). And DPK stands for Data Protection **Key**.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
indisoluble: I agree that validating the number of iterations is not useful, also this is a value that we will change soon enough.
But we have to check the length of the IV, the function `CCCryptorCreate` in the `CommonCrypto` framework assumes that this buffer has the correct size. We also validate this size in `CDTEncryptionKeychainUtils+AES:applyOperation:toData:withKey:iv:` but we use a `NSAssert` in this case.

indisoluble: Fixed in fd4e043

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
indisoluble: Fixed in d2bf1dd

Tests/Tests/Encryption/Keychain/CDTEncryptionKeychainManagerTests.m
indisoluble: I do not think we need a more complicated password, we already tested the code that derives a key with the password and cipher the DPK in: `CDTEncryptionKeychainUtilsAESTests` & `CDTEncryptionKeychainUtilsPBKDF2Tests`.

Tests/Tests/Encryption/Keychain/CDTEncryptionKeychainManagerTests.m
indisoluble: Fixed in a5efbf8

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
indisoluble: As suggested by @rhyshort, I use `UNAVAILABLE_ATTRIBUTE ` to make 'init' unavailable.

Fixed in 10d1a60.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
mikerhodes: I would change the names in Utils, as you suggest. I think it's actually quite important we're specific here rather than general.

Tests/Tests/Encryption/Keychain/CDTEncryptionKeychainManagerTests.m
mikerhodes: Okay, perhaps add a comment above the test default password here. Otherwise it does look over-simplistic and it's not worth someone else asking the question later :)

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager+Internal.h
mikerhodes: Okay, well in that case the comment should probably be removed as the caller can't assume it's true?

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.h
mikerhodes: Note the random values are secure random values -- this is important.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
mikerhodes: Why not check against the constants for key size that CCCryptoCreate uses rather than our own in that case? `kCCKeySizeAES128` or whichever?

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
mikerhodes: Again I'd suggest using the constants defined in CommonCrypto rather than defining our own?

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager+Internal.h
indisoluble: Fixed in e0f0763

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.h
indisoluble: Fixed in e73b0b4

rhyshort:
Once you've addressed all mikes comments, I'm happy for this to go in +1

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
indisoluble: Fixed in a1d0655

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
indisoluble: Are you talking about `CDTENCRYPTION_KEYCHAIN_AES_KEY_SIZE`? In this contrary, on the contrary to the size of IV, there are 3 valid values: 16, 24 & 32.

Tests/Tests/Encryption/Keychain/CDTEncryptionKeychainManagerTests.m
indisoluble: Fixed in 7b563e4

Classes/common/Encryption/Keychain/CDTEncryptionKeychainManager.m
mikerhodes: I think the work done works, you're right we need the key size defined, but the IV size we can pull from the Crypto constants.

mikerhodes:
Okay, :+1:

PR #130, opened 2015-04-21
Add Index class to manage index attribute validation for Query

alfinkel:
_What_

Establish a new index class to house all index related validation and any utility functions related to index processing for use in Query.

_Why_

In an effort to add support for query full text search we are starting the process of migrating index related functionality to this new index class.  It allows us to better compartmentalize index related functionality and pass index data around the various change agents in Query - mobile (such as manager, creator, updater, etc...) in a cleaner manner.

_How_

In this PR we are focused on moving validation currently performed in the index manager and the index creator to the new index class.  Likewise, we added some utility functions that supports index creator functionality such as retrieving the index settings as a String and comparing index types and settings as they relate to index types.

_Notes_

- This implementation is a port of the work done in sync-android.  Specifically https://github.com/cloudant/sync-android/pull/109.  As such, the comments and issues found with the Android implementation have already been applied to this implementation and FB cases have been opened for both platforms whenever appropriate.
- The integration of the index class with the current index management framework will be done in a follow up PR.
- Likewise, query search work will also happen in follow up PRs.
- Finally it is worth noting that the index class in this PR is an initial cut of the class and only meant to provide functionality that will allow us to complete the query full text search work.  A separate FB case (https://cloudant.fogbugz.com/f/cases/46029) is already open to complete additional work on the index class at a future date.

reviewer @mikerhodes 
reviewer @rhyshort 

BugId: 46368

### DIFF
--- a/Classes/common/query/CDTQIndex.h
+++ b/Classes/common/query/CDTQIndex.h
@@ -1,0 +1,75 @@
+//
+//  CDTQIndex.h
+//
+//  Created by Al Finkelstein on 2015-04-20
+//  Copyright (c) 2015 Cloudant. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+//  except in compliance with the License. You may obtain a copy of the License at
+//    http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software distributed under the
+//  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+//  either express or implied. See the License for the specific language governing permissions
+//  and limitations under the License.
+
+#import <Foundation/Foundation.h>
+
+extern NSString *const kCDTQJsonType;
+extern NSString *const kCDTQTextType;
+
+/**
+ * This class provides functionality to manage an index
+ */
+@interface CDTQIndex : NSObject
+
+@property (nonatomic, strong) NSArray *fieldNames;
+@property (nonatomic, strong) NSString *indexName;
+@property (nonatomic, strong) NSString *indexType;
+@property (nonatomic, strong) NSDictionary *indexSettings;
+
+/**
+ * This function sets the index type to the default setting of "json"
+ *
+ * @param indexName the index name
+ * @param fieldNames the field names in the index
+ * @return the Index object or nil if arguments passed in were invalid.
+ */
++ (instancetype)index:(NSString *)indexName withFields:(NSArray *)fieldNames;
+
++ (instancetype)index:(NSString *)indexName
+           withFields:(NSArray *)fieldNames
+               ofType:(NSString *)indexType;
+
+/**
+ * This function handles index specific validation and ensures that the constructed
+ * Index object is valid.
+ *
+ * @param indexName the index name
+ * @param fieldNames the field names in the index
+ * @param indexType the index type (json or text)
+ * @param indexSettings the optional settings used to configure the index.
+ *                      Only supported parameter is 'tokenize' for text indexes only.
+ * @return the Index object or nil if arguments passed in were invalid.
+ */
++ (instancetype)index:(NSString *)indexName
+           withFields:(NSArray *)fieldNames
+               ofType:(NSString *)indexType
+         withSettings:(NSDictionary *)indexSettings;
+
+/**
+ * Compares the index type and accompanying settings with the passed in arguments.
+ *
+ * @param indexType the index type to compare to
+ * @param indexSettings the indexSettings to compare to as an NSString
+ * @return YES/NO - whether there is a match
+ */
+- (BOOL)compareIndexTypeTo:(NSString *)indexType withIndexSettings:(NSString *)indexSettings;
+
+/**
+ * Converts the index settings to a JSON string
+ *
+ * @return the JSON representation of the index settings
+ */
+- (NSString *)settingsAsJSON;
+
+@end
--- a/Classes/common/query/CDTQIndex.m
+++ b/Classes/common/query/CDTQIndex.m
@@ -1,0 +1,171 @@
+//
+//  CDTQIndex.m
+//
+//  Created by Al Finkelstein on 2015-04-20
+//  Copyright (c) 2015 Cloudant. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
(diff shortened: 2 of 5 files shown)


Classes/common/query/CDTQIndex.h
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * This function sets the index type to the default setting of "json"
+ *
+ * @param fieldNames the field names in the index
+ * @param indexName the index name
+ * @return the Index object or nil if arguments passed in were invalid.
+ */
++ (CDTQIndex *)instanceWithFields:(NSArray *)fieldNames indexName:(NSString *)indexName;
+
```
rhyshort: I think it should be `index:WithFields`

Classes/common/query/CDTQIndex.h
```diff
@@ -0,0 +1,75 @@
...
++ (CDTQIndex *)instanceWithFields:(NSArray *)fieldNames
+                        indexName:(NSString *)indexName
+                        indexType:(NSString *)indexType;
+
```
rhyshort: I think it should be `index:WithFields:ofType`

Classes/common/query/CDTQIndex.h
```diff
@@ -0,0 +1,75 @@
...
+ * @param indexType the index type (json or text)
+ * @param indexSettings the optional settings used to configure the index.
+ *                      Only supported parameter is 'tokenize' for text indexes only.
+ * @return the Index object or nil if arguments passed in were invalid.
+ */
++ (CDTQIndex *)instanceWithFields:(NSArray *)fieldNames
+                        indexName:(NSString *)indexName
+                        indexType:(NSString *)indexType
+                    indexSettings:(NSDictionary *)indexSettings;
+
```
rhyshort: I think it should be `index:WithFields:ofType:WithSettings` or similar

Classes/common/query/CDTQIndex.m
```diff
@@ -0,0 +1,174 @@
...
+//  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+//  either express or implied. See the License for the specific language governing permissions
+//  and limitations under the License.
+
+#import "CDTQIndex.h"
+
+#import "CDTQLogging.h"
+
+NSString *const kCDTQJsonType = @"json";
+NSString *const kCDTQTextType = @"text";
```
rhyshort: Since these are constant should they be static too?

alfinkel: These are externs.  I don't believe that I can declare them as both extern and static.

Classes/common/query/CDTQIndex.h
alfinkel: Fixed in d1d74be63df1df223f9d3c274975c21bd3a6a147

Classes/common/query/CDTQIndex.h
alfinkel: Fixed in d1d74be63df1df223f9d3c274975c21bd3a6a147

Classes/common/query/CDTQIndex.h
alfinkel: Fixed in d1d74be63df1df223f9d3c274975c21bd3a6a147

rhyshort:
Looks good to me +1

Tests/Tests/CDTQIndexTests.m
```diff
@@ -0,0 +1,174 @@
...
+        indexName = @"basic";
+    });
+    
+    it(@"constructs an index instance with the default index type", ^{
+        CDTQIndex *index = [CDTQIndex index:indexName withFields:fieldNames];
+        
+        expect(index.indexName).to.equal(@"basic");
+        expect(index.fieldNames).to.containsAllElements(@[ @"name", @"age" ]);
+        expect(index.indexType).to.equal(@"json");
+        expect(index.indexSettings).to.beNil;
```
mikerhodes: I think your `beNil` calls require parentheses to work: https://github.com/specta/expecta/issues/50

Classes/common/query/CDTQIndex.m
```diff
@@ -0,0 +1,171 @@
...
+-(NSString *) settingsAsJSON {
+    if (!self.indexSettings) {
+        LogWarn(@"Index settings are nil.  Nothing to return.");
+        return nil;
+    }
+    NSError *error;
+    NSData *settingsData = [NSJSONSerialization dataWithJSONObject:self.indexSettings
+                                                           options:kNilOptions
+                                                             error:&error];
+    if (error) {
```
mikerhodes: Always check the return value of the method rather than checking the `error` parameter to detect an error. There's no guarantee or convention that the error parameter being `nil` or not means the method failed or succeeded.

Classes/common/query/CDTQIndex.m
```diff
@@ -0,0 +1,171 @@
...
+        _indexType = indexType;
+        _indexSettings = indexSettings;
+    }
+    return self;
+}
+
++ (CDTQIndex *)index:(NSString *)indexName withFields:(NSArray *)fieldNames
+{
+    return [CDTQIndex index:indexName withFields:fieldNames ofType:kCDTQJsonType];
+}
```
mikerhodes: For these three convenience constructors, we should return `instancetype`. In addition, we should call the constructor with more arguments using `[[self class] index:...]. Finally, for the `-initWithFields:...` method at the end of the final convenience constructor, use `[[self class] alloc]...`.

This allows sub-classes to override these methods and have the override method be called.

See the instancetype discussion here: https://developer.apple.com/library/ios//releasenotes/ObjectiveC/ModernizationObjC/AdoptingModernObjective-C/AdoptingModernObjective-C.html.

Tests/Tests/CDTQIndexTests.m
alfinkel: Fixed in a2123c3d9cd81db5b433890d4d30128dccb7b00e

Classes/common/query/CDTQIndex.m
alfinkel: Fixed in a2123c3d9cd81db5b433890d4d30128dccb7b00e

Classes/common/query/CDTQIndex.m
alfinkel: Got it.  Makes sense.  Fixed in a2123c3d9cd81db5b433890d4d30128dccb7b00e.

mikerhodes:
Looking good now :+1:

PR #127, opened 2015-04-21
45570 Generate, retrieve & store a key: Data & storage classes to save a key

indisoluble:
*What:*
Add classes to safely store a DPK (Data Protection Key) in the keychain.

*Why:*
This PR is part of a bigger development to provide a mechanism to generate a key and safely store it.

*How:*
Instead of saving the DPK, we save enough data to recreate the DPK after the user provides the correct password.

*Tests:*
* A data class can only be created if all properties are provided
* A data class can archive and unarchive.

reviewer @mikerhodes 
reviewer @rhyshort 

**NOTE:** Wait to review this PR until PR #126 is closed. This branch will be rebased and there will be only one commit to review.

PR #133, opened 2015-04-22
45597 Provide a key to SQLCipher (key provider based on keychain encryption classes)

indisoluble:
*What:*
Implement a class that conforms to protocol `CDTEncryptionKeyProvider`, i.e. a class that generates a strong key and stores it safely in the keychain.

*Why:*
It is not easy to generate a good key for encryption purposes, also once it is generated it has to be stored in a way that no one but the authorised user can retrieve it. The class implemented here can be used as it is or as an example.

*How:*
It will be based on the classes introduced in PRs: #126, #127 & #128.

*Tests:*
* Provider generates a key if there is not one in the keychain.
* Provider retrieve the key in the keychain if there is one.
* It returns an hexadecimal string in both cases.

reviewer @mikerhodes 
reviewer @rhyshort 

**NOTE:** Wait to review this PR until PR #128  is closed. This branch will be rebased and there will be only one commit to review.

### DIFF
--- a/Classes/common/Encryption/Keychain/CDTEncryptionKeychainProvider.h
+++ b/Classes/common/Encryption/Keychain/CDTEncryptionKeychainProvider.h
@@ -1,0 +1,54 @@
+//
+//  CDTEncryptionKeychainProvider.h
+//  CloudantSync
+//
+//  Created by Enrique de la Torre Fernandez on 21/04/2015.
+//  Copyright (c) 2015 IBM Cloudant. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+//  except in compliance with the License. You may obtain a copy of the License at
+//  http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software distributed under the
+//  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+//  either express or implied. See the License for the specific language governing permissions
+//  and limitations under the License.
+//
+
+#import <Foundation/Foundation.h>
+
+#import "CDTEncryptionKeyProvider.h"
+
+/**
+ This class conforms to protocol CDTEncryptionKeyProvider and it can be used to create an
+ encrypted datastore.
+ 
+ Given an user-provided password and an identifier, it generates a strong key and store it safely
+ in the keychain, so the same key can be retrieved later provided that the user supplies the same
+ password and id.
+ 
+ The password is used to protect the key before saving it to the keychain. The identifier is an
+ easy way to have more than one encryption key in the same app, the only condition is to provide
+ different ids for each of them.
+ 
+ @see CDTEncryptionKeyProvider
+ @see CDTEncryptionKeychainManager
+ @see CDTEncryptionKeychainStorage
+ */
+@interface CDTEncryptionKeychainProvider : NSObject <CDTEncryptionKeyProvider>
+
+- (instancetype)init UNAVAILABLE_ATTRIBUTE;
+
+/**
+ This is a convenience method that creates a CDTEncryptionKeychainManager with the provided
+ identifier before calling the init method to create a provider.
+ 
+ @param password An user-provided password
+ @param identifier The data saved in the keychain will be accessed with this identifier
+ 
+ @return A key provider
+ 
+ @see CDTEncryptionKeychainManager
+ */
++ (instancetype)providerWithPassword:(NSString *)password forIdentifier:(NSString *)identifier;
+
+@end
--- a/Classes/common/Encryption/Keychain/CDTEncryptionKeychainProvider.m
+++ b/Classes/common/Encryption/Keychain/CDTEncryptionKeychainProvider.m
@@ -1,0 +1,80 @@
+//
+//  CDTEncryptionKeychainProvider.m
+//  CloudantSync
+//
+//  Created by Enrique de la Torre Fernandez on 21/04/2015.
+//  Copyright (c) 2015 IBM Cloudant. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+//  except in compliance with the License. You may obtain a copy of the License at
+//  http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software distributed under the
+//  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+//  either express or implied. See the License for the specific language governing permissions
+//  and limitations under the License.
+//
+
+#import "CDTEncryptionKeychainProvider.h"
+
+#import "CDTEncryptionKeychainManager.h"
+
+#import "TDMisc.h"
+#import "CDTLogging.h"
+
+@interface CDTEncryptionKeychainProvider ()
+
+@property (strong, nonatomic) NSString *password;
+@property (strong, nonatomic) CDTEncryptionKeychainManager *manager;
+
+@end
+
+@implementation CDTEncryptionKeychainProvider
+
+#pragma mark - Init method
+- (instancetype)init { return [self initWithPassword:nil forManager:nil]; }
+
+- (instancetype)initWithPassword:(NSString *)password
+                      forManager:(CDTEncryptionKeychainManager *)manager
+{
+    self = [super init];
+    if (self) {
+        if (password && manager) {
+            _password = password;
+            _manager = manager;
+        } else {
+            CDTLogError(CDTDATASTORE_LOG_CONTEXT, @"All parameters are mandatory");
+
+            self = nil;
+        }
+    }
+
+    return self;
+}
+
+#pragma mark - CDTEncryptionKeyProvider methods
+- (NSString *)encryptionKey
+{
+    NSData *data = nil;
+    if ([self.manager keyExists]) {
+        data = [self.manager loadKeyUsingPassword:self.password];
+    } else {
+        data = [self.manager generateAndSaveKeyProtectedByPassword:self.password];
+    }
+    
+    NSString *hexStr = (data ? TDHexFromBytes(data.bytes, data.length) : nil);
+    
+    return hexStr;
+}
+
+#pragma mark - Public class methods
++ (instancetype)providerWithPassword:(NSString *)password forIdentifier:(NSString *)identifier
+{
+    CDTEncryptionKeychainStorage *storage =
+        [[CDTEncryptionKeychainStorage alloc] initWithIdentifier:identifier];
+    CDTEncryptionKeychainManager *manager =
+        [[CDTEncryptionKeychainManager alloc] initWithStorage:storage];
+
+    return [[[self class] alloc] initWithPassword:password forManager:manager];
+}
+
+@end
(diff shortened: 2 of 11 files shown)


Classes/common/Encryption/Keychain/CDTEncryptionKeychainProvider.h
```diff
@@ -0,0 +1,69 @@
...
+ 
+ Given an user-provided password, it generates a strong key and store it safely in the keychain,
+ so the same key can be retrieved later provided that the user supplies the same password.
+ 
+ Data is stored in the keychain through a CDTEncryptionKeychainManager which in turn uses a
+ CDTEncryptionKeychainStorage. A CDTEncryptionKeychainStorage groups the information stored in
+ the keychain with an identifier, this means that it is possible to generate & store multiple
+ keys as long as a different CDTEncryptionKeychainManager with a different
+ CDTEncryptionKeychainStorage is used for each of them (CDTEncryptionKeychainStorage is different
+ from other if it is created with a different identifier).
```
mikerhodes: I think this paragraph goes into too much depth for user-facing documentation. Instead, summarise the use of identifier.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainProvider.h
```diff
@@ -0,0 +1,69 @@
...
+ The returned provider will pass the password to the manager to generate the key or get it from the
+ keychain if it already exists.
+ 
+ @param password An user-provided password
+ @param manager A manager to generate and store the resulting key
+ 
+ @see CDTEncryptionKeychainManager
+ */
+- (instancetype)initWithPassword:(NSString *)password
+                      forManager:(CDTEncryptionKeychainManager *)manager;
```
mikerhodes: I'd question whether this should be provided for a developer using the library -- why's it required over the convenience constructor below?

Classes/common/Encryption/Keychain/CDTEncryptionKeychainProvider.m
```diff
@@ -0,0 +1,78 @@
...
+    if ([self.manager keyExists]) {
+        data = [self.manager loadKeyUsingPassword:self.password];
+    } else {
+        data = [self.manager generateAndSaveKeyProtectedByPassword:self.password];
+    }
+    
+    NSString *hexStr = (data ? TDHexFromBytes(data.bytes, data.length) : nil);
+    
+    return hexStr;
+}
```
mikerhodes: I thought this was just going to provide the NSData of the key, hexing was the responsibility of the SQLCipher specific classes?

Tests/Tests/Encryption/Keychain/CDTEncryptionKeychainProviderTests.m
```diff
@@ -0,0 +1,158 @@
...
+}
+
+- (void)testEncryptionKeyRetrieveEncryptionKeyDataIfDataWasGeneratedBefore
+{
+    self.mockManager.keyExistsResult = YES;
+
+    [self.provider encryptionKey];
+
+    XCTAssertTrue(self.mockManager.keyExistsResult && self.mockManager.loadKeyUsingPasswordExecuted,
+                  @"Get the key from keychain if it was generated before");
```
mikerhodes: Why would `keyExistsResult` be NO here?

doc/encryption.md
```diff
@@ -23,12 +23,16 @@ defines one method:
...
 ```
 
-For now, it is up to you to safely store the key that you have to return with
-this method (if the method return nil, the database won't be encrypted
-regardless of the subspec defined in your `Podfile`).
+Alternatively, you can use the class
+[CDTEncryptionKeychainProvider][CDTEncryptionKeychainProvider]. It already
+implements this protocol, it handles generating a strong key and stores it
+safely in the keychain. However, if you decide to code your own class (and use
+this one as a refence), keep in mind that if the method return nil, the database
```
mikerhodes: `reference`, `returns`

doc/encryption.md
```diff
@@ -23,12 +23,16 @@ defines one method:
 - (NSString *)encryptionKey;
 ```
 
-For now, it is up to you to safely store the key that you have to return with
-this method (if the method return nil, the database won't be encrypted
-regardless of the subspec defined in your `Podfile`).
+Alternatively, you can use the class
+[CDTEncryptionKeychainProvider][CDTEncryptionKeychainProvider]. It already
+implements this protocol, it handles generating a strong key and stores it
```
mikerhodes: `which handles generating a strong key from a user-provided password`

Tests/Tests/Encryption/Keychain/Mock/CDTMockEncryptionKeychainManager.m
```diff
@@ -0,0 +1,76 @@
...
+}
+
+- (BOOL)clearKey
+{
+    self.clearKeyExecuted = YES;
+
+    return self.clearKeyResult;
+}
+
+@end
```
mikerhodes: Okay, I think considering a mocking framework might be needed here. OCMock, for example, allows you to do this without the need for the extra code you're needing to write. I think its syntax is arguably clearer too.

Might be other mocking frameworks you could use.

mikerhodes: Though, clearly feel free to argue against this.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainProvider.h
indisoluble: Fixed in 67d0037

Classes/common/Encryption/Keychain/CDTEncryptionKeychainProvider.h
indisoluble: I need the `init` to test the class (`CDTEncryptionKeychainProviderTests`), to pass a mock manager.

@mikerhodes: What if a move the `init` to an `internal` category (like `CDTEncryptionKeychainManager+Internal.h`)?

Classes/common/Encryption/Keychain/CDTEncryptionKeychainProvider.m
indisoluble: That is done in a later PR: #144.

Tests/Tests/Encryption/Keychain/CDTEncryptionKeychainProviderTests.m
indisoluble: My mistake, the var to check in the assert is `keyExistsExecuted`

indisoluble: Fixed in 984e4fc

doc/encryption.md
indisoluble: Fixed in 5a67d7d

doc/encryption.md
indisoluble: Fixed in c00037c

Tests/Tests/Encryption/Keychain/Mock/CDTMockEncryptionKeychainManager.m
indisoluble: I'm happy to use a mock framework and OCMock looks like a good option. But:
- There is another mock class next to this one: `CDTMockEncryptionKeychainStorage`.
- And also there is a another group called `Mock` with files: `CDTQMatcherIndexManager`, `CDTQMatcherQueryExecutor`, ...

@mikerhodes: If I add OCMock, what should I do with the other classes mentioned above?

Classes/common/Encryption/Keychain/CDTEncryptionKeychainProvider.m
```diff
@@ -0,0 +1,78 @@
...
+
+@property (strong, nonatomic) NSString *password;
+@property (strong, nonatomic) CDTEncryptionKeychainManager *manager;
+
+@end
+
+@implementation CDTEncryptionKeychainProvider
+
+#pragma mark - Init method
+- (instancetype)init { return [self initWithPassword:nil forManager:nil]; }
```
rhyshort: I think we should use __attribute to disable this like we have else where

Classes/common/Encryption/Keychain/CDTEncryptionKeychainProvider.m
```diff
@@ -0,0 +1,78 @@
...
+
+- (instancetype)initWithPassword:(NSString *)password
+                      forManager:(CDTEncryptionKeychainManager *)manager
+{
+    self = [super init];
+    if (self) {
+        if (password && manager) {
```
rhyshort: I think we should use NSParameterAssert to me consistent with CDTDatastore's init method

Classes/common/Encryption/Keychain/CDTEncryptionKeychainProvider.m
indisoluble: Fixed in 3dcf027

Classes/common/Encryption/Keychain/CDTEncryptionKeychainProvider.m
indisoluble: But I think this case is not exactly the same. In the case of `CDTDatastore`, if a key provider is not supplied, we can not return a datastore and without a datastore, you can do pretty much nothing at all. That is why I favoured an `Assert`.
In the case of a `CDTEncryptionKeychainProvider`, the password as well as the manager are mandatory, so I have to validate them but there are many reason why a password might not be supplied, for example if the user chooses to delegate to the `init` method to check if a password was typed in the keyboard and this is not a reason to raise an exception.
Of course, this is only my point of view, it is OK for me to change the validation but I'd like some agreement about it.

@mikerhodes: What do you think?

Tests/Tests/Encryption/Keychain/Mock/CDTMockEncryptionKeychainManager.m
mikerhodes: Probably a new PR and we could replace the other one too if we wanted to go ahead with using mocks.

I think the group is called `Matcher` rather than `Mock`.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainProvider.m
mikerhodes: I think stick with returning `nil`.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainProvider.h
mikerhodes: I think that works.

Tests/Tests/Encryption/Keychain/Mock/CDTMockEncryptionKeychainManager.m
indisoluble: Created **BugzID 47179** to track this change.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainProvider.m
indisoluble: OK.

@rhyshort: Do you have another comment? If not, I'm going to keep the validation as it is now.

Classes/common/Encryption/Keychain/CDTEncryptionKeychainProvider.h
indisoluble: Fixed in 3cd2d20

Classes/common/Encryption/Keychain/CDTEncryptionKeychainProvider.m
rhyshort: I'm not sure why a user would use the init method to indicate bad input from the app's user, but I think I understand the other reasons why we are just returning `nil` so leave it how it is

rhyshort:
+1

mikerhodes:
:+1: needs a bit of squashing, but looking good.

PR #134, opened 2015-04-24
45570 Generate, retrieve & store a key: Use target specific functions to access keychain

indisoluble:
*What:*
Code in PR #127 will not work on Mac OS X.

*Why:*
The code to access the keychain, although it compiles in both targets, it is not going to work properly in Mac OS because iOS and Mac OS use different [paradigms](https://developer.apple.com/library/mac/documentation/Security/Conceptual/keychainServConcepts/02concepts/concepts.html).

*How:*
As suggested in the documentation (under title *Keychain Services Ease of Use*), we are going to use the functions: `SecKeychainFindGenericPassword` and `SecKeychainAddGenericPassword`.

reviewer @mikerhodes 
reviewer @rhyshort

### DIFF
--- a/Classes/common/Encryption/Keychain/CDTEncryptionKeychainStorage.m
+++ b/Classes/common/Encryption/Keychain/CDTEncryptionKeychainStorage.m
@@ -61,13 +61,9 @@
 {
     CDTEncryptionKeychainData *encryptionData = nil;
 
-    NSData *data = nil;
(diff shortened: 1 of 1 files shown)


mikerhodes:
This looks good to me, looking through the docs for the OS X stuff. I'm assuming the iOS stuff is sound. :+1:

rhyshort:
+1

PR #135, opened 2015-04-24
Check DatastoreManager passed to CDTReplicatorFactory

### LINKED ISSUE #132: CDTReplicatorFactory initWithDatastoreManager:nil returns a non-nil object
If you call:
```objc
CDTReplicatorFactory *repFactory = [[CDTReplicatorFactory alloc] initWithDatastoreManager:nil];
```
The result is a non-nil object that can then be used to create a `CDTReplicator` that is non-nil but is essentially useless when you try to use it.

rhyshort:
Check if the DatastoreManager passed to CDTReplicatorFactory
is nil, if so CDTReplicatoryFacotry initWithDatastoreManager
returns nil.

BugzID 46518

Fixes #132

reviewer @gadamc 
reviewer @indisoluble

### DIFF
--- a/Classes/common/CDTReplicator/CDTReplicatorFactory.m
+++ b/Classes/common/CDTReplicator/CDTReplicatorFactory.m
@@ -45,9 +45,16 @@ static NSString *const CDTReplicatorFactoryErrorDomain = @"CDTReplicatorFactoryE
 {
     self = [super init];
     if (self) {
-        self.manager = dsManager;
-        TD_DatabaseManager *dbManager = dsManager.manager;
-        self.replicatorManager = [[TDReplicatorManager alloc] initWithDatabaseManager:dbManager];
+        
+        if(dsManager){
+            _manager = dsManager;
+            TD_DatabaseManager *dbManager = dsManager.manager;
+            _replicatorManager = [[TDReplicatorManager alloc] initWithDatabaseManager:dbManager];
+        } else {
+            self = nil;
+            CDTLogWarn(CDTREPLICATION_LOG_CONTEXT,
+                       @"Datastore manager is nil, there isn't a local datastore to replicate with.");
+        }
     }
     return self;
 }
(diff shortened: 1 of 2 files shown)


gadamc:
Looks good. My only comment is if we think we should an a `withError:NSError **` argument to the method... That is, create a new method with the `withError` and mark this one as deprecated at some point in the future. (Or should I issue a separate FB ticket for that?)

Classes/common/CDTReplicator/CDTReplicatorFactory.m
```diff
@@ -45,9 +45,16 @@ - (id)initWithDatastoreManager:(CDTDatastoreManager *)dsManager
 {
     self = [super init];
     if (self) {
-        self.manager = dsManager;
-        TD_DatabaseManager *dbManager = dsManager.manager;
-        self.replicatorManager = [[TDReplicatorManager alloc] initWithDatabaseManager:dbManager];
+        
+        if(dsManager){
+            self.manager = dsManager;
```
indisoluble: It is better not to call `self` inside `init`, instead use the ivars. There is a good explanation [here](http://qualitycoding.org/objective-c-init/).

rhyshort: Ah did not notice that in the old code, thanks

rhyshort:
@gadamc I think a new FB ticket for the new API, but I agree an API some kind of error indication outside of the logs is the way to go

Classes/common/CDTReplicator/CDTReplicatorFactory.m
rhyshort: fixed in 4457af23f24b9b7a1523f9ca8f099152653f1d96

indisoluble:
In that case, if there will be a new FB ticket to modify the API, this PR looks good to me :+1:

gadamc:
@indisoluble I thought that init method looked funny for some reason... couldn't put my finger on it. ivars!
+1

PR #137, opened 2015-04-28
Give CDTDatastore a reference to its Manager

rhyshort:
Added internal property for CDTDatastoreManager to
CDTDatastore. CDTDatastore now requires CDTDatastoreManager
to be provided during it's init method.

BugzID 42816

reviewer @mikerhodes 
reviewer @indisoluble

### DIFF
--- a/Classes/common/CDTDatastore.h
+++ b/Classes/common/CDTDatastore.h
@@ -14,6 +14,7 @@
 //  and limitations under the License.
 
 #import <Foundation/Foundation.h>
+#import "CDTDatastoreManager.h"
 
 @class CDTDocumentRevision;
 @class CDTDocumentBody;
@@ -70,7 +71,15 @@ extern NSString *const CDTDatastoreChangeNotification;
 
 + (NSString *)versionString;
 
-- (id)initWithDatabase:(TD_Database *)database;
+/**
+ *
+ * Creates a CDTDatastore instance.
+ *
+ * @param manager this datastore's maanger, must not be nil.
+ * @param database the database where this datastore should save documents
+ *
+ */
+- (instancetype)initWithManager:(CDTDatastoreManager *)manager database:(TD_Database *)database;
 
 /**
  * The number of document in the datastore.
--- a/Classes/common/CDTDatastore.m
+++ b/Classes/common/CDTDatastore.m
@@ -39,6 +39,7 @@ NSString *const CDTDatastoreChangeNotification = @"CDTDatastoreChangeNotificatio
 
 @interface CDTDatastore ()
 
+@property (readonly) CDTDatastoreManager *manager;
 - (void)TDdbChanged:(NSNotification *)n;
 - (BOOL)validateBodyDictionary:(NSDictionary *)body error:(NSError *__autoreleasing *)error;
 
@@ -50,10 +51,11 @@ NSString *const CDTDatastoreChangeNotification = @"CDTDatastoreChangeNotificatio
 
 + (NSString *)versionString { return @CLOUDANT_SYNC_VERSION; }
 
-- (id)initWithDatabase:(TD_Database *)database
+- (instancetype)initWithManager:(CDTDatastoreManager *)manager database:(TD_Database *)database
 {
     self = [super init];
     if (self) {
+        NSParameterAssert(manager);
         _database = database;
         if (![_database open]) {
             return nil;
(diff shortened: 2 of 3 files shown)


Classes/common/CDTDatastore.h
```diff
@@ -70,7 +71,7 @@ extern NSString *const CDTDatastoreChangeNotification;
 
 + (NSString *)versionString;
 
-- (id)initWithDatabase:(TD_Database *)database;
+- (id)initWithManager:(CDTDatastoreManager *) manager Database:(TD_Database *)database;
```
indisoluble: I think the first `D` in `Database:` should be lowercase. Rename `initWithManager:Database:` to `initWithManager:database:`

Classes/common/CDTDatastore.m
```diff
@@ -39,6 +39,7 @@
 
 @interface CDTDatastore ()
 
+@property CDTDatastoreManager *manager;
```
indisoluble: Will 'manager' change after the object is initialised? If not and although the property is private, I think we should make it `readonly`.

Classes/common/CDTDatastore.m
```diff
@@ -63,6 +64,7 @@ - (id)initWithDatabase:(TD_Database *)database
         NSString *name = [database name];
         _extensionsDir =
             [dir stringByAppendingPathComponent:[NSString stringWithFormat:@"%@_extensions", name]];
+        _manager = manager;
```
indisoluble: There is not validation here (check the object is not nil). Does it mean that the property is optional? If that is the case, should we keep the old initialiser (`initWithDatabase:`)? Or rename the parameters this way: `- (id)initWithManager:(CDTDatastoreManager *) managerOrNil Database:(TD_Database *)database`.

rhyshort: it shouldn't be `nil` but I've opted not to validate it since CDTDatastore objects should only be created from CDTDatastoreManager

Classes/common/CDTDatastore.m
rhyshort: fixed in e97a7b3

Classes/common/CDTDatastore.h
rhyshort: fixed in e97a7b3

Classes/common/CDTDatastore.m
indisoluble: OK, I see your point. I can not even find a unit test that creates a `CDTDatastore` but in the [branch](https://github.com/cloudant/CDTDatastore/tree/feature-encryption) I'm working on right now, I also modified this `init` method to supply a key provider and I ensure that the provider is always informed with an `Assert`. I also added a few tests that create a `CDTDatastore` without using a `CDTDatastoreManager` and check that I receive an exception if the provider is not informed.

I think we need a second opinion here, @mikerhodes what do you think?

mikerhodes: We should probably get better at validating pre-conditions, and documenting them. At least document if a parameter is required, but asserting is better (particularly as the macros are compiled out in production builds if I recall).

Perhaps [NSParameterAssert](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Miscellaneous/Foundation_Functions/#//apple_ref/c/macro)./NSParameterAssert

Classes/common/CDTDatastore.m
```diff
@@ -50,7 +51,7 @@ @implementation CDTDatastore
 
 + (NSString *)versionString { return @CLOUDANT_SYNC_VERSION; }
 
-- (id)initWithDatabase:(TD_Database *)database
+- (id)initWithManager:(CDTDatastoreManager *) manager database:(TD_Database *)database
```
mikerhodes: Correct spacing.

mikerhodes: Also may as well fix to return instancetype while we're here.

Classes/common/CDTDatastore.h
```diff
@@ -70,7 +71,15 @@ extern NSString *const CDTDatastoreChangeNotification;
...
-- (id)initWithDatabase:(TD_Database *)database;
+/**
+ *
+ * Creates a CDTDatastore instance.
+ *
+ * @param manager this datastore's maanger, must not be nil.
+ * @param database the database where this datastore should save documents
+ *
+ */
+- (id)initWithManager:(CDTDatastoreManager *) manager database:(TD_Database *)database;
```
indisoluble: Spacing should be fixed here too. Also it should return `instancetype `.

rhyshort: done in 96d5f87

indisoluble:
:+1:

Classes/common/CDTDatastore.m
rhyshort: Done in cefb00e

mikerhodes:
:+1: squash and merge.

PR #138, opened 2015-04-28
46517 Check if a db is encrypted or not trying to read it

indisoluble:
*What:*
If a key provider does not return a value, we ensure that the database is not encrypted comparing the first 15 bytes of the file with 'SQLite format 3'. If the SQLite file does not start with that string, we assume that the db is not encrypted. But if the provider does return a key, we set the key and then we execute this select to check that the provided value is correct:

```
SELECT count(*) FROM sqlite_master;
```

However this select is going to fail if no key is set and the database in encrypted.
Instead of reading the beginning of the file, we can execute this select in all cases.

*Why:*
Because comparing the beginning of the file with a hardcoded string is not reliable, it could change with the next version of SQLite.

*How:*
We have to execute the `select` mentioned before in all cases: if a key is set or not.

*Tests:*
The tests already in place are sufficient 

reviewer @mikerhodes 
reviewer @rhyshort 

**NOTE:** Wait to review this PR until PR #133  is closed. This branch will be rebased and there will be only one commit to review.

### DIFF
--- a/Classes/common/Encryption/FMDatabase+EncryptionKey.h
+++ b/Classes/common/Encryption/FMDatabase+EncryptionKey.h
@@ -1,0 +1,47 @@
+//
+//  FMDatabase+EncryptionKey.h
+//  CloudantSync
+//
+//  Created by Enrique de la Torre Fernandez on 27/04/2015.
+//  Copyright (c) 2015 IBM Cloudant. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+//  except in compliance with the License. You may obtain a copy of the License at
+//  http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software distributed under the
+//  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+//  either express or implied. See the License for the specific language governing permissions
+//  and limitations under the License.
+//
+
+#import "FMDatabase.h"
+
+#import "CDTEncryptionKeyProvider.h"
+
+extern NSString *const FMDatabaseEncryptionKeyErrorDomain;
+
+/**
+ Errors for 'setKeyWithProvider:error:'
+ */
+typedef NS_ENUM(NSInteger, FMDatabaseEncryptionKeyError) {
+    FMDatabaseEncryptionKeyErrorKeyNotSet,
+    FMDatabaseEncryptionKeyErrorNoKeyProvidedForEncryptedDB,
+    FMDatabaseEncryptionKeyErrorWrongKeyOrDBNotEncrypted
+};
+
+@interface FMDatabase (EncryptionKey)
+
+/**
+ This method performs 3 steps:
+ - Get an encryption key from the provider
+ - Set the key (if there is any, a provider can return nil)
+ - Check that the db can be read
+
+ @param provider Returns the key to decipher the db (or nil)
+ @param error Output param, it will contain an error if the method does not succeed
+
+ @return `YES` if success, `NO` on error.
+ */
+- (BOOL)setKeyWithProvider:(id<CDTEncryptionKeyProvider>)provider error:(NSError **)error;
+
+@end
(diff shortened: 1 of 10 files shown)


Classes/common/Encryption/FMDatabase+EncryptionKey.m
```diff
@@ -0,0 +1,88 @@
...
+
+    // Set the key (if there is any)
+    if (encryptionKey) {
+        success = [self setKey:encryptionKey];
+        if (!success) {
+            CDTLogError(CDTDATASTORE_LOG_CONTEXT,
+                        @"Key to de-encrypt DB at %@ not set. DB can not be opened.",
+                        [self databasePath]);
+
+            NSString *desc = NSLocalizedString(@"Key to de-encrypt DB not set", nil);
```
rhyshort: `decrypt` instead of `de-encrypt`

rhyshort:
Other than that minor comment, looks good, +1

Classes/common/Encryption/FMDatabase+EncryptionKey.h
```diff
@@ -0,0 +1,47 @@
...
+extern NSString *const FMDatabaseEncryptionKeyErrorDomain;
+
+/**
+ Errors for 'setKeyWithProvider:error:'
+ */
+typedef NS_ENUM(NSInteger, FMDatabaseEncryptionKeyError) {
+    FMDatabaseEncryptionKeyErrorKeyNotSet,
+    FMDatabaseEncryptionKeyErrorNoKeyProvidedForEncryptedDB,
+    FMDatabaseEncryptionKeyErrorWrongKeyOrDBNotEncrypted
+};
```
mikerhodes: These end up in the public error param right? So should we declare them somewhere else?

Classes/common/query/CDTQIndexManager.m
```diff
@@ -453,80 +447,29 @@ + (FMDatabaseQueue *)databaseQueueWithDatastore:(CDTDatastore *)datastore
...
+                NSLocalizedString(@"Problem configuring database with encryption key", nil)
+        };
 
-            *error = [NSError errorWithDomain:CDTQIndexManagerErrorDomain
-                                         code:CDTQIndexErrorEncryptionKeyError
-                                     userInfo:userInfo];
-        }
+        *error = [NSError errorWithDomain:CDTQIndexManagerErrorDomain
+                                     code:CDTQIndexErrorEncryptionKeyError
+                                 userInfo:userInfo];
```
mikerhodes: I assume cleanup of `database` is done elsewhere?

mikerhodes:
Looks sound, could  you check my two comments?

Classes/common/Encryption/FMDatabase+EncryptionKey.m
indisoluble: Fixed in c75dd85

Classes/common/Encryption/FMDatabase+EncryptionKey.h
indisoluble: 'setKeyWithProvider:error:' is used in 2 different places:
* `CDTQIndexManager`.- It defines its own set of errors: `CDTQQueryError`. I can extend this type and if 'setKeyWithProvider:error:'  fails, parse the error returned to the new errors in `CDTQQueryError`.
* `TD_Database`.- It does not pass the error returned by 'setKeyWithProvider:error:' to other methods in the library, to say in another way, it is self-contained so there is no need to declare the errors in `FMDatabase+EncryptionKey` in other place.

Classes/common/query/CDTQIndexManager.m
indisoluble: If the configuration does not succeed, the `database` is set to nil [here](https://github.com/cloudant/CDTDatastore/blob/46517-read-db-to-check-if-encrypted/Classes/common/query/CDTQIndexManager.m#L440)

Classes/common/Encryption/FMDatabase+EncryptionKey.h
mikerhodes: Okay, looks like this doesn't end up the public domain, sorry.

mikerhodes:
:+1:

PR #140, opened 2015-05-11
46372 Remove access to attachment file path

indisoluble:
*What:*
Remove access to attachment file path.

*Why:*
We intent to encrypt the entire content in a datastore, that includes databases and attachments. The encryption for databases is already in place but we are still working on the attachments. The first step to reach this goal is to remove the direct access to the attachments through their paths. We can not allow other parts of the library to read an attachment as if it were a *normal* file because, it the attachment is encrypted, the data will not make any sense (obviously).

*How:*
1. Remove/hide method 'TDBlobStore:pathForKey:'.
2. Define protocol to access the content in an attachment: `CDTBlobReader`.
3. Define protocol to save data in an attachment: `CDTBlobWriter`
4. Implement class that conforms to both protocol: `CDTBlobData`. This way all the code related to access an attachment is in the same place.
5. Modify library to use the new class instead of accessing the files using their paths.

**Check case with BugzID 46372 for more details about the design and how it has evolved until the current point.**

*Tests:*
Tests are for the new class `CDTBlobData`, the rest of the code is already under coverage:
* A blob has to be created with a path (although the file does not have to exist on advance).
* A blob is not open after initialisation.
* To open a blob, a file must exist in the specified path.
* Add data to a blob fails if it was not open before.
* A blob has to be open to read its content.
* A blob can not be read if the file does not exist.
* A blob can not be created/overwritten if it is already open.

reviewer @mikerhodes 
reviewer @rhyshort 

**NOTE:** Wait to review this PR until PR #138  is closed. This branch will be rebased and there will be only commits related to this case.

### DIFF
--- a/Classes/common/Attachments/CDTBlobData.m
+++ b/Classes/common/Attachments/CDTBlobData.m
@@ -1,0 +1,206 @@
+//
+//  CDTBlobData.m
+//  CloudantSync
+//
+//  Created by Enrique de la Torre Fernandez on 05/05/2015.
+//  Copyright (c) 2015 IBM Cloudant. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+//  except in compliance with the License. You may obtain a copy of the License at
+//  http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software distributed under the
+//  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+//  either express or implied. See the License for the specific language governing permissions
+//  and limitations under the License.
+//
+
+#import "CDTBlobData.h"
+
+#import "CDTLogging.h"
+
+NSString *const CDTBlobDataErrorDomain = @"CDTBlobDataErrorDomain";
+
+@interface CDTBlobData ()
+
+@property (strong, nonatomic, readonly) NSString *path;
+
+@property (strong, nonatomic) NSFileHandle *outFileHandle;
+
+@end
+
+@implementation CDTBlobData
+
+#pragma mark - Init object
+- (instancetype)init { return [self initWithPath:nil]; }
+
+- (instancetype)initWithPath:(NSString *)path
+{
+    self = [super init];
+    if (self) {
+        if (path && ([path length] > 0)) {
+            _path = path;
+            _outFileHandle = nil;
+        } else {
+            CDTLogError(CDTDATASTORE_LOG_CONTEXT, @"A non-empty path is mandatory");
+
+            self = nil;
+        }
+    }
+
+    return self;
+}
+
+#pragma mark - Memory management
+- (void)dealloc { [self close]; }
+
+#pragma mark - CDTBlobReader methods
+- (NSData *)dataWithError:(NSError **)error
+{
+    NSData *data = nil;
+    NSError *thisError = nil;
+
+    if ([self isBlobOpenForWriting]) {
+        CDTLogDebug(CDTDATASTORE_LOG_CONTEXT,
+                    @"Blob at %@ is open. Close it before reading its content", self.path);
+
+        thisError = [CDTBlobData errorOperationNotPossibleIfBlobIsOpen];
+    } else {
+        data = [NSData dataWithContentsOfFile:self.path
+                                      options:NSDataReadingMappedIfSafe
+                                        error:&thisError];
+        if (!data) {
+            CDTLogDebug(CDTDATASTORE_LOG_CONTEXT,
+                        @"Data object could not be created with file %@: %@", self.path, thisError);
+        }
+    }
+
+    if (!data && error) {
+        *error = thisError;
+    }
+
+    return data;
+}
+
+- (NSInputStream *)inputStreamWithOutputLength:(UInt64 *)outputLength
+{
+    if ([self isBlobOpenForWriting]) {
+        CDTLogDebug(CDTDATASTORE_LOG_CONTEXT, @"Close blob in order to create an input stream");
+
+        return nil;
+    }
+
+    NSFileManager *defaultManager = [NSFileManager defaultManager];
+    if (![defaultManager fileExistsAtPath:self.path isDirectory:NO]) {
+        CDTLogDebug(CDTDATASTORE_LOG_CONTEXT, @"No file found in %@", self.path);
+        
+        return nil;
+    }
+    
+    NSInputStream *inputStream = [NSInputStream inputStreamWithFileAtPath:self.path];
+
+    if (inputStream && outputLength) {
+        NSError *error = nil;
+        NSDictionary *info = [defaultManager attributesOfItemAtPath:self.path error:&error];
+        if (info) {
+            *outputLength = [info fileSize];
+        } else {
+            CDTLogDebug(CDTDATASTORE_LOG_CONTEXT,
+                        @"Attributes for file %@ could not be obtained: %@", self.path, error);
+
+            inputStream = nil;
+        }
+    }
+
+    return inputStream;
+}
+
+#pragma mark - CDTBlobWriter methods
+- (BOOL)writeEntireBlobWithData:(NSData *)data error:(NSError **)error
+{
+    BOOL success = NO;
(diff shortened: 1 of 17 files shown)


indisoluble:
**NOTE:** Hide property `path` in `TDBlobStore`.

rhyshort:
Looks good to me, :+1:

Classes/common/Attachments/CDTBlobData.m
```diff
@@ -0,0 +1,206 @@
...
+    if (!success && error) {
+        *error = thisError;
+    }
+
+    return success;
+}
+
+- (BOOL)isBlobOpen { return (self.outFileHandle != nil); }
+
+- (BOOL)openBlobToAddData
```
mikerhodes: Should `Add` be `Append`? It's more specific, I'm a little unclear whether this method is for:

- Creating a new file;
- Appending an existing file;
- Overwriting an existing file.

Reading the protocol doc and the implementation has left me thinking perhaps this should be `Append`.

Classes/common/Attachments/CDTBlobReader.h
```diff
@@ -0,0 +1,46 @@
...
+/**
+ Load the content of an attachment in a NSData instance.
+ 
+ @param error Output param that will point to an error (in case there is any)
+ 
+ @return Content of the attachment as a NSData instance or nil if there is an error
+ 
+ @warning Notice that this is a synchronous operation. The entire file is read before returning a
+ NSData instance; if the attachment is big, this operatin will take a while and a lot of memory will
+ be allocated (keep an eye on memory warnings).
```
mikerhodes: "@warning This operation is synchronous and reads the file content into memory before returning; this may cause issues with larger files."?

Classes/common/Attachments/CDTBlobReader.h
```diff
@@ -0,0 +1,46 @@
...
+ */
+- (NSData *)dataWithError:(NSError **)error;
+
+/**
+ Create an input stream to an attachment.
```
mikerhodes: "to read from"

Classes/common/Attachments/CDTBlobWriter.h
```diff
@@ -0,0 +1,66 @@
...
+ Define the methods to store data in an attachment without exposing its path.
+ */
+@protocol CDTBlobWriter <NSObject>
+
+/**
+ Overwrite the content of an attachment with the data provided as a parameter. If the file does
+ not exist, it will create it.
+ 
+ Notice that this method has to work properly with the other methods defined in this protocol, i.e.
+ will it succeed or will it fail if the blob is open?
```
mikerhodes: Probably want to clarify what you chose :)

Classes/common/Attachments/CDTBlobWriter.h
```diff
@@ -0,0 +1,66 @@
...
+ 
+ @param data Data to store in the attachment
+ @param error Output param that will point to an error (if there is any)
+ 
+ @return YES (if the operation succeed) or NO (if there is an error)
+ */
+- (BOOL)createBlobWithData:(NSData *)data error:(NSError **)error;
```
mikerhodes: `create` is misleading, given it will overwrite existing on disk content.

Perhaps `writeEntireBlobWithData:error:`?

Classes/common/Attachments/CDTBlobWriter.h
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ @return YES if the attachment was open before or NO in other case
+ */
+- (BOOL)isBlobOpen;
```
mikerhodes: I think be specific here: `isBlobOpenForWriting`.

Classes/common/Attachments/CDTBlobWriter.h
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ Prepare an attachment to write data in it.
+ 
+ @return YES if the attachment was open (or it was already open) or NO in other case
+ */
+- (BOOL)openBlobToAddData;
```
mikerhodes: I'd possibly give this family slightly different names, specifying writing. Also, read methods don't have blob in the name, do writing methods need it?

- `openForWriting`
- `appendData:`
- `close`

Classes/common/touchdb/TD_Database+Attachments.h
```diff
@@ -55,16 +58,16 @@ typedef enum { kTDAttachmentEncodingNone, kTDAttachmentEncodingGZIP } TDAttachme
...
+                                            named:(NSString *)filename
+                                             type:(NSString **)outType
+                                         encoding:(TDAttachmentEncoding *)outEncoding
+                                           status:(TDStatus *)outStatus;
 
 /** Uses the "digest" field of the attachment dict to look up the attachment in the store and return
- * a file URL to it. DO NOT MODIFY THIS FILE! */
-- (NSURL *)fileForAttachmentDict:(NSDictionary *)attachmentDict;
+ * its blob. */
+- (id<CDTBlobReader>)blobForAttachmentDict:(NSDictionary *)attachmentDict;
```
mikerhodes: I wonder if there should be separate reader/writer classes to prevent modification of the underlying blob data by people asking for `reader` objects.

mikerhodes:
Logic looks good, a few comments on naming and the use of a single reader/writer class.

Classes/common/Attachments/CDTBlobData.m
indisoluble: I'll change the names based on your [comment below](https://github.com/cloudant/CDTDatastore/pull/140#discussion_r31316858).

Classes/common/Attachments/CDTBlobReader.h
indisoluble: Fixed in d45aa59

Classes/common/Attachments/CDTBlobReader.h
indisoluble: Fixed in b9101a7

Classes/common/Attachments/CDTBlobWriter.h
indisoluble: This comment is fixed [here](https://github.com/cloudant/CDTDatastore/pull/153/files#diff-d4043e8e82b45b72082555c7befc53ecR28) in PR #153

Classes/common/Attachments/CDTBlobWriter.h
indisoluble: Fixed in fddd736

Classes/common/Attachments/CDTBlobWriter.h
indisoluble: Fixed in ec3052c

Classes/common/Attachments/CDTBlobData.m
indisoluble: Fixed in 257bb63

Classes/common/Attachments/CDTBlobWriter.h
indisoluble: Fixed in beb7ea7, 257bb63 & 9527be7

Classes/common/touchdb/TD_Database+Attachments.h
indisoluble: Why would someone do that? I mean, when you received an `id<SomeProtocol>`, you know that you should only use the methods defined in `SomeProtocol` because that is the documented behaviour. In other case, after the next update of the library, strange things may start to happen.

Also, I do not think we would solve this problem by diving the class. The method `TDBlobStore:pathForKey:` is not available in the public interface, however it is still possible to call it using `performSelector:withObject:`. Once you have the path to the attachment, you can create a `CDTBlobWriter` instance.

indisoluble:
@mikerhodes @rhyshort Based on [this comment](https://github.com/cloudant/sync-android/pull/146#discussion_r31312135), I'm going to add a new commit to this PR. It is not directly related with these changes but I think I should merge ASAP into `feature-encryption`

Classes/common/touchdb/TD_Database+Attachments.h
mikerhodes: Agreed. Leave as is.

mikerhodes:
I think this looks good now, +1.

PR #141, opened 2015-05-11
Add full text query support

alfinkel:
_What_

This PR adds support for the $text and $search operators in Query.

_Why_

The change is to provide full text search capabilities in Query.

_How_

The normalization/validation process is updated to handle the new operators and the SQL translation process is updated to ensure correct generation of SQL and node placement in the query node tree.

_Notes_

This PR does not include the unit tests. Unit tests will be included in a following PR. Logic has been temporarily added to deactivate the full text search logic until the unit tests are delivered.

reviewer @mikerhodes 
reviewer @rhyshort

BugId: 46775

### DIFF
--- a/Classes/common/query/CDTQQuerySqlTranslator.m
+++ b/Classes/common/query/CDTQQuerySqlTranslator.m
@@ -25,6 +25,8 @@
 @property (nonatomic) BOOL atLeastOneIndexUsed;       // if NO, need to generate a return all query
 @property (nonatomic) BOOL atLeastOneIndexMissing;    // i.e., we need to use posthoc matcher
 @property (nonatomic) BOOL atLeastOneORIndexMissing;  //       we need to use posthoc matcher
+@property (nonatomic) BOOL textIndexRequired;         // A text index needed for a text search
+@property (nonatomic) BOOL textIndexMissing;          // if NO and is required, cannot perform query
 
 @end
 
@@ -73,10 +75,19 @@
     CDTQQueryNode *node =
         [CDTQQuerySqlTranslator translateQuery:query toUseIndexes:indexes state:state];
 
-    // If we haven't used a single index, we need to return a query
-    // which returns every document, so the posthoc matcher can
-    // run over every document to manually carry out the query.
-    if (!state.atLeastOneIndexUsed || state.atLeastOneORIndexMissing) {
+    if (state.textIndexMissing) {
+        LogError(@"No text index defined, cannot execute query containing a text search.");
+        return nil;
+    } else if (state.textIndexRequired && state.atLeastOneIndexMissing) {
+        LogError(@"Query %@ contains a text search but is missing json index(es).  "
+                  "All indexes must exist in order to execute a query containing a text search.  "
+                  "Create all necessary indexes for the query and re-execute.", query);
+        return nil;
+    } else if (!state.textIndexRequired &&
+                  (!state.atLeastOneIndexUsed || state.atLeastOneORIndexMissing)) {
+        // If we haven't used a single index, we need to return a query
+        // which returns every document, so the posthoc matcher can
+        // run over every document to manually carry out the query.
         NSSet *neededFields = [NSSet setWithObject:@"_id"];
         NSString *allDocsIndex =
             [CDTQQuerySqlTranslator chooseIndexForFields:neededFields fromIndexes:indexes];
@@ -126,22 +137,26 @@
         root = [[CDTQOrQueryNode alloc] init];
     }
 
-    //
-    // First handle the simple @"field": @{ @"$operator": @"value" } clauses. These are
-    // handled differently for AND and OR parents, so we need to have the conditional
-    // logic below.
-    //
+    // Compile a list of simple clauses to be handled below.  If a text clause is
+    // encountered, store it separately from the simple clauses since it will be
+    // handled later on its own.
 
     NSMutableArray *basicClauses = [NSMutableArray array];
-
+    __block NSObject *textClause = nil;
+    
     [clauses enumerateObjectsUsingBlock:^(id obj, NSUInteger idx, BOOL *stop) {
         NSDictionary *clause = (NSDictionary *)obj;
         NSString *field = clause.allKeys[0];
         if (![field hasPrefix:@"$"]) {
             [basicClauses addObject:clauses[idx]];
+        } else if ([field.lowercaseString isEqualToString:TEXT]) {
+            textClause = clauses[idx];
         }
     }];
 
+    // Handle the simple "field": { "$operator": "value" } clauses. These are
+    // handled differently for AND and OR parents, so we need to have the
+    // conditional logic below.
     if (basicClauses.count > 0) {
         if (query[AND]) {
             // For an AND query, we require a single compound index and we generate a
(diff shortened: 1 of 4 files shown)


Classes/common/query/CDTQQuerySqlTranslator.m
```diff
@@ -217,6 +232,29 @@ + (CDTQQueryNode *)translateQuery:(NSDictionary *)query
             }
         }
     }
+    
+    // A text clause such as { "$text" : { "$search" : "foo bar baz" } }
+    // by nature uses its own text index.  It is therefor handled
```
mikerhodes: `therefore`

Classes/common/query/CDTQQuerySqlTranslator.m
```diff
@@ -217,6 +232,29 @@ + (CDTQQueryNode *)translateQuery:(NSDictionary *)query
...
+    // separately from other simple clauses.
+    if (textClause != nil) {
+        state.textIndexRequired = YES;
+        NSString *textIndex = [CDTQQuerySqlTranslator getTextIndexFromIndexes:indexes];
+        if (textIndex == nil || [textIndex length] == 0) {
```
mikerhodes: `length` is a property, use dot notation.

Classes/common/query/CDTQQuerySqlTranslator.m
```diff
@@ -299,6 +335,20 @@ + (NSString *)chooseIndexForFields:(NSSet *)neededFields fromIndexes:(NSDictiona
...
 }
 
++ (NSString *)getTextIndexFromIndexes:(NSDictionary *)indexes
+{
+    NSString *textIndex = nil;
+    for (NSString *indexName in [indexes allKeys]) {
+        NSDictionary *indexDefinition = indexes[indexName];
+        NSString *indexType = indexDefinition[@"type"];
+        if ([indexType.lowercaseString isEqualToString:@"text"]) {
+            textIndex = indexName;
```
mikerhodes: Could `break` here, or return the index.

Classes/common/query/CDTQQuerySqlTranslator.m
```diff
@@ -299,6 +335,20 @@ + (NSString *)chooseIndexForFields:(NSSet *)neededFields fromIndexes:(NSDictiona
     return chosenIndex;
 }
 
++ (NSString *)getTextIndexFromIndexes:(NSDictionary *)indexes
+{
+    NSString *textIndex = nil;
+    for (NSString *indexName in [indexes allKeys]) {
+        NSDictionary *indexDefinition = indexes[indexName];
+        NSString *indexType = indexDefinition[@"type"];
```
mikerhodes: I'd probably use a single line here as you have elsewhere, `NSString *type = indexes[indexName][@"type"]`.

Classes/common/query/CDTQQuerySqlTranslator.m
```diff
@@ -495,4 +545,32 @@ + (CDTQSqlParts *)selectStatementForAndClause:(NSArray *)clause usingIndex:(NSSt
...
+    if (![clause isKindOfClass:[NSDictionary class]]) {
+        return nil; // should never get here as this would not pass normalization
+    }
+    
+    NSDictionary *textClause = (NSDictionary *) clause;
+    NSDictionary *searchClause = textClause[TEXT];
+    
+    NSString *tableName = [CDTQIndexManager tableNameForIndex:indexName];
+    NSString *search = [searchClause[SEARCH] stringByReplacingOccurrencesOfString:@"'"
+                                                                       withString:@"''"];
```
mikerhodes: Perhaps note this is escaping for SQLite?

mikerhodes: And is this required if we're using a placeholder for it? I guess it was for Java?

Classes/common/query/CDTQQueryValidator.m
```diff
@@ -300,7 +300,18 @@ + (NSArray *)compressMultipleNotOperators:(NSArray *)clause
 
 #pragma validation class methods
 
+/**
+ * This method runs the list of clauses making up the selector through a series of
+ * validation steps and returns whether the clause list is valid or not.
```
mikerhodes: Maybe add a note that it logs the error if not valid?

Classes/common/query/CDTQQueryValidator.m
```diff
@@ -392,12 +453,26 @@ + (BOOL)validatePredicateValue:(NSObject *)predicateValue forOperator:(NSString
...
     }
 }
 
++ (BOOL)validateTextSearchArgument:(NSObject *)textSearch
+{
+    BOOL valid = YES;
+    
+    if(![textSearch isKindOfClass:[NSString class]]){
+        valid = NO;
+        LogError(@"$search operator expects an NSString");
```
mikerhodes: Replace `expects` with `requires`. Just a little clearer.

Classes/common/query/CDTQQuerySqlTranslator.m
alfinkel: Fixed in 09d8a4e530f4bd4a50e24d1ae14149b1e27dd27d

Classes/common/query/CDTQQuerySqlTranslator.m
alfinkel: Fixed in 09d8a4e530f4bd4a50e24d1ae14149b1e27dd27d

Classes/common/query/CDTQQuerySqlTranslator.m
alfinkel: Added `break` in 09d8a4e530f4bd4a50e24d1ae14149b1e27dd27d

Classes/common/query/CDTQQuerySqlTranslator.m
alfinkel: Changed in 09d8a4e530f4bd4a50e24d1ae14149b1e27dd27d

Classes/common/query/CDTQQuerySqlTranslator.m
alfinkel: The short answer here is that the escaping is not necessary in either iOS or Android.  The functionality is not affected by the escaping but...

Removed escaping in 09d8a4e530f4bd4a50e24d1ae14149b1e27dd27d.

Added FB case 47078 to remove from sync-android.

Classes/common/query/CDTQQueryValidator.m
alfinkel: Done in 09d8a4e530f4bd4a50e24d1ae14149b1e27dd27d

Classes/common/query/CDTQQueryValidator.m
alfinkel: Done in 09d8a4e530f4bd4a50e24d1ae14149b1e27dd27d

Classes/common/query/CDTQQuerySqlTranslator.m
```diff
@@ -495,4 +545,31 @@ + (CDTQSqlParts *)selectStatementForAndClause:(NSArray *)clause usingIndex:(NSSt
...
+    if (!clause) {
+        return nil;  // no query here
+    }
+    
+    if (!indexName) {
+        return nil;
+    }
+    
+    if (![clause isKindOfClass:[NSDictionary class]]) {
+        return nil; // should never get here as this would not pass normalization
```
rhyshort: Since this should never happen, I'm wondering if we should just let it fail with doesn't respond to selector errors. Failing hard would make it easier to find the issue if someone breaks normalisation. So I'm thinking remove the check and leave a comment like:
```
Should never be anything but a dictionary, anything else should fail normalisation
```

Thoughts @alfinkel ?

alfinkel: A valid point @rhyshort.  I've changed the method to expect an NSDictionary instead of an NSObject and I've updated the calling code block to cast the clause as an NSDictionary.  Since this is how the other basic clauses are handled.  I've also added a comment there as you suggested.  Done in 66a688789ff1d951b6b921ba97852d52a6e0b903.

rhyshort: Okay cool

rhyshort:
+1 :+1:

mikerhodes:
Looks good, :+1:

PR #144, opened 2015-05-13
46425 Deactivate key derivation in SQLCipher

indisoluble:
*What:*
Set the key to cipher the database with [PRAGMA key = "x'...'"](https://www.zetetic.net/sqlcipher/sqlcipher-api/#key), instead of [sqlite3_key](https://www.zetetic.net/sqlcipher/sqlcipher-api/#sqlite3_key).

*Why:*
`sqlite3_key` does the same that `PRAGMA key = '...'`, this means that the key provided is interpreted as a **passphrase** and converted to a **key using PBKDF2 key derivation**. But the key that we provide is already strong and it can and must be used as it is. With `PRAGMA key = "x'...'"` we can pass our key to SQLCipher and no further transformation will be performed.

*How:*
1. SQLCipher requires a 32 bytes length key. Add a new class named `CDTEncryptionKey` to enforce that the key has the right size. `CDTEncryptionKeyProvider` will return a `CDTEncryptionKey` instance instead of a string.
2. `CDTEncryptionKey` has to return a buffer, not a string. We will pass the key to SQLCipher as a hexadecimal string but later on we will use this same key to cipher the attachments, in this case we need the buffer.
3. Update the code to set the key with `PRAGMA key = "x'...'"`. *NOTE*: If we execute this command on SQLite, it will not fail; we have to detect if SQLCipher is included (use macro `ENCRYPT_DATABASE `) and act accordingly. 

**Check case with BugzID 46425 for more details about the design and how it has evolved until the current point.**

*Tests:*
No more tests added, only updated to work with `CDTEncryptionKey`.

reviewer @mikerhodes 
reviewer @rhyshort 

**NOTE:** Wait to review this PR until PR #140  is closed. This branch will be rebased and there will be only commits related to this case.

### DIFF
--- a/Classes/common/Encryption/CDTEncryptionKey.h
+++ b/Classes/common/Encryption/CDTEncryptionKey.h
@@ -1,0 +1,52 @@
+//
+//  CDTEncryptionKey.h
+//  CloudantSync
+//
+//  Created by Enrique de la Torre Fernandez on 12/05/2015.
+//  Copyright (c) 2015 IBM Cloudant. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+//  except in compliance with the License. You may obtain a copy of the License at
+//  http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software distributed under the
+//  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+//  either express or implied. See the License for the specific language governing permissions
+//  and limitations under the License.
+//
+
+#import <Foundation/Foundation.h>
+
+/**
+ Use this class to pass around a DPK (Data Protection Key).
+
+ This class is essentially a wrapper for a NSData instance. However this instance has to have a
+ specific size: CDTENCRYPTIONKEY_KEYSIZE bytes. The reason for this is that we rely on SQLCipher to
+ encrypt databases and it requires a key of CDTENCRYPTIONKEY_KEYSIZE bytes.
+ */
+
+// Size (bytes) of the DPK (Data Protection Key)
+#define CDTENCRYPTIONKEY_KEYSIZE 32
+
+@interface CDTEncryptionKey : NSObject
+
+/**
+ Array of CDTENCRYPTIONKEY_KEYSIZE bytes with the DPK
+ */
+@property (assign, nonatomic, readonly) const void *bytes;
+
+- (instancetype)init UNAVAILABLE_ATTRIBUTE;
+
+/**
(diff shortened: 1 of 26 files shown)


indisoluble:
**NOTE:** Make a copy of the data provided to `CDTEncryptionKey` (like in [Android](https://github.com/cloudant/sync-android/commit/2a9251ab1a686e1c441d2513cad3833a65d3ce1b)). Also return a copy when requested.

Classes/common/Encryption/CDTEncryptionKey.h
```diff
@@ -0,0 +1,52 @@
...
+- (instancetype)init UNAVAILABLE_ATTRIBUTE;
+
+/**
+ Initialise an encryption key with a buffer.
+
+ @param data Buffer of size CDTENCRYPTIONKEY_KEYSIZE bytes with the DPK
+ 
+ @warning If data.length is not CDTENCRYPTIONKEY_KEYSIZE, init will return nil
+ */
+- (instancetype)initWithData:(NSData *)data NS_DESIGNATED_INITIALIZER;
```
rhyshort: `NS_DESIGNATED_INITIALIZER` isn't available for platforms below 10.10, it should either be removed or write and ifdef as described here http://stackoverflow.com/questions/24119896/ns-designated-initializer-expected-colon

Tests/Tests/Helpers/CDTHelperFixedKeyProvider.h
```diff
@@ -19,6 +19,8 @@
 
 @interface CDTHelperFixedKeyProvider : NSObject <CDTEncryptionKeyProvider>
 
-- (id)initWithKey:(NSString *)key;
+- (instancetype)initWithKey:(NSData *)key NS_DESIGNATED_INITIALIZER;
```
rhyshort: `NS_DESIGNATED_INITIALIZER` used here too

Classes/common/Encryption/CDTEncryptionKey.h
rhyshort: You should checked `UNAVAILABLE_ATTRIBUTE` it make sure its available on the platforms we support / build for.

rhyshort:
other than those comments, it looks fine to me +1

Tests/Tests/Helpers/CDTHelperFixedKeyProvider.h
indisoluble: Fixed in 79c1aba

Classes/common/Encryption/CDTEncryptionKey.h
indisoluble: Fixed in 79c1aba / 020822a

mikerhodes:
Looks good, +1.

PR #147, opened 2015-05-18
Add ensureIndexed methods to datastore interface

alfinkel:
_What_

The CDTDatastore+Query category is missing ensureIndexed methods that allow a developer to specify index type and index settings.  These methods need to be visible to the developer.

_Why_

In order to allow developers to utilize text search functionality they must have the ability to create TEXT indexes.  Without these additional methods, it is unclear as to how this can be done.

_How_ 

Add the following methods to CDTDatastore+Query

```objc
- (NSString *)ensureIndexed:(NSArray * /* NSString */)fieldNames
                   withName:(NSString *)indexName
                       type:(NSString *)type;
...
- (NSString *)ensureIndexed:(NSArray * /* NSString */)fieldNames
                   withName:(NSString *)indexName
                       type:(NSString *)type
                   settings:(NSDictionary *)indexSettings;
```

reviewer @mikerhodes 
reviewer @rhyshort 

BugId: 47232

### DIFF
--- a/Classes/common/query/CDTDatastore+Query.h
+++ b/Classes/common/query/CDTDatastore+Query.h
@@ -57,6 +57,39 @@
  */
 - (NSString *)ensureIndexed:(NSArray * /* NSString */)fieldNames
                    withName:(NSString *)indexName;
+/**
+ Create a new index based on an index type over a set of fields.
+ 
+ Index type can be either "json" or "text".  A TEXT index provides
+ the ability to perform text searches.
+ */
+- (NSString *)ensureIndexed:(NSArray * /* NSString */)fieldNames
+                   withName:(NSString *)indexName
+                       type:(NSString *)type;
+
+/**
+ Create a new index based on an index type with specific index 
+ settings over a set of fields.
+ 
+ Index settings currenly only apply to a TEXT index.
+ 
+ An example:
+ 
+ Where ds is your datastore and fields is an array of fieldnames...
+ 
+ [ds ensureIndexed: fields 
+          withName: @"text_idx"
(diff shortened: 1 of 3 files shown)


rhyshort:
+1

Classes/common/query/CDTDatastore+Query.h
```diff
@@ -57,6 +57,38 @@
...
+ 
+     Where ds is your datastore and fields is an array of fieldnames...
+ 
+     [ds ensureIndexed: fields withName: @"text_idx"
+                                   type: @"text"
+                               settings: @{ @"tokenize": @"porter" } ]
+     
+     This will create a TEXT index named text_idx and override the 
+     default tokenizer used to construct the TEXT index with the 
+     "porter" algorithm tokenizer.
```
mikerhodes: This is markdown-based, so this formatting will come out funny.

Also, wrap the ensureIndexed method correctly.

mikerhodes:
Looks good when the doc comment formatting is cleaned up.

Classes/common/query/CDTDatastore+Query.h
alfinkel: Fixed in c290622

mikerhodes:
:+1:

Does the sample app use Query yet?

alfinkel:
I believe that it does.  I remember ripping out the old indexing code and replacing it with the new query code when we retired the previous indexing code.  It just does not use text search.

alfinkel:
Just to clarify:  The sample app used the old indexing code so it was a necessary to replace it with the new query code when we retired indexing.

PR #146, opened 2015-05-15
[BLOCKED] 46915 TDBlobStore must not generate filenames

indisoluble:
*What:*
Take out of `TDBlobStore` the code to generate a filename.

*Why:*
`TDBlobStore` receives a `NSData` instance and saves it to disk returning a filename. This filename is a SHA1 value generated based on the provided data. However, once we start to encrypt the attachments, this SHA1 value has to be based on the encrypted data. In PR #140 we took the code to read/write a file out of `TDBlobStore` to new classes, now we also have to move the code to generate the filenames to these classes given that they will be the only one with access to the ciphered data.

*How:*
We have simplified the protocol `CDTBlobWriter` and defined a new one called `CDTBlobMultipartWriter`. Both have a property called `sha1Digest` that contains the SHA1 value for the data saved to disk. Then we have implemented a few new classes that conforms to these protocols, essentially we have divided the class `CDTBlobData` into 3 classes and added a few more lines to generate the SHA1 digest.

**Check BugzID 46915 for more details about the design and how it has evolved until the current point.**

*Tests:*
As mentioned before, we have created 3 new classes that implement the protocols to access the content of an attachment: `CDTBlobDataReader`, `CDTBlobDataWriter` & `CDTBlobDataMultipartWriter`. We have added tests to ensure that these classes mirror the previous behaviour of `TDBlobStore`. We do not need to add test for `TDBlobStore` because the replication acceptance tests already test it.

reviewer @mikerhodes 
reviewer @rhyshort 

**NOTE:** Wait to review this PR until PR #144  is closed. This branch will be rebased and there will be only commits related to this case.

PR #148, opened 2015-05-19
[BLOCKED] 46915 Read & write encrypted attachments

indisoluble:
*What:*
`TDBlobStore` has to be able to handle encrypted attachments.

*Why:*
Because we have to cipher all the content in a datastore, i.e. databases and **attachments**.

*How:*
Implementing 3 new classes that conform to protocols: `CDTBlobReader`, `CDTBlobWriter` & `CDTBlobMultipartWriter` and are able to read/write encrypted attachments. These classes are: `CDTBlobEncryptedDataReader`, `CDTBlobEncryptedDataWriter` & `CDTBlobEncryptedDataMultipartWriter`. These classes are initialised with a `CDTEncryptionKey` and they use the key and AES to encrypt or decrypt the attachments. AES also requires a Initialisation Vector (IV), in all cases the IV is a buffer initialised to 0's, we will modify/improve this in a later PR.

Also there is another class named `CDTBlobHandleFactory`. `TDBlobStore` will initialise one of this factories with a key provider and it will use it to create instances that conform to the protocols mentioned before whenever it needs to read or write an attachment. The factory, not `TDBlobStore`, is responsible for creating an instance able to handle an encrypted attachment or plain data depending of the value/key returned by the key provider.

**Check BugzID 46915 for more details about the design and how it has evolved until the current point.**

*Tests:*
Tests consist on checking that the mentioned classes generate the expected values. Using the methods in `CDTEncryptionKeychainUtils+AES`:
* We provide some encrypted data to `CDTBlobEncryptedDataReader` and we check that the returned data is as expected.
* We provide some plain data to `CDTBlobEncryptedDataWriter` and `CDTBlobEncryptedDataMultipartWriter` and we validate that the generated file has the expected encrypted data. Also, we check that the SHA1 digest is generated based on the encrypted data.

We also tested that `TDBlobStore` is able to read an encrypted attachment that was saved before using the same key.

reviewer @mikerhodes 
reviewer @rhyshort 

**NOTE:** Wait to review this PR until PR #146  is closed. This branch will be rebased and there will be only commits related to this case.


indisoluble:
**To the reviewers:**
Even before starting with the reviewing I need to ask you a couple of question that I've not been able to clarify ... not with 100% certainty. 

I commented the [following property](https://github.com/cloudant/CDTDatastore/blob/46915-reader-writer-for-encrypted-attachments-continuation/Classes/common/touchdb/TDBlobStore.h#L103) in `TDBlobStoreWriter`:

```
/**
 The number of bytes added to the blob.
 
 If the attachment is encrypted, the size of the attachment on disk might be bigger than the
 value in this property.
 */
@property (readonly) UInt64 length;
```

As I understand it, this is OK given that the nobody is going to deal with the content of an encrypted attachment outside the `TDBlobStore`. And so far, I could only find one place where this value is checked: [here](https://github.com/cloudant/CDTDatastore/blob/46915-reader-writer-for-encrypted-attachments-continuation/Classes/common/touchdb/TDMultipartDocumentReader.m#L323):

```
            if (!lengthObj ||
                writer.length != [$castIf(NSNumber, lengthObj) unsignedLongLongValue]) {
                CDTLogWarn(CDTTD_REMOTE_REQUEST_CONTEXT,
                           @"%@: Attachment '%@' has invalid length %@ (should be %llu)", self,
                           attachmentName, lengthObj, writer.length);
                return NO;
            }
```

However, a second opinion would be much appreciated.

But my main concern is with this [other property](https://github.com/cloudant/CDTDatastore/blob/46915-reader-writer-for-encrypted-attachments-continuation/Classes/common/touchdb/TDBlobStore.h#L121):

```
/**
 After finishing, this is the SHA1 digest of the blob, in base64 with a "sha1-" prefix
 (the SHA1 digest is the value in property 'blobKey').
 */
@property (readonly) NSString* SHA1DigestString;
```

The SHA1 digest or blob key is the filename of the attachment in the local file system. As far as I know, this filename is only useful for the client library and the server does nothing with it. Having said this, I do not fully understand the purpose of [this line of code](https://github.com/cloudant/CDTDatastore/blob/46915-reader-writer-for-encrypted-attachments-continuation/Classes/common/touchdb/TDMultipartDocumentReader.m#L292):

```
                if (digest && !$equal(digest, actualDigest) &&
                    !$equal(digest, writer.SHA1DigestString)) {
                    CDTLogWarn(CDTTD_REMOTE_REQUEST_CONTEXT,
                               @"%@: Attachment '%@' has incorrect MD5 digest (%@; should be %@)",
                               self, attachmentName, digest, actualDigest);
                    return NO;
                }
```

Also the SHA1 digest is part of the metadata of the attachment, it is saved to disk with prefix "sha1-" and key `digest` ... Also this same key is used [here](https://github.com/cloudant/CDTDatastore/blob/46915-reader-writer-for-encrypted-attachments-continuation/Classes/common/touchdb/TDMultipartDocumentReader.m#L286) to get the MD5 from the attachment download from server ... I think. I need some help to understand this code.

tomblench:
@indisoluble I think you should be safe in treating the SHA-1 digest as a private property in order to derive the filename of the attachment. Only the MD5 is ever stored on couchdb and (optionally, I think) used to detect upload/download corruptions.

I think the TouchDB authors were mistaken in expecting the attachment's "digest" value to be either MD5 or SHA-1. See http://docs.couchdb.org/en/latest/api/document/common.html#attachments - "digest (string): Content hash digest. It starts with prefix which announce hash type (md5-) and continues with Base64-encoded hash digest". Also note that this has been carried over into couchbase-lite-ios - perhaps these digests are supported on couchbase? (https://github.com/couchbase/couchbase-lite-ios/blob/master/Source/CBLMultipartDocumentReader.m#L364).

As regards attachment length, you must send the correct (unencrypted) attachment length (both the gzipped and uncompressed values, if appropriate) to the server, as the unencrypted attachment is stored there (as I understand it).

indisoluble:
@tomblench Thank you Tom.

After discussing about this, I'm going to create a new PR to run the replication test with datastore encryption. It is the only to ensure that we are sending/receiving the data as before.

mikerhodes:
There are a few things to separate out here:

1. The `digest` field that Cloudant/Couch sends us in the `_attachments` field when we are replicating an attachment. We can use that as an integrity check. From our point of view, they can send md5, sha1, sha256 or whatever -- just some of those digests we'll be unable to use to verify the integrity of the downloaded attachment.
2. The `digest` field that CDTDatastore sends _to_ Couch in the `_attachments` field. We need to make sure that's a digest of the content we're sending.
3. The sha1 of file content which is used to generate the on-disk filename. 

The sad fact is that (3) is logically separate from (2) above, but the code currently conflates the two. See [here][1] where the code creating the attachments object for assumes the `key` database column is a valid sha1 digest of the plaintext blob content which can be used in the `_attachment` arrays `digest` field. The `key` field is also used all over the place as the filename.

The question here is whether it's worth our while implementing a new filename mechanism to prevent leaking information. And whether we actually do leak information because the fact that the attachment data is shared between multiple attachments to documents is encrypted anyway in the JSON data...

[1]: https://github.com/cloudant/CDTDatastore/blob/master/Classes/common/touchdb/TD_Database%2BAttachments.m#L347

indisoluble:
So the SHA1 digest is not only useful for the client library, it is sent to the server and we also use the value returned by the server to access the blobs.

This invalidates my previous assumption and also invalidates the changes done so far. I have to cancel this pull request as well as the previous one: #146, given that it was done only to facilitate the changes in this PR.

PR #150, opened 2015-05-21
Relax FTS availability check to support both standard and enhanced query syntax

alfinkel:
_What_

This PR changes the check for full text search availability in SQLite to include support for either query syntax.

_Why_

Currently, if FTS is available with standard query syntax the FTS availability check will not allow a text index to be created.  This restriction needs to be relaxed so that text search can be used regardless of which query syntax is currently supported by SQLite.

_How_

Remove the check for the `ENABLE_FTS3_PARENTHESIS` compile option and leave only the `ENABLE_FTS3` compile option check.

reviewer @mikerhodes 
reviewer @rhyshort 

BugId: 47321

### DIFF
--- a/Classes/common/query/CDTQIndexManager.m
+++ b/Classes/common/query/CDTQIndexManager.m
@@ -379,8 +379,7 @@ static const int VERSION = 2;
     __block BOOL ftsOptionsExist = NO;
     
     [db inDatabase:^(FMDatabase *db) {
-        NSMutableArray *ftsCompileOptions =
-            [NSMutableArray arrayWithArray:@[ @"ENABLE_FTS3", @"ENABLE_FTS3_PARENTHESIS" ] ];
+        NSMutableArray *ftsCompileOptions = [NSMutableArray arrayWithArray:@[ @"ENABLE_FTS3" ] ];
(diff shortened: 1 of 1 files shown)


mikerhodes:
I guess this basically brings us into line with Android -- that we can "create a text index", though it's the user's responsibility to use the basic capabilities -- this should be documented along with this PR?

alfinkel:
It actually is already documented in that we state that we support which ever syntax is enabled in SQLite.  This fix will actually bring the iOS version in line with the docs and with Android as you said.

Also, I think it is more accurate to say that it's the user's responsibility to use which ever syntax is enabled on their SQLite.  We support either one and only care whether FTS is enabled.  The docs state this.  Also the docs state that we've tested iOS 7.1 - iOS 8.3 and in each case enhanced query syntax was enabled.  So the developer should be prepared for enhanced query syntax out of the box if they read the docs.  We have a similar statement in the Android docs except that it is standard query syntax that is typically enabled.

rhyshort:
Sounds reasonable to me, +1

mikerhodes:
:+1: though at some point it'd be nice to add an isEnhancedTextQuerySyntax available for both Android and iOS.

alfinkel:
Agreed.  I added a FB case 47338 for this.

PR #152, opened 2015-05-21
Add $mod operator support to Query

alfinkel:
_What_

This PR adds support for the `$mod` operator for use in Query.

_Why_

In an effort to provide as much functionality to Query as is available elsewhere (such as MongoDB), we are adding support to the `$mod` operator so that users can write queries containing modulo arithmetic.

_How_

Add logic to the validator code, the SQLtranslator code, and the unindexed matcher code to support the safe use of `$mod` in queries.

_Tests_

- A series of validation tests have been added to CDTQQuerySqlTranslatorTests to exercise argument validation and query normalization for queries containing the `$mod` operator.
- A series of SQL WHERE clause generation tests have been added to CDTQQuerySqlTranslatorTests for queries containing the `$mod` operator.
- A series of execution tests have been added to CDTQQueryExecutorTests to exercise execution by the SQL engine and the unindexed matcher engine for queries containing the `$mod` operator.

reviewer @mikerhodes 
reviewer @gadamc 

BugId: 44632

### DIFF
--- a/doc/query.md
+++ b/doc/query.md
@@ -203,7 +203,7 @@ sign. This is due to restrictions imposed by SQLite's virtual table syntax.
 
 Use `-listIndexes` to retrieve a dictionary containing all of the query indexes in a datastore.  The key to the dictionary is the index name.
 
-The format of the dictionary returned by `-listinidexes` is:
+The format of the dictionary returned by `-listindexes` is:
 
 ```objc
 @{ @"jsonIdxName": @{ @"fields": @[ @"field1", @"field2" ],
@@ -263,6 +263,23 @@ To query for documents where `age` is greater than twelve use the `$gt` conditio
 ```
 
 See below for supported operators (Selections -> Conditions).
+
+#### Modulo operation in queries
+
+Using the `$mod` operator in queries allows you to select documents based on the value of a field divided by an integer yielding a specific remainder.
+
+To query for documents where `age` divided by 5 has a remainder of  4, do the following:
+
+```objc
+@{ @"age": @{ @"$mod": [ @5, @4 ] } }
+```
+
+A few things to keep in mind when using `$mod` are:
+
+- The array argument to the `$mod` operator must contain two number elements. The first element is the divisor and the second element is the remainder.
+- Division by zero is not allowed so the divisor cannot be zero.
+- The dividend (field value), divisor, and the remainder can be positive or negative.
+- The dividend, divisor, and the remainder can be represented as whole numbers or by using decimal notation.  However internally, prior to performing the modulo arithmetic operation, all three are truncated to their logical whole number representations.  So, for example, the query `@{ @"age": @{ @"$mod": [ @5.6, @4.2 ] } }` will provide the same result as the query `@{ @"age": @{ @"$mod": [ @5, @4 ] } }`.
 
 #### Text search
 
@@ -599,6 +616,7 @@ Selectors -> Condition -> Objects
 Selectors -> Condition -> Misc
 
 - `$text` in combination with `$search`
+- `$mod`
 
 Selectors -> Condition -> Array
 
@@ -649,7 +667,6 @@ Selectors -> Condition -> Array
 
 Selectors -> Condition -> Misc
 
-- `$mod`
 - `$regex`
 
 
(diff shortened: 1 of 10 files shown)


mikerhodes:
So the SQL for `$mod` is something like `fieldName % 3 = 2` so I can check your SQL generation?

mikerhodes:
A nice easy to review commit sequence. Thanks :)

Just a couple of minor queries/suggestions.

mikerhodes:
As noted on query.md, we need to figure out the precision issues here. Using fmod/double can at least give us something more predicatable.

doc/query.md
```diff
@@ -264,6 +264,23 @@ To query for documents where `age` is greater than twelve use the `$gt` conditio
 
 See below for supported operators (Selections -> Conditions).
 
+#### Modulo operation in queries
+
+Using the `$mod` operator in queries allows you to select documents based on the value of a field divided by a number yielding a specific remainder.
```
gadamc: s /a number/an integer

doc/query.md
```diff
@@ -264,6 +264,23 @@ To query for documents where `age` is greater than twelve use the `$gt` conditio
...
+@{ @"age": @{ @"$mod": [ @5, @4 ] } }
+```
+
+A few things to keep in mind when using `$mod` are:
+
+- The array argument to the `$mod` operator must contain two number elements. The first element is the divisor and the second element is the remainder.
+- Division by zero is not allowed so the divisor cannot be zero.
+- The dividend (field value), divisor, and the remainder can be positive or negative.
+- The dividend, divisor, and the remainder can be represented as whole numbers or by using decimal notation.  However internally, prior to performing the modulo arithmetic operation, the dividend and divisor are rounded down to the nearest integer for positive values and rounded up to the nearest integer for negative values while the remainder is untouched.  So the query `@{ @"age": @{ @"$mod": [ @5.6, @4 ] } }` will provide the same result as the query `@{ @"age": @{ @"$mod": [ @5, @4 ] } }`.
+
```
gadamc: Perhaps put the "However, internally, prior to performing the modulo ....  while the remainder is untouched" in **bold** ? Or some other way to highlight this.

alfinkel: No longer necessary.  This has been updated and the remainder is now also truncated.

gadamc:
So, I've been looking at this for quite some time. I guess I'd like to have a well defined specification for this. What is the exact behavior that we want?  My confusion lies around what kind of modulation operation we're performing and the expected result under particular conditions.  From @alfinkel's latest changes, it seems that we are only doing integer operations. But it wasn't clear to me that this was the plan since @mikerhodes mentioned using `fmod`. (Although, a recent private discussion with @alfinkel has clarified some things... so I think it's integer only. I leave this statement here just to make sure it's clear.)

Also, do we truncate everything or do we floor everything? That is, does -1.2  -> -2 or -1? @alfinkel says "truncated". @mikerhodes mentioned "floor" in a previous comment. So, I just wanted to make sure that was clarified. 

I assume that we're trying to match our own CloudantQuery "spec" (it's not clear to me if that exactly matches Mongo's spec): https://docs.cloudant.com/api.html#condition-operators

There are two things with this "spec". 
1. It says that if a document's field is not an integer, it automatically returns false.
2. It says that it returns true when field % divisor == remainder.   The sign of the remainder is platform/language dependent, from what I understand. But, if we restrict everything to be positive, then there's no ambiguity. 

Should we change the behavior such that a query automatically fails if the caller uses negative or floating point values in the [divisor,remainder] rather than truncating those values?  What does CloudantQuery do? Does it fail or does it truncate the values? We should match it's behavior, no?

Finally, how do we handle large floating point numbers (i.e. > NSIntegerMax)? 6e23 is an "integer", but it will convert to -9223372036854775808 on [NSNumber -integerValue] .

alfinkel:
I'd be perfectly fine with restricting the field value (dividend), divisor and the remainder to positive integers.  It might simplify some things, although restricting the field value could add some complexity.  Up until now we've used Mongo's query as our model.  Using Mongo as a reference point has been one of the more consistent aspects during the development of Query - mobile and has served us well up until now.  So I am hesitant to deviate from that path.  However, @gadamc makes some good points.  Let me know what you think @mikerhodes.

gadamc:
@alfinkel @mikerhodes (I slightly modified my comment above... I revised my point #2 about the CloudantQuery spec)

gadamc:
@alfinkel @mikerhodes - I removed a huge chunk of my comment. After further reading and testing, I wrote up a little document that I've sent to you by email.

alfinkel:
@gadamc, A spec summarizing the `$mod` approach can be found here: https://github.com/cloudant/cloudant-sync/blob/master/specs/query_mod_operator.md.  Hopefully it removes any confusion.  Going forward I will be sure to add a spec for each additional operator that we intend to implement.

mikerhodes:
I'm +1 on this one now we've nailed down the behaviour Mongo seems to be using.

gadamc:
+1

PR #156, opened 2015-05-26
Document the use of dollar sign and period in a field name

alfinkel:
_What_

This PR adds content to the documentation to explain the use of `$` and `.` in field names.

_Why_

A user needs to understand the expected behavior of the query and indexing engine when naming fields.

_How_

Add a "Restrictions" sub-heading under the "Creating Indexes" heading to describe the behaviors of using a dollar sign as the first character in a field name and a period within a field name.

reviewer @mikerhodes 
reviewer @gadamc 

BugId: 47451

### DIFF
--- a/doc/query.md
+++ b/doc/query.md
@@ -14,8 +14,15 @@ This query engine uses indexes explicitly defined over the fields in the documen
 indexes can be created for use in different queries, the same field may end up indexed in
 more than one index.
 
-Querying is carried out by supplying a query in the form of a dictionary which describes the
-query.
+Query offers a powerful way to find documents within your datastore. There are a couple of restrictions on field names you need to be aware of before using query:
+
+- A dollar sign (`$`) cannot be the first character of any field name.  This is because, when querying, a dollar sign tells the query engine to handle the object as a query operator and not a field.
+- A field with a name that contains a period (`.`) cannot be indexed nor successfully queried.  This is because the query engine assumes dot notation refers to a sub-object.
+
+These come from Query's MongoDB heritage where these characters are not allowed in field names, which we don't share. Hopefully we'll work around these restrictions in the future.
+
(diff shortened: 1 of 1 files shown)


mikerhodes:
:+1:

doc/query.md
```diff
@@ -14,8 +14,15 @@ This query engine uses indexes explicitly defined over the fields in the documen
 indexes can be created for use in different queries, the same field may end up indexed in
 more than one index.
 
-Querying is carried out by supplying a query in the form of a dictionary which describes the
-query.
+Query offers a powerful way to find documents within your datastore. There are a couple of restrictions on field names you need to be aware of before using query:
+
+- A dollar sign (`$`) cannot be the first character of any field name.  This is because, when querying, a dollar sign tells the query engine to handle the object as a query operator and not a field.
+- A field with a name that contains a period (`.`) cannot be indexed.  This is because using dot notation tells the query engine to index a sub-object when creating an index.
```
gadamc: Does this mean, however, that a dot-notated field can be searched for in the post-hoc matcher? That is, given a set of docs like
```
{
  name : mike
  age : 34
  pet.species : cat
  pet.name : bob
}
```
and an index a field "name", can I make the query `{selector : {name : mike, pet.species = cat}}`?

alfinkel: No.  The phrase `query engine` means the engine that handles queries for Cloudant Query - mobile which is comprised of both the matcher and the SQL translator.  The notion of the matcher is not exposed in this documentation as it is internal to the implementation.

gadamc: Okay... so, `{selector : {name : mike, pet.species = cat}}` would be a valid query and would return that example document?

alfinkel: Updated based on comments.  Done in a95adf8

gadamc:
+1

PR #153, opened 2015-05-22
47318 Encrypt attachments

indisoluble:
*What:*
`TDBlobStore` has to be able to handle encrypted attachments.

*Why:*
We have to cipher the entire datastore, i.e. databases and **attachments**.

*How:*
1. Implement a new class named `CDTBlobEncryptedData` that conforms to protocols `CDTBlobReader` & `CDTBlobWriter`. This class relies on `CDTBlobData` and `CDTEncryptionKeychainUtils` to generate encrypted attachments (for more details about the structure of theses attachments check [encryption-at-rest.md](https://github.com/cloudant/cloudant-sync/blob/master/specs/encryption-at-rest.md)).
2. Factory named `CDTBlobHandleFactory` to create instances that conform to `CDTBlobReader` & `CDTBlobWriter`.
3. Modify `TDBlobStore` to use `CDTBlobHandleFactory`. So the factory is responsible for creating an instance able to handle an encrypted attachment or plain data depending of the value/key returned by the key provider.

*Tests:*
* `CDTBlobEncryptedData`.- Passes the same tests that `CDTBlobData` and:
    1. Generate attachments with the correct structure.

* `CDTBlobHandleFactory`
    1. Create instances of `CDTBlobData` is the key provider returns nil.
    2. Create instances of `CDTBlobEncryptedData` is the key provider returns an instance of `CDTEncryptionKey`.

* `TDBlobStore`
    1. Generate the same SHA1 digest if the attachments are encrypted.
    2. It is able to generate encrypted files.
    3. It is still able to return plain data even if the attachments are encrypted before saving to disk.

reviewer @mikerhodes 
reviewer @rhyshort 

**NOTE:** Wait to review this PR until PR #144  is closed. This branch will be rebased and there will be only commits related to this case.

PR #157, opened 2015-05-26
47447 Implement a simple key provider

indisoluble:
*What:*
Implement a new class called `CDTEncryptionKeySimpleProvider` that conforms to `CDTEncryptionKeyProvider` and simply holds a developer-provided key.

*Why:*
It's useful to provide `CDTEncryptionKeySimpleProvider` as an example of the simplest thing which could work.

*How:*
This class is initialised with a `NSData` instance of 32 bytes length. When the method `CDTEncryptionKeyProvider:encryptionKey` is called, it return the `NSData` instance wrapped in a `CDTEncryptionKey`.

*Tests:*
1. Initialisation fails if no data is provided.
2. Initialisation fails if the data does not have the right size.
3. It returns the expected key.

reviewer @mikerhodes 
reviewer @rhyshort 

**NOTE:** Wait to review this PR until PR #153  is closed. This branch will be rebased and there will be only commits related to this case.

### DIFF
--- a/Classes/common/Encryption/CDTEncryptionKeySimpleProvider.m
+++ b/Classes/common/Encryption/CDTEncryptionKeySimpleProvider.m
@@ -1,0 +1,53 @@
+//
+//  CDTEncryptionKeySimpleProvider.m
+//  CloudantSync
+//
+//  Created by Enrique de la Torre Fernandez on 26/05/2015.
+//  Copyright (c) 2015 IBM Cloudant. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+//  except in compliance with the License. You may obtain a copy of the License at
+//  http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software distributed under the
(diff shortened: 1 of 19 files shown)


mikerhodes:
It looks good.

However, the changes which change CDTHelperFixedKeyProvider from alloc,init to provider should be a separate commit.

Classes/common/Encryption/CDTEncryptionKeySimpleProvider.m
```diff
@@ -0,0 +1,53 @@
...
+
+@interface CDTEncryptionKeySimpleProvider ()
+
+@property (strong, nonatomic, readonly) CDTEncryptionKey *thisKey;
+
+@end
+
+@implementation CDTEncryptionKeySimpleProvider
+
+- (instancetype)init { return [self initWithKey:nil]; }
```
rhyshort: Shouldn't we disable the initializer? Or is it needed for something?

rhyshort: Ignore me, just noticed the header file

rhyshort:
+1

mikerhodes:
Thanks for splitting, +1.

PR #159, opened 2015-06-01
Remove most warnings when building as a cocoapod

jimix:
I have been sitting on these for a while.  Reducing existing warnings make it easier to know if I've added any.
I have broken them up in a way that the individual commits can be cherry-picked if necessary.
If there is interest, I'd be happy to go after the remainder.

### DIFF
--- a/Classes/vendor/MYUtilities/MYErrorUtils.m
+++ b/Classes/vendor/MYUtilities/MYErrorUtils.m
@@ -139,22 +139,19 @@ NSString* MYErrorName( NSString *domain, NSInteger code ) {
 #if !TARGET_OS_IPHONE || defined(__SEC_TYPES__)
     else if ($equal(domain,NSOSStatusErrorDomain)) {
         // If it's an OSStatus, check whether CarbonCore knows its name:
-        const char *name = NULL;
-#if !TARGET_OS_IPHONE
-        name = GetMacOSStatusErrorString((int)code);
-#endif
-        if (name && *name)
-            result = [NSString stringWithCString: name encoding: NSMacOSRomanStringEncoding];
-        else {
+        NSError *osErr = [NSError errorWithDomain:NSOSStatusErrorDomain code:code userInfo:nil];
+        result = [osErr localizedDescription];
+
 #if MYERRORUTILS_USE_SECURITY_API
+        if (!result) {
             result = (id) SecCopyErrorMessageString((OSStatus)code,NULL);
             if (result) {
                 [NSMakeCollectable(result) autorelease];
                 if ([result hasPrefix: @"OSStatus "])
                     result = nil; // just a generic message
             }
-#endif
         }
+#endif
     }
 #endif
     
--- a/Classes/common/Attachments/CDTAttachment.m
+++ b/Classes/common/Attachments/CDTAttachment.m
@@ -188,8 +188,6 @@
 {
     BOOL stub = [jsonData[@"stub"] boolValue];
     NSNumber *length = jsonData[@"length"];
-    NSString *digest = jsonData[@"digest"];
-    NSNumber *revpos = jsonData[@"revpos"];
     NSString *contentType = jsonData[@"content_type"];
     NSString *data = jsonData[@"data"];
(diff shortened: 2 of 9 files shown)


rhyshort:
@jimix I'd like to understand why the property was causing failing on OS X 10.9, so what errors etc. were you getting?

jimix:
re: cancelled
the 10.9 SDK does not define the `cancelled` property so the build breaks.
The setter `isCancelled` is there and continues to be available in 10.10

rhyshort:
Okay, I'm happy for these to get merged, but before we do could you change the commit message for d702312 to something that explains what was wrong with the build? Something like:
```
Replace `cancelled` property with `isCancelled` method

`cancelled` property is only available on targets iOS 8+ and OS X 10.10+
replacing with  its getter `isCancelled`

```
or  similar so we can tell from the commit message why the build was failing.

jimix:
@rhyshort done.. my comment was ***horrible***, sorry about that and thanks for suggesting something better.

Classes/vendor/MYUtilities/MYErrorUtils.m
```diff
@@ -139,22 +139,19 @@ BOOL MYMiscError( NSError **error, NSString *message, ... )
...
         // If it's an OSStatus, check whether CarbonCore knows its name:
-        const char *name = NULL;
-#if !TARGET_OS_IPHONE
-        name = GetMacOSStatusErrorString((int)code);
-#endif
-        if (name && *name)
-            result = [NSString stringWithCString: name encoding: NSMacOSRomanStringEncoding];
-        else {
+        NSError *osErr = [NSError errorWithDomain:NSOSStatusErrorDomain code:code userInfo:nil];
+        result = [osErr localizedDescription];
```
indisoluble: If we return the localised description, what would happen if we run the test with another language (not english)? I mean, will [these tests](https://github.com/cloudant/CDTDatastore/blob/master/Classes/vendor/MYUtilities/MYErrorUtils.m#L212) pass?

jimix: @indisoluble, excellent question. It appears that `GetMacOSStatusErrorString()` also returns a localized string, but I cannot tell you exactly what will happen.

indisoluble: Well, in that case, I think is OK

indisoluble:
:+1:

Classes/vendor/MYUtilities/MYErrorUtils.m
jimix: Actually.. playing around with this a bit, here is what I've learned, build the following CLI application:
```objc
#import <Foundation/Foundation.h>

int main(int argc, const char * argv[]) {
    @autoreleasepool {
        // insert code here...
        NSLog(@"Hello, World!");

        int code = fnfErr;

        const char *name = GetMacOSStatusErrorString(code);
        NSString *orig = [NSString stringWithCString: name encoding: NSMacOSRomanStringEncoding];
        NSLog(@"orig: %@", orig);

        NSError *osErr = [NSError errorWithDomain:NSOSStatusErrorDomain code:code userInfo:nil];
        NSString *loc = [osErr localizedDescription];
        NSLog(@"loc: %@", loc);

        NSString *desc = [osErr description];
        NSLog(@"desc: %@\n", desc);

    }
    return 0;
}
```
The result is:
```shell
2015-06-03 09:28:02.320 oserr[14220:3115096] Hello, World!
2015-06-03 09:28:02.321 oserr[14220:3115096] orig: fnfErr
2015-06-03 09:28:02.324 oserr[14220:3115096] loc: The operation couldn’t be completed. (OSStatus error -43.)
2015-06-03 09:28:02.325 oserr[14220:3115096] desc: Error Domain=NSOSStatusErrorDomain Code=-43 "The operation couldn’t be completed. (OSStatus error -43.)" (fnfErr: File not found)
```

So the result is significantly different.

If this makes anyone nervous, we can resort to squelching the warning with a `#pragma`.

PR #161, opened 2015-06-08
Correctly generate attachment URLs for non-winning rev

rhyshort:
Correctly generate attachment URLs for non-winning rev
by manually creating a  URL using URL components.

BugzId: 47846

reviewer @indisoluble 
reviewer @tomblench

### DIFF
--- a/Classes/common/CDTDocumentRevision.m
+++ b/Classes/common/CDTDocumentRevision.m
@@ -79,11 +79,17 @@
 
     // build the attachment objects
     for (NSString *key in [attachmentData allKeys]) {
-        NSURL *url = [NSURL URLWithString:key relativeToURL:documentURL];
+        
+        NSURLComponents *documentUrlComponents = [NSURLComponents componentsWithString:[documentURL absoluteString]];
+        NSString *documentPathString = documentUrlComponents.path;
+        NSString *attachmentPath = [NSString stringWithFormat:@"%@/%@", documentPathString,key];
+        documentUrlComponents.path = attachmentPath;
+        
+        
         CDTSavedHTTPAttachment *attachment =
             [CDTSavedHTTPAttachment createAttachmentWithName:key
                                                     JSONData:[attachmentData objectForKey:key]
-                                               attachmentURL:url
+                                               attachmentURL:[documentUrlComponents URL]
                                                        error:error];
         if (*error) {
             return nil;
--- a/ReplicationAcceptance/ReplicationAcceptance/Attachments.m
+++ b/ReplicationAcceptance/ReplicationAcceptance/Attachments.m
@@ -100,6 +100,118 @@
 }
 
 #pragma mark Tests
+-(void)testSavedHttpAttachmentWithRemote
+{
+    //
+    // Set up remote database
+    //
+    
+    NSString *remoteDbName = [NSString stringWithFormat:@"%@-test-database-%@",
+                              self.remoteDbPrefix,
+                              [CloudantReplicationBase generateRandomString:5]];
+    NSURL *remoteDbURL = [self.remoteRootURL URLByAppendingPathComponent:remoteDbName];
+    
+    [self createRemoteDatabase:remoteDbName
+                   instanceURL:self.remoteRootURL];
+    
+    //
+    // Add attachments to a document in the local store
+    //
+    
+    // Contains {attachmentName: attachmentContent} for later checking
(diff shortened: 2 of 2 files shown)


tomblench:
Tests? Also what's the behaviour with basic auth?

Classes/common/CDTDocumentRevision.m
```diff
@@ -79,11 +79,17 @@ + (CDTDocumentRevision *)createRevisionFromJson:(NSDictionary *)jsonDict
 
     // build the attachment objects
     for (NSString *key in [attachmentData allKeys]) {
-        NSURL *url = [NSURL URLWithString:key relativeToURL:documentURL];
+        
+        NSURLComponents *documentUrlComponents = [NSURLComponents componentsWithURL:documentURL resolvingAgainstBaseURL:NO];
+        NSString *documentPathString = documentUrlComponents.path;
+        NSString *attachmentPath = [NSString stringWithFormat:@"%@/%@", documentPathString,key];
+        documentUrlComponents.path = attachmentPath;
```
indisoluble: Based on the description in BugzId 47846, I think the problem was that the URL did not include the `<REVID>`.
Where are you adding the `<REVID>`? I'm not able to see it.

indisoluble:
I agree with @tomblench, at least one test for the new code would be appreciated.

Classes/common/CDTDocumentRevision.m
rhyshort: the query is already in the `NSURLComponents` object, we are just replacing the path

rhyshort:
I've yet to figure out how to test it since the URL is an internal property and the suite the tests for this are part of are don't use a couch server

Classes/common/CDTDocumentRevision.m
indisoluble: Thanks.

tomblench:
@rhyshort it should be easy enough to write a test in acceptance tests which runs against local couch, or am I missing something? Although you can't directly check the URL you can test that createRevisionFromJson picks up the right rev and therefore you know it generated the right URL.

rhyshort:
@tomblench I can't just check the URL generated since its private on the CDTSavedHTTPAttachment class unless I do some digging with the objective c runtime

ReplicationAcceptance/ReplicationAcceptance/Attachments.m
```diff
@@ -100,6 +100,118 @@ - (void)deleteAttachmentNamed:(NSString*)attachmentName
...
+    NSMutableDictionary *attachments = [NSMutableDictionary dictionary];
+        NSString *content = @"blahblah";
+        NSData *data = [content dataUsingEncoding:NSUTF8StringEncoding];
+        NSString *name = @"attachment";
+        CDTAttachment *attachment = [[CDTUnsavedDataAttachment alloc] initWithData:data
+                                                                              name:name
+                                                                              type:@"text/plain"];
+        [attachments setObject:attachment forKey:name];
+        
+        [originalAttachments setObject:data forKey:name];
```
indisoluble: These lines are not correctly tabulated.

rhyshort: done in 1748ba8

indisoluble:
:+1:

ReplicationAcceptance/ReplicationAcceptance/Attachments.m
```diff
@@ -100,6 +100,118 @@ - (void)deleteAttachmentNamed:(NSString*)attachmentName
...
+    XCTAssertTrue([self compareDatastore:self.datastore withDatabase:remoteDbURL],
+                  @"Local and remote database comparison failed");
+    
+    XCTAssertTrue([self compareAttachmentsForCurrentRevisions:self.datastore
+                                                 withDatabase:remoteDbURL],
+                  @"Local and remote database attachment comparison failed");
+    
+    NSDictionary *responseDict = response.body.JSONObject;
+    
+    CDTDocumentRevision *remoteRevision = [CDTDocumentRevision createRevisionFromJson:responseDict forDocument:docURL error:&error];
```
tomblench: Assert on remoteRevision revid?

tomblench:
+1

ReplicationAcceptance/ReplicationAcceptance/Attachments.m
rhyshort: I wouldn't say its needed since the attachments are deleted on the current winning rev.

PR #162, opened 2015-06-09
47559 Fixing sha1 naming

indisoluble:
*What:*
Save attachments to disk with a random-generated filename.

*Why:*
To save an attachment to disk, we were using the SHA1 digest of the plain data to generate its filename. This is not a problem as long as the datastore is not encrypted but if it is and the attachment has a well-known content, an attacker can simply generate the SHA1 values for all well-known content and discover the content of any attachment without decrypting the file.

*How:*
1. Add a new table to the database to relate keys (SHA1 digest of the plain data) with filenames.
2. Populate the new table with the current content in the database:
    1. Get the keys in the attachment table
    2. For each key, generate a filename equal to: <key>.blob
    3. Insert keys and filenames in the new table
3. Modify methods in `TDBlobStore` to receive a `FMDatabase` instance. These methods need access to the database to get the filename for a key or insert a key and a filename when an attachment is saved to disk.
4. Instead of generating the filename based on the plain content of the attachment, use a random function to generate the filename.

*Tests:*
* Create table if the database version is less than or equal to 100
* The new table is populated with the current attachments
* The same key can not be inserted twice in the database
* However, if an user tries to save the same attachment twice (with the same key), the operation succeeds and returns the filename generated the first time. But no file is saved to disk and the db is not updated.
* Check that a migrated key (moved to the new table) can be retrieved as any other key.
* Check that we are still able to delete attachment from disk and also from db.

reviewer @mikerhodes 
reviewer @rhyshort


rhyshort:
Looks good to me now +1

mikerhodes:
+1

PR #163, opened 2015-06-09
Fix containsAllElements matcher function

alfinkel:
_What:_

Fix the `containsAllElements` matcher test function to test that the expected and the actual arrays contain the same set of elements.

_Why:_

The behavior of `containsAllElements` is supposed to mirror the behavior of the Hamcrest `containsInAnyOrder` method from the sync-android tests which checks to see that the expected and the actual arrays both contain the same set of elements.  Up until now, the `containsAllElements` function was only ensuring that the expected array elements were a subset of the actual array elements.

_How:_

Add an additional loop to validate the expected array contents against the full set of contents of the actual array.

reviewer @mikerhodes 
reviewer @rhyshort 

BugId: 47952


Tests/Tests/Matchers/CDTQContainsAllElementsMatcher.m
```diff
@@ -31,6 +31,16 @@
         break;
     }
     
+    if (matches) {
```
rhyshort: I think we can save this loop by doing this before the first pass of the arrays:

```
if([actual count] != [expected count]){
        return NO;
}
```

alfinkel: Done in 4152bb8

rhyshort:
+1

mikerhodes:
Could you also rename the method so it matches the hamcrest method's name if it has the same behaviour.

alfinkel:
Renamed the matcher method to `containsInAnyOrder:`.  Done in 30ef62d.  Replaced all references in tests.  Done in b12da7c.

mikerhodes:
+1

PR #166, opened 2015-06-16
Add $size operator support to Query

alfinkel:
_What:_

This PR adds support for the $size operator for use in Query.  See [$size spec](https://github.com/cloudant/cloudant-sync/blob/master/specs/query_size_operator.md) for details.

_Why:_

In an effort to provide as much functionality to Query as is available elsewhere (such as MongoDB), we are adding support to the $size operator.  It provides a way to query for documents based on the size of an array that is a document field value.

_How:_

Add logic to the validator code, the SQLtranslator code, and the unindexed matcher code to support the use of `$size` in queries.  `$size` operator clauses are handled strictly by the unindexed matcher code.  There is no SQL translator component for `$size`.

_Tests:_

- Validation tests have been added to ensure that queries containing the `$size` operator are properly normalized and validated.
- Tests have been added to ensure that the new utility method used by the SQL translator to check for the `$size` operator functions as desired.
- Unindexed matcher tests have been added to ensure that the matcher functionality handles `$size` operations appropriately.
- Since the unindexed matcher handles all `$size` processing, execution tests have been added to ensure that `$size` query execution performs correctly.

reviewer @mikerhodes 
reviewer @rhyshort 

BugId: 44634

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CDTDatastore CHANGELOG
 
+## Unreleased
+
+- [NEW] Added query support for the `$size` operator.
+
 ## 0.17.1 (2015-06-24)
 
 - [FIX] Allow using encrypted datastores and FTS together.
--- a/Classes/common/query/CDTQQueryConstants.h
+++ b/Classes/common/query/CDTQQueryConstants.h
@@ -41,3 +41,5 @@ extern NSString *const NIN;
 extern NSString *const SEARCH;
 
 extern NSString *const MOD;
+
+extern NSString *const SIZE;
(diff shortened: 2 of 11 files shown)


rhyshort:
+1

mikerhodes:
Could you expand on why we can't use SQL for $size queries?

alfinkel:
We don't use SQL for `$size` queries because we would need to alter the way in which we handle our internal indexing.  To start, we would need to track that a field in an index is an array field.  The most likely way to do this would be to add a column to every Query index that we create allowing us to store the name of an array field for tracking purposes.  Without this ability we cannot honor the `$size` behavior of mongoDB query.  It is theoretically possible to implement a SQL component for `$size` but probably not worth it in this case as the changes would impact other things throughout the Query code base.  We can revisit this in the future, if the need arises but for now I think we should handle the `$size` operator in the matcher logic only.  As far as mongoDB query goes, they also do not use indexing when dealing with the `$size` operator. 

From mongoDB query docs:

```
Queries cannot use indexes for the $size portion of a query, although the other portions of a query can use indexes if applicable.
```

ref:  http://docs.mongodb.org/manual/reference/operator/query/size/#op._S_size

mikerhodes:
+1

PR #167, opened 2015-06-17
48270 Activate FTS on SQLCipher

indisoluble:
*What:*
If you try to use full text indexing (FTS) on iOS with an encrypted datastore, it fails.

*Why:*
It turns out that SQLCipher's podspec doesn't build with FTS by default.

*How:*
Force cocoapods to compile SQLCipher with FTS. To do that, we simply have to replace this line in `CDTDatastore.podspec`:

```
sp.dependency 'SQLCipher'
```

With:

```
sp.dependency 'SQLCipher/fts'
```

*Tests:*
Add a couple of tests to EncryptionTests workspace:
* Text search is enabled for a CDTQIndexManager instance.
* Text search is enabled for a Datastore instance.

reviewer @alfinkel 
reviewer @rhyshort

### DIFF
--- a/EncryptionTests/Encryption Tests/CDTQIndexManagerEncryptionTests.m
+++ b/EncryptionTests/Encryption Tests/CDTQIndexManagerEncryptionTests.m
@@ -23,6 +23,7 @@
 #import "FMDatabase+SQLCipher.h"
 
 #import "CDTQIndexManager.h"
+#import "DBQueryUtils.h"
 
 @interface CDTQIndexManagerEncryptionTests : CloudantSyncTests
 
@@ -207,6 +208,23 @@
 
     XCTAssertNil(im, @"indexManager is nil");
     XCTAssertNotNil(err, @"There is an error");
+}
+
+- (void)testTextSearchIsAvailable
+{
+    CDTHelperFixedKeyProvider *provider = [CDTHelperFixedKeyProvider provider];
+    CDTDatastore *datastore = [self.factory datastoreNamed:@"is_fts_activated"
+                                 withEncryptionKeyProvider:provider
+                                                     error:nil];
+    
+    CDTQIndexManager *im = [[CDTQIndexManager alloc] initUsingDatastore:datastore error:nil];
+
+    NSSet *compileOptions = [DBQueryUtils compileOptions:im.database];
+    NSSet *ftsCompileOptions = [NSSet setWithArray:@[ @"ENABLE_FTS3" ]];
+
+    XCTAssertTrue([ftsCompileOptions isSubsetOfSet:compileOptions]);
+
+    XCTAssertTrue([im isTextSearchEnabled], @"It should be activated");
 }
 
 @end
--- a/EncryptionTests/Encryption Tests/DatastoreEncryptionTests.m
+++ b/EncryptionTests/Encryption Tests/DatastoreEncryptionTests.m
@@ -23,6 +23,7 @@
 
 #import "CDTDatastoreManager+EncryptionKey.h"
 #import "CDTDatastore+EncryptionKey.h"
+#import "CDTDatastore+Query.h"
 
 @interface DatastoreEncryptionTests : CloudantSyncTests
 
--- a/CDTDatastore.podspec
+++ b/CDTDatastore.podspec
@@ -89,11 +89,14 @@ Pod::Spec.new do |s|
 
     sp.library = 'z'
 
-    # Some CDTDatastore classes use SQLite functions, we have to include
-    # 'SQLCipher' although 'FMDB/SQLCipher' also depends on 'SQLCipher' or they
-    # will not compile (linker will not find some symbols)
-    sp.dependency 'SQLCipher'
     sp.dependency 'FMDB/SQLCipher', '= 2.3'
+
+    # Some CDTDatastore classes use SQLite functions, therefore we have
+    # to include 'SQLCipher' although 'FMDB/SQLCipher' also depends on it
+    # or they will not compile (linker will not find some symbols).
+    # Also, we have to force cocoapods to configure SQLCipher with support
+    # for FTS.
+    sp.dependency 'SQLCipher/fts'
   end
 
   s.subspec 'common-dependencies' do |sp|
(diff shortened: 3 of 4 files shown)


EncryptionTests/Encryption Tests/CDTQIndexManagerEncryptionTests.m
```diff
@@ -209,4 +210,21 @@ - (void)testCreateQueryIndexManagerWithFixedKeyProviderFailsIfDBIsEncryptedButKe
...
+{
+    CDTHelperFixedKeyProvider *provider = [CDTHelperFixedKeyProvider provider];
+    CDTDatastore *datastore = [self.factory datastoreNamed:@"is_fts_activated"
+                                 withEncryptionKeyProvider:provider
+                                                     error:nil];
+    
+    CDTQIndexManager *im = [[CDTQIndexManager alloc] initUsingDatastore:datastore error:nil];
+
+    NSSet *compileOptions = [DBQueryUtils compileOptions:im.database];
+    NSSet *ftsCompileOptions = [NSSet setWithArray:@[ @"ENABLE_FTS3", @"ENABLE_FTS3_PARENTHESIS" ]];
```
alfinkel: I think the check for `ENABLE_FTS3` and `ENABLE_FTS3_PARENTHESIS` is too restrictive. It is only necessary to confirm that the `ENABLE_FTS3` compile option is available to ensure that text search will work.  That is all that we check for with `isTextSearchEnabled` in the index manager.  `ENABLE_FTS3_PARENTHESIS` is a SQLite compile option that will allow you to use enhanced query syntax along with FTS but is not necessary for text search itself to function.

EncryptionTests/Encryption Tests/CDTQIndexManagerEncryptionTests.m
```diff
@@ -209,4 +210,21 @@ - (void)testCreateQueryIndexManagerWithFixedKeyProviderFailsIfDBIsEncryptedButKe
...
+
+    XCTAssertTrue([ftsCompileOptions isSubsetOfSet:compileOptions]);
+
+    XCTAssertTrue([im isTextSearchEnabled], @"It should be activated");
```
gadamc: My initial thought was that perhaps there could be a real test showing FTS working with a document rather than just checking for the compile options and if -isTextSearchEnabled return true

however... @mikerhodes replied that -isTextSearchEnabled should be sufficient.

And I guess we can assume that FMDB gets it right: https://github.com/cloudant/CDTDatastore/blob/master/Classes/common/query/CDTQIndexManager.m#L354-L372.

So, effectively this is a "no comment", but I put it here since you asked me nicely (on Slack).

EncryptionTests/Encryption Tests/DatastoreEncryptionTests.m
```diff
@@ -192,4 +193,15 @@ - (void)testInitReturnsNilIfEncryptionKeyProviderDoesNotReturnTheKeyUsedToOpenTh
...
 
+- (void)testTextSearchIsAvailable
+{
+    CDTHelperFixedKeyProvider *provider = [CDTHelperFixedKeyProvider provider];
+    
+    CDTDatastore *datastore = [self.factory datastoreNamed:@"test_copyprovider"
+                                 withEncryptionKeyProvider:provider
+                                                     error:nil];
+    
+    XCTAssertTrue([datastore isTextSearchEnabled], @"It should be activated");
```
gadamc: Pretty sure this test is redundant with the one above.  `CDTDatastore+Query -isTextSearchEnabled` ultimately rests upon the same bit of code: https://github.com/cloudant/CDTDatastore/blob/master/Classes/common/query/CDTQIndexManager.m#L354-L372.

EncryptionTests/Encryption Tests/CDTQIndexManagerEncryptionTests.m
alfinkel: In looking at the index manager tests, I realized that you probably copied this check from there.  Shortly, I'll be fixing it in the index manager tests to only test for `ENABLE_FTS3` as well. :)

EncryptionTests/Encryption Tests/CDTQIndexManagerEncryptionTests.m
indisoluble: @gadamc Thank you!!

EncryptionTests/Encryption Tests/CDTQIndexManagerEncryptionTests.m
indisoluble: @alfinkel: You are right, I copied this test from [here](https://github.com/cloudant/CDTDatastore/blob/master/Tests/Tests/CDTQIndexManagerTests.m#L249)

EncryptionTests/Encryption Tests/DatastoreEncryptionTests.m
indisoluble: @gadamc: @alfinkel: I also copied this test from [here](https://github.com/cloudant/CDTDatastore/blob/master/Tests/Tests/CDTDatastoreQueryTests.m#L130)

alfinkel: @indisoluble: That bit of code tests that the `CDTDatastore+Query` category exposes all of the desired methods of the index manager to the end user.  So in your case, @gadamc is correct, it is redundant to perform both checks.

EncryptionTests/Encryption Tests/CDTQIndexManagerEncryptionTests.m
indisoluble: Fixed in c3c5617

EncryptionTests/Encryption Tests/DatastoreEncryptionTests.m
indisoluble: @alfinkel: Ok, thanks.

indisoluble: Fixed in 03fc52f

alfinkel:
:+1: looks good.

rhyshort:
:+1:

PR #170, opened 2015-06-18
48048 Remove files not related to an attachment in db

indisoluble:
*What:*
Modify `TDBlobStore:deleteBlobsExceptWithKeys:withDatabase:` to delete all the files in the `TDBlobStore` folder excepts those specify in **keysToKeep**.

*Why:*
This method loops through the filenames in the database and deletes them from database and disk but if a file is only in the disk, it will never be deleted.

How can this happen? `TDBlobStore:storeBlob:creatingKey:withDatabase:error:` receives a `FMDatabase` instance as a parameter, it will use this instance to record in the database the filename for the new attachment. If this operation is cancelled (rollback), the entry with the filename will be discarded but the file will not be removed from disk.

*How:*
1. Get from db the filenames for the files to delete. Delete them from disk and db unless they are an exception.
2. Get all the remaining files in the `TDBlobStore` folder, delete all of them unless they are an exception.

*Tests:*
* Prepare a blob store with 2 attachment and 2 extra files not related to any attachment. Delete all attachments but 1. At the end of the operation, only one file (the exception) should remain.

reviewer @emlaver 
reviewer @tomblench 
reviewer @mikerhodes

### DIFF
--- a/Classes/common/touchdb/TDBlobStore.m
+++ b/Classes/common/touchdb/TDBlobStore.m
@@ -179,6 +179,9 @@ NSString *const CDTBlobStoreErrorDomain = @"CDTBlobStoreErrorDomain";
     BOOL errors = NO;
     NSUInteger numDeleted = 0;
 
+    NSMutableSet* filesToKeep = [NSMutableSet setWithCapacity:keysToKeep.count];
+
+    // Delete attachments from database and disk
     NSArray* allRows = [TD_Database rowsInBlobFilenamesTableInDatabase:db];
 
     for (TD_DatabaseBlobFilenameRow* oneRow in allRows) {
@@ -187,6 +190,8 @@ NSString *const CDTBlobStoreErrorDomain = @"CDTBlobStoreErrorDomain";
             [NSData dataWithBytes:oneRow.key.bytes length:sizeof(oneRow.key.bytes)];
         if ([keysToKeep containsObject:curKeyData]) {
             // Do not delete blob. It is an exception.
+            [filesToKeep addObject:oneRow.blobFilename];
+
             continue;
         }
 
@@ -208,7 +213,42 @@ NSString *const CDTBlobStoreErrorDomain = @"CDTBlobStoreErrorDomain";
         }
     }
 
+    // Delete files not related to an attachments
+    [self deleteFilesExceptWithFilenames:filesToKeep];
+
+    // Return
     return (errors ? -1 : numDeleted);
+}
+
+- (void)deleteFilesExceptWithFilenames:(NSSet*)filesToKeep
+{
+    NSFileManager* defaultManager = [NSFileManager defaultManager];
+
+    // Read directory
+    NSError* thisError = nil;
+    NSArray* remainingFiles = [defaultManager contentsOfDirectoryAtPath:_path error:&thisError];
+    if (!remainingFiles) {
+        CDTLogError(CDTDATASTORE_LOG_CONTEXT, @"Can not read dir %@: %@", _path, thisError);
+
+        return;
+    }
+
+    // Delete all files but exceptions
+    for (NSString* oneFilename in remainingFiles) {
+        if ([filesToKeep containsObject:oneFilename]) {
+            // Do not delete file. It is an exception.
+
+            continue;
+        }
+
+        NSString* filePath = [TDBlobStore blobPathWithStorePath:_path blobFilename:oneFilename];
+
+        if (![defaultManager removeItemAtPath:filePath error:&thisError]) {
+            CDTLogError(CDTDATASTORE_LOG_CONTEXT,
+                        @"%@: Failed to delete '%@' not related to an attachment: %@", self,
+                        oneFilename, thisError);
+        }
+    }
 }
 
 - (NSString*)tempDir
--- a/Classes/common/touchdb/TDBlobStore.m
+++ b/Classes/common/touchdb/TDBlobStore.m
@@ -237,7 +237,6 @@ NSString *const CDTBlobStoreErrorDomain = @"CDTBlobStoreErrorDomain";
     for (NSString* oneFilename in remainingFiles) {
         if ([filesToKeep containsObject:oneFilename]) {
             // Do not delete file. It is an exception.
-
             continue;
         }
 
--- a/Classes/common/touchdb/TDBlobStore.m
+++ b/Classes/common/touchdb/TDBlobStore.m
@@ -214,37 +214,37 @@ NSString *const CDTBlobStoreErrorDomain = @"CDTBlobStoreErrorDomain";
     }
 
     // Delete files not related to an attachments
-    [self deleteFilesExceptWithFilenames:filesToKeep];
+    [TDBlobStore deleteFilesNotInSet:filesToKeep fromPath:_path];
 
     // Return
     return success;
 }
 
-- (void)deleteFilesExceptWithFilenames:(NSSet*)filesToKeep
++ (void)deleteFilesNotInSet:(NSSet*)filesToKeep fromPath:(NSString *)path
 {
     NSFileManager* defaultManager = [NSFileManager defaultManager];
 
     // Read directory
     NSError* thisError = nil;
-    NSArray* remainingFiles = [defaultManager contentsOfDirectoryAtPath:_path error:&thisError];
-    if (!remainingFiles) {
-        CDTLogError(CDTDATASTORE_LOG_CONTEXT, @"Can not read dir %@: %@", _path, thisError);
+    NSArray* currentFiles = [defaultManager contentsOfDirectoryAtPath:path error:&thisError];
+    if (!currentFiles) {
+        CDTLogError(CDTDATASTORE_LOG_CONTEXT, @"Can not read dir %@: %@", path, thisError);
         return;
     }
 
     // Delete all files but exceptions
-    for (NSString* oneFilename in remainingFiles) {
-        if ([filesToKeep containsObject:oneFilename]) {
+    for (NSString* filename in currentFiles) {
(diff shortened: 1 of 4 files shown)


Classes/common/touchdb/TDBlobStore.m
```diff
@@ -187,6 +190,8 @@ - (NSInteger)deleteBlobsExceptWithKeys:(NSSet*)keysToKeep withDatabase:(FMDataba
             [NSData dataWithBytes:oneRow.key.bytes length:sizeof(oneRow.key.bytes)];
         if ([keysToKeep containsObject:curKeyData]) {
             // Do not delete blob. It is an exception.
+            [filesToKeep addObject:oneRow.blobFilename];
+
             continue;
         }
```
tomblench: You can remove this chunk of code down here because the deletion now gets deleted by `deleteFilesExceptWithFilenames`

indisoluble: But if I do that, I will not update the value of [numDeleted](https://github.com/cloudant/CDTDatastore/pull/170/files#diff-ce77b2812a143c42aa7fdb8b3818581dR207) ... Uhm I realise of another situation:
1. The file is not deleted so I do not remove the row from db
2. I tried to delete again the same file in `deleteFilesExceptWithFilenames:` and this time, it works
I'll end up with a row in the database without a file.

tomblench: `numDeleted` is only used in `garbageCollectAttachments` to check for non-zero return value and for the log message.

So how about changing the signature of this method to return true/false and put the log message where the deletion actually happens, in `deleteFilesExceptWithFilenames`?

At the moment the logic seems a bit silly, where it's "delete them here and maybe try to delete them over there as well"

indisoluble: You are right, `numDeleted` is only used in `garbageCollectAttachments` but also notice that if we are able to delete the file from disk, we [delete the row from db](https://github.com/cloudant/CDTDatastore/pull/170/files#diff-ce77b2812a143c42aa7fdb8b3818581dR205).

Probably `deleteFilesExceptWithFilenames:` is not a good name or it does not explain completely what it does: it deletes the files in disk that are not related to an attachment in db.

indisoluble: `deleteFilesExceptWithFilenames:` returns a BOOL. Also, if an attachment could not be deleted, I do not try to delete it again because it won't be deleted from db.

Let me know what you think.

Fixed in f14f8e2

mikerhodes: After these changes it seems redundant to attempt to delete the file here, given it would be deleted anyway by `deleteFilesExceptWithFilenames` lower down?

Is the reason for this that otherwise we might delete a row in the attachments table, then assign the same filename later but still have the old file with that file name still on disk (so end up with the wrong attachment data because the new attachment data would not be saved over it?)

Classes/common/touchdb/TDBlobStore.m
```diff
@@ -198,17 +202,53 @@ - (NSInteger)deleteBlobsExceptWithKeys:(NSSet*)keysToKeep withDatabase:(FMDataba
         if ([[NSFileManager defaultManager] removeItemAtPath:blobPath error:&thisError]) {
```
mikerhodes: If the file doesn't exist, does this method return `YES`? I ask as we can also remove the database row for the key if the file didn't exist in the first place (for some reason).

Classes/common/touchdb/TDBlobStore.m
```diff
@@ -198,17 +202,53 @@ - (NSInteger)deleteBlobsExceptWithKeys:(NSSet*)keysToKeep withDatabase:(FMDataba
...
+    NSError* thisError = nil;
+    NSArray* remainingFiles = [defaultManager contentsOfDirectoryAtPath:_path error:&thisError];
+    if (!remainingFiles) {
+        CDTLogError(CDTDATASTORE_LOG_CONTEXT, @"Can not read dir %@: %@", _path, thisError);
+
+        return;
+    }
+
+    // Delete all files but exceptions
+    for (NSString* oneFilename in remainingFiles) {
```
mikerhodes: Remove the `one` prefix where a singular would do, `oneFilename` is the same as `filename`?

Also, I can see you named `remainingFiles` to be "remaining after deleting the other ones in the method that calls this one", however, if you only consider the context of this method, the name doesn't make sense (remaining files implies remaining after an operation, but getting the filenames is the first thing that this method does). I suggest `currentFiles` or just `filenames`.

I'd also be tempted to pass the path into the method, gives you a nicer name: `deleteFilesNotInSet:fromPath:`.

Classes/common/touchdb/TDBlobStore.m
```diff
@@ -198,17 +202,53 @@ - (NSInteger)deleteBlobsExceptWithKeys:(NSSet*)keysToKeep withDatabase:(FMDataba
...
+- (void)deleteFilesExceptWithFilenames:(NSSet*)filesToKeep
+{
+    NSFileManager* defaultManager = [NSFileManager defaultManager];
+
+    // Read directory
+    NSError* thisError = nil;
+    NSArray* remainingFiles = [defaultManager contentsOfDirectoryAtPath:_path error:&thisError];
+    if (!remainingFiles) {
+        CDTLogError(CDTDATASTORE_LOG_CONTEXT, @"Can not read dir %@: %@", _path, thisError);
+
```
mikerhodes: No need for the newline

Classes/common/touchdb/TDBlobStore.m
```diff
@@ -198,17 +202,53 @@ - (NSInteger)deleteBlobsExceptWithKeys:(NSSet*)keysToKeep withDatabase:(FMDataba
...
+        return;
+    }
+
+    // Delete all files but exceptions
+    for (NSString* oneFilename in remainingFiles) {
+        if ([filesToKeep containsObject:oneFilename]) {
+            // Do not delete file. It is an exception.
+
+            continue;
```
mikerhodes: No need for the newline

Classes/common/touchdb/TD_Database+Attachments.m
```diff
@@ -818,11 +818,11 @@ - (TDStatus)garbageCollectAttachments:(FMDatabase*)db
...
     }
     [r close];
-    NSInteger numDeleted = [_attachments deleteBlobsExceptWithKeys:allKeys withDatabase:db];
-    if (numDeleted < 0) {
+    BOOL blobDeleted = [_attachments deleteBlobsExceptWithKeys:allKeys withDatabase:db];
+    if (!blobDeleted) {
         return kTDStatusAttachmentError;
     }
-    CDTLogInfo(CDTDATASTORE_LOG_CONTEXT, @"Deleted %d attachments", (int)numDeleted);
+    CDTLogInfo(CDTDATASTORE_LOG_CONTEXT, @"Attachments deleted");
```
mikerhodes: Suggest making this a little less scary, `Unneeded attachment blobs deleted.`

Tests/Tests/Encryption/TDBlobStoreEncryptionTests.m
```diff
@@ -362,6 +362,55 @@ - (void)testDeleteBlobsExceptWithKeysRemovesExpectedRowsFromDBAndDisk
...
+    // Assert
+    NSFileManager *defaultManager = [NSFileManager defaultManager];
+
+    NSArray *allFilenames =
+        [defaultManager contentsOfDirectoryAtPath:self.encryptedBlobStorePath error:nil];
+    XCTAssertEqual(allFilenames.count, 1, @"Only 1 file should remain");
+
+    filePath = [self.encryptedBlobStorePath stringByAppendingPathComponent:filename];
+    XCTAssertTrue([defaultManager fileExistsAtPath:filePath],
+                  @"There should be a file at this path");
```
mikerhodes: I'd create maybe a few tens of valid and invalid files to test that there's not something special when you have just one valid file.

Classes/common/touchdb/TD_Database+Attachments.m
indisoluble: Fixed in bf6639f

Tests/Tests/Encryption/TDBlobStoreEncryptionTests.m
indisoluble: Fixed in b9e9cd0

Classes/common/touchdb/TDBlobStore.m
indisoluble: Fixed in a95e7a6

Classes/common/touchdb/TDBlobStore.m
indisoluble: Fixed in a21488d

Classes/common/touchdb/TDBlobStore.m
indisoluble: Fixed in f387c98

Classes/common/touchdb/TDBlobStore.m
indisoluble: Fixed in 9712a1b

Classes/common/touchdb/TDBlobStore.m
indisoluble: @mikerhodes: If we ever generate the same filename, the content of the file will be overwritten with the new data.

indisoluble: Fixed in c75c671

mikerhodes:
lgtm +1

emlaver:
+1.  The variable and method name changes clarified my understanding.

tomblench:
OK +1

PR #173, opened 2015-06-25
Fix query without any index capability

alfinkel:
_What:_

There is a bug in Query that requires at least one index be created before any query can execute.

_Why:_

The desired behavior is to allow for querying of documents regardless of the existence of any indexes.

_How:_

Remove the logic that causes a query to fail when no indexes are found and replace it with logic that catches the condition and instead retrieves a full list of document ids from the main datastore for the post-hoc matcher to use when generating the query results.

This logic is found in the SQL translator and the query executor.  Additionally a new method to retrieve just a list of document ids needs to be added to the datastore API.

reviewer @mikerhodes 
reviewer @gadamc 

BugId: 45362

### DIFF
--- a/Classes/common/CDTDatastore.m
+++ b/Classes/common/CDTDatastore.m
@@ -288,6 +288,33 @@ NSString *const CDTDatastoreChangeNotification = @"CDTDatastoreChangeNotificatio
     return result;
 }
 
+- (NSArray *)getAllDocumentIds
+{
+    if (![self ensureDatabaseOpen]) {
+        return nil;
+    }
+    
+    NSMutableArray *result = [NSMutableArray array];
+    struct TDQueryOptions query = {.limit = UINT_MAX,
+                                   .inclusiveEnd = YES,
+                                   .skip = 0,
+                                   .descending = NO,
+                                   .includeDocs = NO};
+    
+    NSDictionary *dictResults;
+    do {
+        dictResults = [self.database getDocsWithIDs:nil options:&query];
+        for (NSDictionary *row in dictResults[@"rows"]) {
+            [result addObject:row[@"id"]];
+        }
+        
+        query.skip = query.skip + query.limit;
+    } while (((NSArray *)dictResults[@"rows"]).count > 0);
+    
+    return [NSArray arrayWithArray:result];
+    
+}
+
 - (NSArray *)getAllDocumentsOffset:(NSUInteger)offset
                              limit:(NSUInteger)limit
                         descending:(BOOL)descending
--- a/Classes/common/query/CDTQQuerySqlTranslator.m
+++ b/Classes/common/query/CDTQQuerySqlTranslator.m
@@ -85,26 +85,19 @@
         return nil;
     } else if (!state.textIndexRequired &&
                   (!state.atLeastOneIndexUsed || state.atLeastOneORIndexMissing)) {
-        // If we haven't used a single index, we need to return a query
-        // which returns every document, so the posthoc matcher can
+        // If we haven't used a single index or an OR clause is missing an index,
+        // we need to return every document id, so that the post-hoc matcher can
         // run over every document to manually carry out the query.
-        NSSet *neededFields = [NSSet setWithObject:@"_id"];
-        NSString *allDocsIndex =
-            [CDTQQuerySqlTranslator chooseIndexForFields:neededFields fromIndexes:indexes];
-
-        if (!allDocsIndex) {
-            LogError(@"No indexes defined, cannot execute query for all documents");
-            return nil;
-        }
-
-        NSString *tableName = [CDTQIndexManager tableNameForIndex:allDocsIndex];
-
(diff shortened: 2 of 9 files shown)


Classes/common/query/CDTQQuerySqlTranslator.m
```diff
@@ -85,27 +85,20 @@ + (CDTQQueryNode *)translateQuery:(NSDictionary *)query
...
         NSSet *neededFields = [NSSet setWithObject:@"_id"];
-        NSString *allDocsIndex =
-            [CDTQQuerySqlTranslator chooseIndexForFields:neededFields fromIndexes:indexes];
+        NSString *allDocsIndex = [CDTQQuerySqlTranslator chooseIndexForFields:neededFields
+                                                                  fromIndexes:indexes];
 
-        if (!allDocsIndex) {
-            LogError(@"No indexes defined, cannot execute query for all documents");
-            return nil;
+        if (allDocsIndex.length > 0) {
```
gadamc: I'd be concerned if we didn't get into this block. This should be somewhat guaranteed right? There's always an index for the "_id" key, right?

If this is true, and allDocsIndex.length == 0 for some crazy, what should be done? Should it send a Log message and bail?

alfinkel: In order for this fix to work, we populate the SQLNode with a SELECT statement ONLY if an index table exists.  If an index does not exist then we obviously cannot construct a SELECT statement.  Instead we leave the SQLNode alone and when the query node gets passed back to the query executor, the executor (https://github.com/cloudant/CDTDatastore/blob/45362-fix-create-default-index/Classes/common/query/CDTQQueryExecutor.m#L267) in turn will handle the "empty" SQL SELECT condition by getting the list of doc ids from the datastore.

Classes/common/CDTDatastore.m
```diff
@@ -289,6 +289,41 @@ - (NSArray *)getAllDocuments
...
+        
+        result = [result arrayByAddingObjectsFromArray:batch];
+        
+        done = ((NSArray *)dictResults[@"rows"]).count == 0;
+        
+        query.skip = query.skip + query.limit;
+        
+    } while (!done);
+    
+    return result;
```
gadamc: The only thing I could suggest here is just using fewer arrays

```
    NSMutableArray *result = [NSMutableArray array];
    struct TDQueryOptions query = {.limit = UINT_MAX,
                                   .inclusiveEnd = YES,
                                   .skip = 0,
                                   .descending = NO,
                                   .includeDocs = NO};
    
    BOOL done = NO;
    do {        
        NSDictionary *dictResults = [self.database getDocsWithIDs:nil options:&query];
        for (NSDictionary *row in dictResults[@"rows"]) {
            [result addObject:row[@"id"]];
        }
             
        done = ((NSArray *)dictResults[@"rows"]).count == 0;   
        query.skip = query.skip + query.limit;
        
    } while (!done);
    
    return [NSArray arrayWithArray:result];

```

gadamc:
Other than that minor detail about using fewer arrays, it seems okay to me. +1

Classes/common/query/CDTQQuerySqlTranslator.m
mikerhodes: This patch is for the case when no indexes are defined, meaning that this condition is true.

Classes/common/CDTDatastore.m
mikerhodes: Adam's code is a little clearer to my eyes.

Tests/Tests/DatastoreCRUD.m
```diff
@@ -765,6 +765,40 @@ -(void)test_getAllDocumentsOffsetLimitDescending
     //[self getAllDocuments_testCountAndOffset:objectCount expectedDbObjects:reversedObjects descending:YES];
 }
 
+-(void)testGetAllDocumentIds
+{
+    XCTAssertEqual([self.datastore getAllDocumentIds].count, 0, @"No documents should exist.");
+    
+    NSError *error;
+    int objectCount = 5;
```
mikerhodes: tbh I'd put a larger number, e.g., 1000, here.

alfinkel: Changed `objectCount` to 1000 in 45e17aa70c46c0a27263162d7bf9b9e64a31f29b

Classes/common/CDTDatastore.m
alfinkel: I copied this logic from `getAllDocuments` which uses `done` to manage the `do...while` loop.  It is essentially the same logic.  I made the change suggested by @gadamc in `getAllDocumentIds`.  Let me know if I should update the `getAllDocuments` logic as well, as part of this PR, open an FB case for the work, or simply leave it alone.

Also, in order to remove `done` from the `do...while`, `dictResults` needs to be declared outside of the loop.  This is done in 45e17aa70c46c0a27263162d7bf9b9e64a31f29b.

mikerhodes:
+1

PR #176, opened 2015-06-29
45943 Deleting document fails with a mutable document revision

indisoluble:
*What:*
Using a CDTMutableDocumentRevsion to delete a document fails. It fails with a "conflict" message, even though the object's sourceRevId is correct.

*Why:*
Method `CDTDatastore:deleteDocumentFromRevision:error:` tries to get the revision ID from the property `revision.revId`; however a `CDTMutableDocumentRevision` stores this value in `sourceRevId`.

*How:*
Check if `revision` is an instance of `CDTMutableDocumentRevision` and in that case, take the revision if from property `sourceRevId`

*Tests:*
One extra tests in `DatastoreCRUD.m`:
* Create a document. Get a mutable copy. Delete the document with the mutable copy.

reviewer @rhyshort 
reviewer @emlaver 
reviewer @tomblench

### DIFF
--- a/Tests/Tests/DatastoreCRUD.m
+++ b/Tests/Tests/DatastoreCRUD.m
@@ -1797,6 +1797,26 @@
     
 }
 
+- (void)testDeleteMutableDocumentRevision
+{
+    NSString *docId = @"testDeleteWithMutableDocumentRevision";
+
+    CDTMutableDocumentRevision *mutableRev = [CDTMutableDocumentRevision revision];
+    mutableRev.docId = docId;
+    mutableRev.body = @{ @"hello" : @"world" };
+    CDTDocumentRevision *rev = [self.datastore createDocumentFromRevision:mutableRev error:nil];
+
+    mutableRev = [rev mutableCopy];
+
+    NSError *error = nil;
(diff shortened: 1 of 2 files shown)


rhyshort:
+1

Tests/Tests/DatastoreCRUD.m
```diff
@@ -1797,6 +1797,20 @@ -(void) testDeleteDocumentUsingIdHasMultipleLeafNodesInTree
...
+    mutableRev.body = @{ @"hello" : @"world" };
+
+    NSError *error;
+    CDTDocumentRevision *rev = [self.datastore createDocumentFromRevision:mutableRev error:&error];
+
+    mutableRev = [rev mutableCopy];
+
+    XCTAssertNotNil([self.datastore deleteDocumentFromRevision:mutableRev error:&error],
+                    @"It should not fail: %@", error);
+}
```
tomblench: Now try to fetch the document from the datastore and assert that it has the deleted flag set

indisoluble: Fixed in a37d4dc

emlaver:
+1

tomblench:
+1

PR #177, opened 2015-06-30
46052 Implement resultsLimit and moreComing in CDTFetchChanges

indisoluble:
*What:*
It's difficult to batch updates in `CDTQIndexUpdater` while recording the latest sequence number.

*Why:*
Because the sequence number is only delivered in the completion block. Therefore we insert batches of document changes into the index tables but don't update the last sequence number.

*How:*
Implementing `resultsLimit` and `moreComing` would allow the indexing code to more efficiently batch changes, while bookmarking where it's up to as it goes along.

*Tests:*
1. Set `resultsLimit` to a value smaller than the total number of documents in the datastore. The `CDTFetchChanges` instance has to return a number of documents equal to the limit and `moreComing` has to be `true`.
2. Set `resultsLimit` to the total number of values in the datastore. The `CDTFetchChanges` instance has to return all the documents in the datastore and `moreComing` has to be `false`.
3. Set `resultsLimit` to a number bigger than the total number of documents in the datastore. The `CDTFetchChanges` instance will return all the documents in the datastore and `moreComing` will be `false`.
4. Create a `CDTFetchChanges` instance with `resultsLimit` equal to a value smaller than the total number of documents in the datastore. Fetch the documents and save the sequence value and the number of documents read so far. Create a new  `CDTFetchChanges` instance with `resultsLimit` equal to a value smaller than the total number of documents in the datastore and fetch the rest of the documents using the sequence value mentioned before. The total number of documents read has to be equal to the total number of documents in the datastore.

reviewer @rhyshort 
reviewer @tomblench

### DIFF
--- a/Classes/common/CDTFetchChanges.h
+++ b/Classes/common/CDTFetchChanges.h
@@ -67,6 +67,28 @@
  */
 @property (nonatomic, copy) NSString *startSequenceValue;
 
+/**
+ Use this property to limit the total number of results you want to get.
+ 
+ By default, its value 0 which means that it will returns as many as possible.
+ 
+ Set it before executing the operation or submitting it to a queue.
+ */
+@property(nonatomic, assign) NSUInteger resultsLimit;
+
+/**
+ It will be YES only if there are more results available.
+ 
+ A CDTFetchChanges instance will deliver all results available unless a limit is set with
+ resultsLimit.
+ 
+ This property will set to YES before calling fetchRecordChangesCompletionBlock.
+ 
+ @see resultsLimit
+ @see fetchRecordChangesCompletionBlock
+ */
+@property(nonatomic, readonly) BOOL moreComing;
+
 #pragma mark Callbacks
 
 /**
--- a/Classes/common/CDTFetchChanges.m
+++ b/Classes/common/CDTFetchChanges.m
@@ -20,6 +20,8 @@
 
 #import "TD_Database.h"
 
+#define CDTFETCHCHANGES_DEFAULT_OPTION_LIMIT 500
+
 @interface CDTFetchChanges ()
 
 @property (nonatomic, copy) void (^mDocumentChangedBlock)(CDTDocumentRevision *revision);
@@ -46,6 +48,8 @@
     if (self) {
         _datastore = datastore;
         _startSequenceValue = [startSequenceValue copy];
+        _resultsLimit = 0;
+        _moreComing = NO;
     }
     return self;
 }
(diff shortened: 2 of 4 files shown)


Classes/common/CDTFetchChanges.m
```diff
@@ -61,16 +65,23 @@ - (void)main
...
 
-    TDChangesOptions options = {.limit = 500,
-        .contentOptions = 0,                                
+    NSUInteger mResultsLimit = self.resultsLimit;
+    BOOL thereIsALimit = (mResultsLimit > 0);
+
+    TDChangesOptions options = {
+        .limit = (thereIsALimit && (mResultsLimit < CDTFETCHCHANGES_DEFAULT_OPTION_LIMIT)
+                      ? (unsigned)mResultsLimit
+                      : CDTFETCHCHANGES_DEFAULT_OPTION_LIMIT),
```
rhyshort: I think the setting of `.limit` would be easier to read if it was was a `if` statement.

Classes/common/CDTFetchChanges.h
```diff
@@ -67,6 +67,28 @@
...
 @property (nonatomic, copy) NSString *startSequenceValue;
 
+/**
+ Use this property to limit the total number of results you want to get.
+ 
+ By default, its value 0 which means that it will returns as many as possible.
+ 
+ Set it before executing the operation or submitting it to a queue.
+ */
+@property(nonatomic, assign) NSUInteger resultsLimit;
```
rhyshort: It should be documented that if the resultsLimit is above the default defined in the implementation it will be ignored and the default used.

Classes/common/CDTFetchChanges.m
```diff
@@ -61,16 +65,23 @@ - (void)main
...
         .includeConflicts = FALSE,
         .sortBySequence = TRUE};
-    
+
+    BOOL doLoop = YES;
     TD_RevisionList *changes;
     SequenceNumber lastSequence = [self.mStartSequenceValue longLongValue];
 
-    do {
+    while (doLoop) {
```
tomblench: this can be `while (doLoop && ![self isCancelled])`

Classes/common/CDTFetchChanges.m
```diff
@@ -79,12 +90,46 @@ - (void)main
...
+                // not more data coming and we can stop looping.
+                doLoop = NO;
+            } else {
+                // Subtract the results so far
+                mResultsLimit -= [changes count];
+
+                if (mResultsLimit == 0) {
+                    // We reached the limit but we need to know if there are more data coming.
+                    // Loop again
+                    options.limit = 1;
```
tomblench: I don't really understand this. Is the purpose to see if there is at least one more change available from `changesSinceSequence`? it seems like a bit of a hack

Classes/common/CDTFetchChanges.m
indisoluble: Fixed in 95eb6a3

Classes/common/CDTFetchChanges.m
```diff
@@ -79,12 +90,46 @@ - (void)main
...
+        if (!thereIsALimit) {
+            // Limit not set. Get as many results as we can.
+            lastSequence = [self notifyChanges:changes startingSequence:lastSequence];
+
+            doLoop = (changes.count > 0);
+        } else if (mResultsLimit > 0) {
+            // There is a limit but we have not reached it yet.
+            lastSequence = [self notifyChanges:changes startingSequence:lastSequence];
+
+            if (changes.count == 0) {
```
tomblench: I wonder if it's possible to simplify the logic in these if/else branches. You have
```
lastSequence = [self notifyChanges:changes startingSequence:lastSequence];
```
in two branches (this one and the one above) and also in those same branches you have
```
doLoop = (changes.count > 0);
```
and 
```
if (changes.count == 0) {
  doLoop = NO;
}
```
which mean the same thing.

Classes/common/CDTFetchChanges.h
indisoluble: Do you refer to this [line](https://github.com/cloudant/CDTDatastore/pull/177/files#diff-bac15dfde454859cd8e3a8dc04fcf29fR72)? It is not ignored, `CDTFetchChanges` gets the changes in the datastore in groups of 500 records or less if `resultsLimit` is smaller than that. But if `resultsLimit` is bigger than 500, `CDTFetchChanges` will get the changes in groups of 500 records until it reaches the limit.

Classes/common/CDTFetchChanges.m
indisoluble: That is exactly what I'm doing. Once I get all the requested changes and the next sequence value, I check if there are more data to set `moreComing` to the right value.

Classes/common/CDTFetchChanges.h
rhyshort: oh right I see. The inline if had me confused what was going on

Classes/common/CDTFetchChanges.m
indisoluble: That is true, I have:

```
lastSequence = [self notifyChanges:changes startingSequence:lastSequence];
```

In 2 branches but not in the [third one](https://github.com/cloudant/CDTDatastore/pull/177/files#diff-bac15dfde454859cd8e3a8dc04fcf29fR127) where I only update `moreComing`.

About:

```
doLoop = (changes.count > 0);
```

And

```
if (changes.count == 0) {
  doLoop = NO;
}
```

The `if` right above also has an [else](https://github.com/cloudant/CDTDatastore/pull/177/files#diff-bac15dfde454859cd8e3a8dc04fcf29fR109) where I prepare the `.limit` for the next loop.

Classes/common/CDTFetchChanges.m
indisoluble: I mean, I have to return at most the number of results specified in `resultsLimit` and the `newSequenceValue` has to be equal to the sequence of the last result returned. As I see it I have no other options but to make one last `select` to see if there are more data after `newSequenceValue`.

Classes/common/CDTFetchChanges.m
indisoluble: Fixed in b5aaecc

Classes/common/CDTFetchChanges.m
indisoluble: Fixed in 3400658

rhyshort:
I'd like to see a test where the limit is set to `-1` to be sure it the API behaves as if the limit wasn't set the guard against potential regressions.

indisoluble:
@rhyshort [resultLimit](https://github.com/cloudant/CDTDatastore/pull/177/files#diff-5def4eccc422b6baec13e859be3128b4R77) is a `NSUInteger`:

```
typedef unsigned long NSUInteger;
```

rhyshort:
oops, well its looks like I'm having a bad day.

rhyshort:
+1

tomblench:
+1

PR #180, opened 2015-07-15
Correct init chain for CDTDocumentRevision

rhyshort:
## What

Correct the init chain for CDTDocumentRevision so each init method
calls down into the full argument version of init. Change document body and attachments to initalize with a empty dictionary if nil is passed for their arguments.

## How

Make all the init methods call down to one single designated initialiser for `CDTMutableDocumentRevision`. The interface for `CDTMutableDocumentRevision` has changed as well, instead of `body` and `attachments` being defined as nil by default, they are now specified as empty dictionaries by default.

## Why

The implementation this replaces is confused, and doesn't use the designated initialiser, this work makes it so it matches the coding guidelines. It also improves the behaviour to make it easier to map into Swift using Nullability annotations.

## Reviewers

reviewer @alfinkel 
reviewer @mikerhodes  

BugzId: 45942

### DIFF
--- a/Classes/common/CDTMutableDocumentRevision.h
+++ b/Classes/common/CDTMutableDocumentRevision.h
@@ -13,11 +13,39 @@
 @property (nonatomic, strong, readwrite) NSString *sourceRevId;
 @property (nonatomic, strong, readwrite) NSString *docId;
 
+/**
+ *   Creates an empty CDTMutableDocumentRevision
+ **/
 + (CDTMutableDocumentRevision *)revision;
 
-- (id)initWithDocumentId:(NSString *)documentId body:(NSMutableDictionary *)body;
+/**
+ * Initializes a CDTMutableDocumentRevision revision
+ *
+ * @param documentId The id of the document
+ * @param body The body of the document
+ *
+ **/
+- (instancetype)initWithDocumentId:(NSString *)documentId body:(NSMutableDictionary *)body;
 
-- (id)initWithSourceRevisionId:(NSString *)sourceRevId;
+/**
+ * Initializes a CDTMutableDocumentRevision
+ * 
+ * @param sourceRevId the parent revision id
+ **/
+- (instancetype)initWithSourceRevisionId:(NSString *)sourceRevId;
+
+/**
+ Initializes a CDTMutableDocumentRevision
+ 
+ @param documentId the id of the document
+ @param body the body of the document
+ @param attachments the document's attachments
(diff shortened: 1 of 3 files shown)


mikerhodes:
Looks good bar the fact it needs to be run through formatting.

Classes/common/CDTMutableDocumentRevision.h
```diff
@@ -13,11 +13,39 @@
...
+ * @param sourceRevId the parent revision id
+ **/
+- (instancetype)initWithSourceRevisionId:(NSString *)sourceRevId;
+
+/**
+ Initializes a CDTMutableDocumentRevision
+ 
+ @param documentId the id of the document
+ @param body the body of the document
+ @param attachment's the documents attachments
```
mikerhodes: `document's` not `attachment's`.

Tests/Tests/DatastoreCRUD.m
```diff
@@ -478,10 +481,19 @@ -(void)testCreateWithaDocumentIdNoBodyCDTMutableDocumentRevision{
...
     XCTAssertNil(saved, @"Document with Id but no body created");
 }
 
+- (void)testCreateWithDefaultMutableRevision {
+    NSError *error;
+    CDTMutableDocumentRevision *doc = [CDTMutableDocumentRevision revision];
+    CDTDocumentRevision *saved = [self.datastore createDocumentFromRevision:doc error:&error];
+    XCTAssertNotNil(saved, @"Default document was not created");
+    XCTAssertNil(error,"reccieved an error, expected no error");
+}
```
mikerhodes: This just tests that you can save an unmodified document, it doesn't check whether the changes you've made in this PR accomplish what it sets out to do. Add:

You can write to the body  without initialising the dict:

```objc
CDTMutableDocumentRevision *doc = [CDTMutableDocumentRevision revision];
doc.body[@"name"] = @"mike";
```

You can add to the attachments without initialising the dict:

```objc
CDTMutableDocumentRevision *doc = [CDTMutableDocumentRevision revision];
doc.attachments[@"name"] =  /* attachment */;
```

rhyshort: done in 441e027

Classes/common/CDTMutableDocumentRevision.h
rhyshort: fixed in 8683638

alfinkel:
+1

mikerhodes:
+1

PR #182, opened 2015-07-23
Fix to ensure that property type matches

bluemixgaragelondon:
The property was declared differently in the protocol.
The problem was detected on Xcode7-beta.


rhyshort:
Despite travis' unhappiness ( I'm not sure why) it builds on my machine, and the tests pass. I'm merging to master. Thanks for the patch @bluemixgaragelondon

PR #183, opened 2015-07-28
Make CDTDocumentRevision as it's subclasses easier to use in swift

rhyshort:
## What

Make CDTDocument Revision and it's subclasses easier to use in swift.

## Why

Currently when interacting with swift you need to use the setters and getters for `attachments` and `body` so your code looks like:

```swift
let doc = CDTMutableDocumentRevision.revision()
doc.setBody(["hello":"world"])

let body = doc.body()

```

with this change code looks like: 

```swift
let doc = CDTMutableDocumentRevision.revision()
doc.body = ["hello":"world"]

let body = doc.body

```

The change make it a lot easier to work with the APIs..

## How

Instead of using methods to make the interface seem like properties for `body` and `attachments` actually use properties and `@synthesize` to correctly override the properties set in the superclass.

## Reviewers

reviewer @tomblench 
reviewer @gadamc


Classes/common/CDTMutableDocumentRevision.m
```diff
@@ -50,25 +49,14 @@ - (instancetype)initWithDocumentId:(NSString*) documentId
...
 
-- (void)setBody:(NSDictionary *)body { self.private_body = [body mutableCopy]; }
-
-- (NSMutableDictionary *)body { return self.private_body; }
-
-- (NSMutableDictionary *)attachments { return self.private_attachments; }
-
-- (void)setAttachments:(NSMutableDictionary *)attachments
-{
-    self.private_attachments = [attachments mutableCopy];
```
tomblench: This changes the contract for the `attachments` setter. Previously we made a copy of the dictionary, now we're just storing the pointer. Is this intentional and if we're making this change is it likely to cause problems?

rhyshort: although it changes the contract, add `copy` to the property results in errors since the calling `copy` on `NSMutableDictionary` results in an instance of `NSDictionary` so we can only store the pointer. There are a bunch of warnings in the tests, which I will go and clean up though.

gadamc: I have't fully thought through this suggestion, but: Can we make the `attachments` an `NSDictionary` instead of a mutable one?  Would that still allow us to make a `copy` and still keep the CDTMutableDocumentRevision a mutable thing?

rhyshort: We could but then it would make it harder for someone to use. Instead of 
```objc
CDTMutableDocumentRevision *rev = [CDTMutableDocumentRevision revision];
rev.attachments["myAttachment"] = attachment;
```
we would be forcing a use to 
```objc
CDTMutableDocumentRevision *rev = [CDTMutableDocumentRevision revision];
NSMutableDictionary *attachments = [rev.attachments mutableCopy];
attachments["myAttachment"] = attachment;
rev.attachments
```
 I think the former is a better solution, it keeps the current syntax (no breaking changes), and it is still easy to use.

Tests/Tests/DatastoreConflictResolvers.m
```diff
@@ -188,8 +188,8 @@ -(id) init
 -(CDTDocumentRevision *)resolve:(NSString *)docId conflicts:(NSArray *)conflicts
 {
     CDTMutableDocumentRevision * mutableRev = [CDTMutableDocumentRevision revision];
-    mutableRev.body = @{};
-    mutableRev.attachments = @{};
+    mutableRev.body = [NSMutableDictionary dictionary];
```
tomblench: These lines can now be removed because body and attachments will already have been constructed

Tests/Tests/Attachments/AttachmentCRUD.m
```diff
@@ -1121,7 +1121,7 @@ - (void)testRetriveAttachmentsViaAllDocuments
     CDTDocumentRevision *rev = [self.datastore createDocumentFromRevision:document
                                                                     error:&error];
     document = [rev mutableCopy];
-    document.attachments = @{};
+    document.attachments = [NSMutableDictionary dictionary];
```
tomblench: This can be removed

Tests/Tests/Attachments/AttachmentCRUD.m
```diff
@@ -288,7 +288,7 @@ - (void)testCreateWithMutableBodyAndAttachments
...
```
tomblench: This can be removed

Tests/Tests/Attachments/AttachmentCRUD.m
```diff
@@ -212,12 +212,12 @@ - (void)testCreate
     NSDictionary *dict = @{@"hello": @"world"};
     
     CDTMutableDocumentRevision *document = [CDTMutableDocumentRevision revision];
-    document.body = dict;
+    document.body = [dict mutableCopy];
     
     CDTDocumentRevision *rev = [self.datastore createDocumentFromRevision:document
                                                                     error:&error];
     document = [rev mutableCopy];
-    document.attachments = @{};
```
tomblench: This can be removed

Classes/common/CDTMutableDocumentRevision.m
mikerhodes: The reason we have the awkward `private_blah` pattern is to allow people to set `NSDictionary` objects for the body and attachments, which we then make mutable copies of. You can see this pain below, with the changes to the tests where setting `body` on mutable document revisions now requires a `mutableCopy` call:

```
rev.body = [@{ @"name" : @"mike", @"age" : @34, @"pet" : @"cat" } mutableCopy];
```

This is a breaking change I'd rather avoid. Is there a way to simplify the Swift developer's code without losing this ability?

For my money, the better fix here is removing the concept of a mutable revision in its entirety, which we've not yet committed to doing.

rhyshort: I could't figure out another way, as far as I've gathered swift properties need to be actual objc properties rather than our current implementation. They keep getting transferred to methods otherwise. I guess we could make a swift shim instead which wraps all the front facing classes into more swift like constructs,

rhyshort:
Being superseded by work to remove CDTMutableDocument revision

PR #184, opened 2015-07-29
Remove deprecated start method from CDTReplicator

rhyshort:
## What

Remove deprecated `start` method from CDTReplicator.

## Why

The main motivation behind removing it now, is because it will make it easier to use the replacement API for `start`, `startWithError:` because the method signature in Swift 2 becomes:

```swift
func start() throws
``` 
rather than 
```swift 
func startWithError() throws 
```

## Reviewer

reviewer @gadamc 
reviewer @tomblench

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - [NEW] Added query support for the `$size` operator.
 - [FIX] Fixed issue where at least one index had to be created before a query would execute.  You can now query for documents without the existence of any indexes.
+- [REMOVED] Removed deprecated method `start` on CDTReplicator, use `startWithError` instead.
 
 ## 0.17.1 (2015-06-24)
 
--- a/doc/replication.md
+++ b/doc/replication.md
@@ -67,7 +67,7 @@ devices. This web service needs to:
 
 Replications are set up in code on a device. Use `CDTPullReplication` and
 `CDTPushReplication` objects to create pre-configured `CDTReplicator` objects
-using a `CDTReplicatorFactory`. Then call `-start` on the `CDTReplicator` object
+using a `CDTReplicatorFactory`. Then call `-startWithError:` on the `CDTReplicator` object
 to start a replication. Each `CDTReplicator` object can be assigned a delegate
 to receive messages when replication completes or encounters an error.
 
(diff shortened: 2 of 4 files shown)


tomblench:
@rhyshort at the risk of being thought a pedant, please check your spelling before issuing PRs:
* depercated
* startWithErorr
Everyone seems to have 'an eye for detail' according to their CV...

tomblench:
@rhyshort I haven't been keeping track of changes here but how long has `start` been deprecated? Are we sure we want to make a breaking change just for the convenience of swift users?

rhyshort:
@tomblench it has been deprecated for over a year now ( git blame says it was deprecated in https://github.com/cloudant/CDTDatastore/commit/8ec1c27063de80395143a85b1e6962543fdfab64 ) so we aren't breaking anything just for convenience.

doc/replication.md
```diff
@@ -98,8 +98,12 @@ CDTReplicator *replicator = [replicatorFactory oneWay:pushReplication error:&err
 
 // Check replicator isn't nil, if so check error
 
-// Start the replication and wait for it to complete
-[replicator start];
+// Start the replication
+[replicator startWithError:&error];
+
+//check replication started
```
tomblench: Comments don't match the code, since the document is supposed to illustrate best practice, you should check and exit with a log message if there was a problem.

tomblench:
Need to update CHANGES

doc/replication.md
rhyshort: done in cfcfb00 however I didn't spell out the best way to handle the error as that would be down to their application's business logic.

rhyshort:
Changes updated in cfcfb00

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - [NEW] Added query support for the `$size` operator.
 - [FIX] Fixed issue where at least one index had to be created before a query would execute.  You can now query for documents without the existence of any indexes.
+- [REMOVED] Removed deprecated method `start` on CDTReplicator
```
tomblench: "... `startWithError` should be used instead"

doc/replication.md
```diff
@@ -98,11 +98,15 @@ CDTReplicator *replicator = [replicatorFactory oneWay:pushReplication error:&err
...
 // Check replicator isn't nil, if so check error
 
-// Start the replication and wait for it to complete
-[replicator start];
-while (replicator.isActive) {
-    [NSThread sleepForTimeInterval:1.0f];
-    NSLog(@" -> %@", [CDTReplicator stringForReplicatorState:replicator.state]);
+// Start the replication
+if ([replicator startWithError:&error]){
+	//handle error
```
tomblench: Mixture of tabs and spaces here - spaces should be preferred

rhyshort: fixed in 28ae90d

tomblench:
OK +1

gadamc:
+1

PR #187, opened 2015-08-19
Remove CDTMutableDocumentRevision

mikerhodes:
_What_

This commit removes the `CDTMutableDocumentRevision` class and fixes users of the class to use the newly mutable `CDTDocumentRevision` class.

_Why_

The aim is to simplify the modification of document revisions for developers using the library. This PR removes the API surface of `CDTMutableDocumentRevision`. The specific changes in this PR make things seem a little moot overall, as we don't lose any lines when updating a document, going from:

```objc
CDTMutableDocumentRevision *mutable = [rev mutableCopy];
mutable.body[@"foo"] = @"bar";
[datastore updateDocumentWithRevision:mutable error:nil];
```
to:
```objc
NSMutableDictionary *mutableBody = [rev.body mutableCopy];
mutableBody[@"foo"] = @"bar";
rev.body = mutableBody;
[datastore updateDocumentWithRevision:rev error:nil];
```

However, future work replaces the immutable `body` and `attachments` properties with mutable versions that can be properly watched for changes, which means we do end up with an overall shorter syntax in addition to fewer new classes for developers to learn about.

In the same way `CDTMutableDocumentRevision` simplified and improved things from how things were beforehand (you couldn't add more than one attachment at a time, for example), this change goes a step further in reducing the number of classes we expose and the hoops a developer needs to jump through.

_How_

There are a lot of changes, but the vast majority in the Tests are just fixups changing `CDTMutableDocumentRevision` to `CDTDocumentRevision` and the fact that 
`CDTDocumentRevision` can't have a settable `docId`.

In this version, revision body and attachments are not mutable, and users must make copies of them to change the content, then reassign the new dictionary to the property. This adds boilerplate code where
changing revisions. This is fixed in a later commit. The `private_body` and `private_attachments` are removed, so properties for `body` and `attachments` can be used.

In addition, some conflict resolving tests are broken at this point as a `CDTDocumentRevision` can't report whether it's been changed so the resolving code must assume all resolved revision objects are new [1]. This creates unnecessary new revisions if the user's resolver chooses an existing revision as a resolution. The broken tests are those which pass in an existing revision.

I'll try to cleanup out of date comments like [2] in a final pass -- let me know if you spot them and I'll add them to my list.

Documentation of new properties (e.g., [3] is added in a future commit. If things are confusing, I can move the doc comments here.

[1]: https://github.com/cloudant/CDTDatastore/compare/feature-50707-remove-mut-rev...50707-remove-class#diff-51ff5d1dc435df9958a5463953080658R61

[2]: https://github.com/cloudant/CDTDatastore/compare/feature-50707-remove-mut-rev...50707-remove-class#diff-09611c62c3ef826c53341a9070bc3d7dR746

[3]: https://github.com/cloudant/CDTDatastore/compare/feature-50707-remove-mut-rev...50707-remove-class#diff-cdfac8e485fc1dcc48f8e0fc73d9dcffR68

For reference, the full sequence of proposed changes is [here](https://github.com/cloudant/CDTDatastore/compare/50707-no-mutableCopy).

reviewer @tomblench 
reviewer @rhyshort


Classes/common/CDTDatastore.m
```diff
@@ -745,12 +744,7 @@ - (CDTDocumentRevision *)deleteDocumentFromRevision:(CDTDocumentRevision *)revis
     }
 
     // A mutable document revision stores the revId in other property
```
rhyshort: Comment should be removed

rhyshort: never mind, spotted that you know about this one already

rhyshort:
+1, as long as the failing tests are expected

tomblench:
OK +1

PR #189, opened 2015-08-20
50707 - Use mutable dictionaries in CDTDocumentRevision

mikerhodes:
_What_

Change CDTDocumentRevision's `body` and `attachment` properties into mutable dictionaries.

These dictionaries are wrapped with observer classes which report changes to the dictionaries and their deep contents, which allows us to report whether a particular document revision object has been changed since it was read from the database. This allows us to fix the conflict resolver tests, so they are re-enabled.

_Why_

The aim is to simplify the modification of document revisions for developers using the library. The PR allows developers to again modify the body and attachments, as they were able to on `CDTMutableDocumentRevision` objects.

So the previous way was this:

```objc
CDTDocumentRevision *rev = [datastore getDocumentWithId:id error:nil];
CDTMutableDocumentRevision *mutable = [rev mutableCopy];
mutable.body[@"foo"] = @"bar";
[datastore updateDocumentWithRevision:mutable error:nil];
```
with this PR, this is reduced to:
```objc
CDTDocumentRevision *rev = [datastore getDocumentWithId:id error:nil];
rev.body[@"foo"] = @"bar";
[datastore updateDocumentWithRevision:mutable error:nil];
```

_How_

First, the `body` and `attachment` properties are changed to be `NSMutableDictionary` objects. Internally, these are actually wrapped in NSMutableDictionary subclasses to enable the revision to observe changes as the users alters the body or attachments of the revision.

`isChanged` is a new property on `CDTDocumentRevision` and the new classes which wrap dictionaries and arrays to observe changes, `CDTChangedDictionary` and `CDTChangedArray` respectively. 

The class names  `CDTChangedDictionary` and `CDTChangedArray` seem a bit imperfect, but after thinking for a while I couldn't think of something better -- suggestions welcome.

A `CDTDocumentRevision` wraps the body and attachments passed into the `init...` method using a class method on `CDTChangedDictionary` which deep copies the contents of any dictionaries and attachments into new mutable dictionaries -- this helps prevent changes happening to the underlying containers which wouldn't be observed by the wrappers.

The `Changed*` classes report their changed status via both a property and a delegate. Parent containers register themselves as delegate on their children. The document revision object registers itself as delegate on the top level body and attachments dictionary. This way, changes to nested containers will cascade up to the document revision.

_Suggestions_

I think the easiest way to follow this PR -- if you'd rather I can break it up -- is:

1. CDTChangedArray -- this implements the primitive methods of NS(Mutable)Array, and fires off a message to its delegate.
1. CDTChangedDictionary -- similarly implements the primitive methods of NS(Mutable)Dictionary.
1. CDTChangedDictionary -- check the class methods for deep-wrapping dictionaries.
1. CDTDocumentRevision -- check the logic of isChanged, particularly whether it's always guaranteed to be set correctly on any change to body/attachments. The main hole here is non-wrapped objects, but I think for JSON, we only have primitive types which can't change without being replaced within the dict/array structure. I could be wrong.

For info on _primitive methods_, see https://mikeash.com/pyblog/friday-qa-2010-03-12-subclassing-class-clusters.html.

reviewer @rhyshort 
reviewer @tomblench

### DIFF
--- a/Classes/common/Utils/CDTChangedArray.h
+++ b/Classes/common/Utils/CDTChangedArray.h
@@ -1,0 +1,68 @@
+//
+//  CDTChangedArray.h
+//
+//
+//  Created by Michael Rhodes on 17/08/2015.
+//  Copyright (c) 2015 Cloudant. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+//  except in compliance with the License. You may obtain a copy of the License at
+//    http://www.apache.org/licenses/LICENSE-2.0
+//  Unless required by applicable law or agreed to in writing, software distributed under the
+//  License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+//  either express or implied. See the License for the specific language governing permissions
+//  and limitations under the License.
+//
+
+#import <Foundation/Foundation.h>
+
+#import "CDTChangedObserver.h"
+
+/**
+ A wrapper around NSMutableArray which has a changed flag that's
+ set when the content is changed.
+
+ Changes are tracked very simply, isChanged is set to true when any of the
+ following are called:
+
+  - insertObject:atIndex:
+  - removeObjectAtIndex:
+  - addObject:
+  - removeLastObject
+  - replaceObjectAtIndex:withObject:
+
+ The changed flag will only be set for modifications through this class, not
+ when the underlying array changes itself, therefore it's important
+ not to let the contained array "escape" from this class if all changes
+ need to be tracked.
+
+ This class implements `CDTChangedObserver` so that it can bubble changes up
+ through to its parent containers within a deep structure that has been
+ wrapped with CDTChangedDictionary's `dictionaryWrappingContents:`. That
+ method copies the contents of all containers within the passed dictionary
+ into new CDTChanged* containers, assigning each wrapped container its parent
+ container as delegate.
+ */
+@interface CDTChangedArray : NSMutableArray <CDTChangedObserver>
+
+/**
+ Create an empty array.
+ */
++ (CDTChangedArray *)emptyArray;
+
+/**
+ Will be notified if this object is changed.
+ */
+@property (nonatomic, weak) NSObject<CDTChangedObserver> *delegate;
+
+/**
+ Init with array. This constructor must be used.
+ */
+- (instancetype)initWithMutableArray:(nonnull NSMutableArray *)array;
+
+/**
+ Set to YES if this dictionary has been modified.
+ */
+@property (nonatomic, getter=isChanged) bool changed;
+
+@end
--- a/Tests/Tests/Utils/CDTChangedDictionaryJSONWrappingTests.m
+++ b/Tests/Tests/Utils/CDTChangedDictionaryJSONWrappingTests.m
@@ -1,0 +1,88 @@
+//
+//  CDTChangedDictionaryJSONWrappingTests.m
+//  Tests
+//
+//  Created by Michael Rhodes on 17/08/2015.
+//  Copyright (c) 2015 Cloudant. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+//  except in compliance with the License. You may obtain a copy of the License at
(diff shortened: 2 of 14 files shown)


rhyshort:
General comment, new files need the licence header

rhyshort:
+1

Classes/common/Utils/CDTChangedArray.h
```diff
@@ -0,0 +1,61 @@
...
+  - addObject:
+  - removeLastObject
+  - replaceObjectAtIndex:withObject:
+
+ The changed flag will only be set for modifications through this class, not
+ when the underlying array changes itself, therefore it's important
+ not to let the contained array "escape" from this class if all changes
+ need to be tracked.
+ */
+@interface CDTChangedArray : NSMutableArray <CDTChangedObserver>
```
tomblench: Why does the array itself have to adopt this protocol? The interested party will set the appropriate delegate object and if they don't want a delegate they can access `isChanged` directly.

Classes/common/Utils/CDTChangedArray.h
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ Init with array. This constructor must be used.
+ */
+- (instancetype)initWithMutableArray:(nonnull NSMutableArray *)array;
+
+/**
+ Set to YES if this dictionary has been modified.
+ */
+@property (nonatomic) bool isChanged;
```
tomblench: `isChanged` sounds awkward although `isXXX` is standard for booleans. Does `hasChanged` sound too much like a delegate callback? Or `wasChanged`, or `didChange`?

Tests/Tests/Utils/CDTChangedDictionaryJSONWrappingTests.m
```diff
@@ -0,0 +1,88 @@
...
+#import <CDTChangedDictionary.h>
+#import <CDTChangedArray.h>
+
+@interface CDTChangedDictionaryJSONWrappingTests : XCTestCase
+
+@property (nonatomic, strong) CDTChangedDictionary *dictionary;
+
+@end
+
+@implementation CDTChangedDictionaryJSONWrappingTests
```
tomblench: would it be useful to test the `changed` true → false transition which occurs when calling `dictionaryWrappingContents`?

Classes/common/Utils/CDTChangedArray.h
mikerhodes: The array and dictionary are delegates of the objects they contain.

Classes/common/Utils/CDTChangedArray.h
mikerhodes: Hmm. I think `isChanged` is standard; one is asking about the current state?

Tests/Tests/Utils/CDTChangedDictionaryJSONWrappingTests.m
mikerhodes: That should never be externally visible -- the newly created dictionary doesn't leave the `dictionaryWrappingContents` scope before `changed` is set to `true`.

Tests/Tests/Utils/CDTChangedDictionaryJSONWrappingTests.m
```diff
@@ -0,0 +1,88 @@
...
+
+    NSDictionary *dictionary = @{
+        @"dict" :
+            @{@"array" : @[ @"one", @"two" ], @"dict" : @{@"one" : @"two"}, @"hello" : @"world"},
+
+        @"array" : @[ @[ @"foo", @YES ], @{@"foo" : @YES}, @"two" ]
+    };
+
+    self.dictionary = [CDTChangedDictionary dictionaryWrappingContents:dictionary];
+}
```
tomblench: Would it be useful to have `XCTAssertFalse(self.dictionary.isChanged);` either at the end of setup or at the start of each test?

mikerhodes: I test this here https://github.com/cloudant/CDTDatastore/pull/189/files#diff-7de81db5bbf96ef38818303145945047R44

tomblench:
OK +1

mikerhodes:
Squashing and merging.

PR #190, opened 2015-08-27
Implement request Retries

rhyshort:
## What
Added the ability to filter responses and cause requests to be retried.

## How
Implement request retires in CDTURLSessionTask and add protocol method
to intercept responses to CDTHTTPInterceptor

## Testing
CDTURLSessionTests validate that the response and request filters successfully run.

## Reviewers

reviewer @tomblench 
reviewer @mikerhodes

PR #191, opened 2015-09-02
Expose HTTP interceptors in public API

rhyshort:
## What

Expose HTTP interceptors in the public API via the
`CDTAbstractReplication` object.

## How

The following APIs have been added to the CDTAbstractReplication object:
`-addInterceptor:`
`-addInterceptors:`
`-clearInterceptors:`

## Reviewers

reviewer @tomblench 
reviewer @mikerhodes

### DIFF
--- a/Classes/common/CDTReplicator/CDTAbstractReplication.h
+++ b/Classes/common/CDTReplicator/CDTAbstractReplication.h
@@ -14,6 +14,7 @@
 //  and limitations under the License.
 
 #import <Foundation/Foundation.h>
+#import "CDTHTTPInterceptor.h"
 
 @class CDTDatastore;
 
@@ -126,6 +127,16 @@ typedef NS_ENUM(NSInteger, CDTReplicationErrors) {
 
 */
 @property (nonatomic, copy) NSDictionary* optionalHeaders;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@property (nonnull, nonatomic, readonly, strong) NSArray* httpInterceptors;
+
+- (void)addInterceptor:(nonnull NSObject<CDTHTTPInterceptor>*)interceptor;
+- (void)addInterceptors:(nonnull NSArray*)interceptors;
+- (void)clearInterceptors;
+
+NS_ASSUME_NONNULL_END
 
 /**
  Returns the default "User-Agent" header value used in HTTP requests made during replication.
--- a/Classes/common/CDTReplicator/CDTAbstractReplication.m
+++ b/Classes/common/CDTReplicator/CDTAbstractReplication.m
@@ -19,6 +19,14 @@
 
 NSString *const CDTReplicationErrorDomain = @"CDTReplicationErrorDomain";
 
+@interface CDTAbstractReplication ()
+
+NS_ASSUME_NONNULL_BEGIN
+@property (nonnull, nonatomic, readwrite, strong) NSArray *httpInterceptors;
+NS_ASSUME_NONNULL_END
+
+@end
+
 @implementation CDTAbstractReplication
 
 + (NSString *)defaultUserAgentHTTPHeader { return [TDRemoteRequest userAgentHeader]; }
@@ -28,10 +36,46 @@ NSString *const CDTReplicationErrorDomain = @"CDTReplicationErrorDomain";
     CDTAbstractReplication *copy = [[[self class] allocWithZone:zone] init];
     if (copy) {
         copy.optionalHeaders = self.optionalHeaders;
+        copy.httpInterceptors = [self.httpInterceptors copyWithZone:zone];
     }
 
     return copy;
 }
+
+NS_ASSUME_NONNULL_BEGIN
+
+- (instancetype)init
+{
+    self = [super init];
+    if (self) {
+        _httpInterceptors = @[];
+    }
+    return self;
+}
+
+/**
+ * This method is a convience method and is the same as calling
+ * -addinterceptors: with a single element array.
+ */
+- (void)addInterceptor:(NSObject<CDTHTTPInterceptor> *)interceptor
+{
+    [self addInterceptors:@[ interceptor ]];
+}
+
+/**
+ * Appends the interceptors in the array to the list of
+ * interceptors to run for each request made to the
+ * server.
+ *
+ * @param interceptors the interceptors to append to the list
+ **/
+- (void)addInterceptors:(NSArray *)interceptors
+{
+    self.httpInterceptors = [self.httpInterceptors arrayByAddingObjectsFromArray:interceptors];
+}
+
+- (void)clearInterceptors { self.httpInterceptors = @[]; }
+NS_ASSUME_NONNULL_END
 
 /**
  This method sets all of the common replication parameters. The subclasses,
@@ -123,6 +167,7 @@ NSString *const CDTReplicationErrorDomain = @"CDTReplicationErrorDomain";
 
         doc[@"headers"] = self.optionalHeaders;
     }
+    doc[@"interceptors"] = self.httpInterceptors;
 
     return [NSDictionary dictionaryWithDictionary:doc];
 }
--- a/ReplicationAcceptance/ReplicationAcceptance/ReplicationAcceptance.m
+++ b/ReplicationAcceptance/ReplicationAcceptance/ReplicationAcceptance.m
@@ -38,6 +38,7 @@
 #import "ChangeTrackerDelegate.h"
 #import "ChangeTrackerNSURLProtocolTimedOut.h"
 #import "TDInternal.h"
+#import "CDTHTTPInterceptor.h"
 
 @interface ReplicationAcceptance () 
 
@@ -113,7 +114,7 @@ static NSUInteger largeRevTreeSize = 1500;
     
     pull.filter = filterName;
     pull.filterParams = params;
-    
+
     NSError *error;
     CDTReplicator *replicator =  [self.replicatorFactory oneWay:pull error:&error];
     XCTAssertNil(error, @"%@",error);
@@ -128,10 +129,9 @@ static NSUInteger largeRevTreeSize = 1500;
         [NSThread sleepForTimeInterval:1.0f];
         NSLog(@" -> %@", [CDTReplicator stringForReplicatorState:replicator.state]);
     }
-    
+
     return replicator;
 }
-
 
 /**
  Create a new replicator, and wait for replication from the local database to complete.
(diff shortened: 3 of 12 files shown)


Classes/common/CDTReplicator/CDTAbstractReplication.h
```diff
@@ -127,6 +128,15 @@ typedef NS_ENUM(NSInteger, CDTReplicationErrors) {
 */
 @property (nonatomic, copy) NSDictionary* optionalHeaders;
 
+NS_ASSUME_NONNULL_BEGIN
+
+@property (nonnull, nonatomic, readonly, strong) NSArray* httpInterceptors;
+
+- (void)addInterceptor:(nonnull NSObject<CDTHTTPInterceptor>*)interceptor;
+- (void)clearInterceptors;
```
tomblench: `addInterceptors` should be declared here

Classes/common/CDTReplicator/CDTAbstractReplication.m
```diff
@@ -28,11 +36,47 @@ - (instancetype)copyWithZone:(NSZone *)zone
...
+    return self;
+}
+
+/**
+ * This method is a convience method and is the same as calling
+ * -addinterceptors: with a single element array.
+ */
+- (void)addInterceptor:(NSObject<CDTHTTPInterceptor> *)interceptor
+{
+    [self addInterceptors:@[ interceptor ]];
```
tomblench: is the spacing correct for array literals here?

ReplicationAcceptance/ReplicationAcceptance/ReplicationAcceptance.m
```diff
@@ -128,7 +148,9 @@ -(CDTReplicator *) pullFromRemoteWithFilter:(NSString*)filterName params:(NSDict
         [NSThread sleepForTimeInterval:1.0f];
         NSLog(@" -> %@", [CDTReplicator stringForReplicatorState:replicator.state]);
     }
-    
+
+    XCTAssertGreaterThanOrEqual(interceptor.timesRequestIntercepted, 1);
+
```
tomblench: re-run replication after clearing the interceptors and check the count hasn't gone up?

Classes/common/CDTReplicator/CDTAbstractReplication.m
rhyshort: `clang-format` thinks so

tomblench: fair enough, looks nasty to me though!

tomblench:
I'm generally +1 once my comments are addressed. One question, why does this have a `clearInterceptors` but sync-android doesn't? This seems inconsistent.

ReplicationAcceptance/ReplicationAcceptance/ReplicationAcceptance.m
rhyshort: done in 0e908e3

rhyshort:
The API isn't perfect, and there is some more work needed around what the replication API should look like, now that sync-android has a much simpler API with the `ReplicatorBuilder`. Its definitely something we need to sit down and discuss to see what we can do to remove the whole replication -> factory -> replicator dance.

Classes/common/CDTReplicator/CDTAbstractReplication.h
```diff
@@ -127,6 +128,15 @@ typedef NS_ENUM(NSInteger, CDTReplicationErrors) {
 */
 @property (nonatomic, copy) NSDictionary* optionalHeaders;
 
+NS_ASSUME_NONNULL_BEGIN
+
+@property (nonnull, nonatomic, readonly, strong) NSArray* httpInterceptors;
```
mikerhodes: Remove the `nonnull` around here as we have assume nonnull? Throughout.

Classes/common/CDTReplicator/CDTAbstractReplication.m
```diff
@@ -19,6 +19,14 @@
 
 NSString *const CDTReplicationErrorDomain = @"CDTReplicationErrorDomain";
 
+@interface CDTAbstractReplication ()
+
+NS_ASSUME_NONNULL_BEGIN
+@property (nonnull, nonatomic, readwrite, strong) NSArray *httpInterceptors;
```
mikerhodes: Why not just use the `addInterceptors:` method rather than making this readwrite, looks like you just use this in `copy`?

Classes/common/CDTReplicator/CDTAbstractReplication.m
```diff
@@ -123,6 +167,7 @@ - (NSDictionary *)dictionaryForReplicatorDocument:(NSError *__autoreleasing *)er
 
         doc[@"headers"] = self.optionalHeaders;
     }
+    doc[@"interceptors"] = self.httpInterceptors;
```
mikerhodes: Did we still need this?

rhyshort: Yes :disappointed: it's so we can pass the interceptors to the creation method for the TDReplicator

Classes/common/CDTReplicator/CDTAbstractReplication.m
rhyshort: Code style says to prefer [accessing properties over iVars](https://github.com/cloudant/CDTDatastore/blob/master/doc/style-guide.md#properties), and we use the `readwrite` nature [here](https://github.com/cloudant/CDTDatastore/pull/191/files#diff-c7f6d1c619815916886c1d6b3abc1be2R74). I could use the `addInterceptors` method, but I thought it would make sense to be consistent and access the property directly.

Classes/common/CDTReplicator/CDTAbstractReplication.h
rhyshort: We can remove it, however I think its beneficial to have it spelt out until nullability annotations is norm in Objective-c code.

Classes/common/CDTReplicator/CDTAbstractReplication.h
rhyshort: done in 6699fab

Classes/common/CDTReplicator/CDTAbstractReplication.m
mikerhodes: Accepted.

Classes/common/CDTReplicator/CDTAbstractReplication.m
mikerhodes: Sure, I reread this and saw the way it's passed around.

mikerhodes:
@tomblench is this the place for checking the multiple interceptors, as requested https://github.com/cloudant/CDTDatastore/pull/190#issuecomment-137498714?

Classes/common/CDTReplicator/CDTAbstractReplication.h
mikerhodes: Accepted.

ReplicationAcceptance/ReplicationAcceptance/ReplicationAcceptance.m
```diff
@@ -128,7 +148,28 @@ -(CDTReplicator *) pullFromRemoteWithFilter:(NSString*)filterName params:(NSDict
...
+        XCTFail(@"CDTReplicator -startWithError: %@", error);
+    }
+
+    while (replicator.isActive) {
+        [NSThread sleepForTimeInterval:1.0f];
+        NSLog(@" -> %@", [CDTReplicator stringForReplicatorState:replicator.state]);
+    }
+
+    XCTAssertEqual(timesRequestIntercepted, interceptor.timesRequestIntercepted);
+
```
mikerhodes: This is an inappropriate addition of functionality to this method as I don't think adding overhead to every test is wise. In addition, it doesn't appear to be testing anything which isn't tested in your unit tests?

tomblench:
@mikerhodes yes this would an appropriate PR for that.

ReplicationAcceptance/ReplicationAcceptance/ReplicationAcceptance.m
rhyshort: I'm not exactly sure, which function you are referring to, is it the whole set of changes? or the re run the replicator change?

mikerhodes: Yes, 151 to 172 -- not sure why we are doing this to every thing done in ReplicationAcceptance, what it adds?

mikerhodes: tbh, I'd remove this from replication acceptance for this PR. I don't think it's necessary -- unless it caught some bugs?

rhyshort: It adds a check to make sure the interceptors get passed down, it was a path of least resistance, going to revert, and we can work on it in a different PR, I already have a test that makes sure the interceptors are passed down to the change tracker

rhyshort: removed in commit d7d78bd

Tests/Tests/CDTReplicationTests.m
```diff
@@ -33,6 +64,35 @@ @interface CDTReplicationTests : CloudantSyncTests
 
 @implementation CDTReplicationTests
 
+- (void)testFiltersWithChangesFeed
+{
+    NSError *error;
+    NSString *remoteUrl = @"https://example.com";
```
tomblench: Some comments here would be helpful - what we discussed about how this has to be a real URL because of the reachability tests, but we don't care that it there's not a real replicator on the end of this

Tests/Tests/CDTReplicationTests.m
```diff
@@ -26,6 +26,37 @@
...
+        _changesFeedRequestMade = NO;
+    }
+    return self;
+}
+
+- (CDTHTTPInterceptorContext *)interceptRequestInContext:(CDTHTTPInterceptorContext *)context
+{
+    NSURL *url = context.request.URL;
+
+    if ([[url path] containsString:@"/_changes"]) {
```
tomblench: Add a comment to show that this tests the fact that the interceptors are being run on the changes tracker as that is the only code path which accesses the `_changes` URL

mikerhodes:
+1 after Tom's comments addressed.

Tests/Tests/CDTReplicationTests.m
rhyshort: added in d2de76b

Tests/Tests/CDTReplicationTests.m
rhyshort: added in d2de76b

Issue #194, opened 2015-09-08
Cannot build as a framework


jhs:
(I am working from one week of total iOS, and XCode experience, so please bear with me!)

I think this project cannot build as a framework when used by a Swift project (at least with XCode 7 beta 6). The main problem is "Include of non-modular header inside framework module".

From my limited research, I gather that this issue is somewhat new (as apparently frameworks are not the only way to organize subprojects); but it is becoming more urgent, since Swift projects cannot work around it. (One can disable this compiler error; however, the disabling is not honored in Swift projects, whether by bug or design.) A clear description of the problem is here: http://inessential.com/2015/03/16/swift_blocker

In any case, it sounds like the situation XCode caught is indeed problematic and should be fixed. (My understanding is that, .h files should not import system-wide .h files--.m files should; something about a naming conflict as multiple frameworks compile together). Here is your dependency, CocoaLumberjack [struggling with the same issue][lumberjack-issue].

[lumberjack-issue]: https://github.com/CocoaLumberjack/CocoaLumberjack/issues/544

So, modulo my naïve understanding, to be complete (except of course as a NOFIX), CDTDatastore needs to fix its own issues and then also upgrade to all dependencies when they fix theirs. This issue only applies to this project's specific build failure.

AFAICT, there is only one problem in the codebase:

    /Users/jhs/src/cloudant/CDTDatastore/Classes/common/Encryption/Attachments/CDTBlobEncryptedDataConstants.h:20:9: Include of non-modular header inside framework module 'CDTDatastore.CDTBlobEncryptedDataConstants'

rhyshort:
Unfortunately FMDB, also lacks support for frameworks, although that is being worked on see: https://github.com/ccgus/fmdb/issues/309 .There is a way around the frameworks issue. You need to add a bridging header to your project and use that to import CDTDatastore for use in your swift code.

The easiest way I've found to create a bridging header, is to create a new cocoa class from in the new file wizard, changing the language to objective-c and selecting the option to create the bridging header. In the bridging header you then add the line `#import "CloudantSync.h"`. You'll have access to all the CDTDatastore classes from your swift code automatically, no need for `import` statements in your swift files.

jhs:
Thanks @rhyshort! 

Yes, my larger mission was, in fact, to work without the bridging header. That seems to be the future, AFAIK. CocoaPods tells Swift users to `use_frameworks!` and the GitHub issues on all these projects indicates that we are in a transitional period.

[...]

PR #195, opened 2015-09-08
Use CocoaLumberjack v2.0.0 since 2.0.1 cannot build as a framework

jhs:
See https://github.com/CocoaLumberjack/CocoaLumberjack/issues/544

and also the same issue affecting this project itself: #194


rhyshort:
Thanks for bringing this to our attention. However currently due to https://github.com/ccgus/fmdb/issues/309 we are unable to build as a framework. We'll take a look at pinning CocoaLumberjack  to a version when we can build as a framework.

PR #198, opened 2015-09-09
Rename CDTDocumentRevision's -mutableCopy to -copy

mikerhodes:
_What & Why_

CDTDocumentRevision is now mutable, so having a method named -mutableCopy didn't make sense -- the simple -copy method is more appropriate.

_How_

- 73315e9
    - Rename `-mutableCopy` to `-copy` and fix method calls to the previous method. Content of method doesn't change.
    - Also rename the `xWrappingY` methods to `copy` to be more consistent.
- f90db3d
    - This caused some tests to fail, so they are ignored -- a future commit on the feature branch will fix.

reviewer @rhyshort

### DIFF
--- a/Classes/common/CDTDocumentRevision.h
+++ b/Classes/common/CDTDocumentRevision.h
@@ -86,10 +86,14 @@
 - (NSData *)documentAsDataError:(NSError *__autoreleasing *)error;
 
 /**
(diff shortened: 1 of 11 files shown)


rhyshort:
+1

tomblench:
+1

PR #201, opened 2015-09-11
51928 Add isFullRevision property to CDTDocumentRevision and subclasses

mikerhodes:
_Why_

Some operations, for example querying with a field filter, produce document revision objects which are not complete representations of the revision. In order to avoid data loss in applications, these revisions should not be allowed to be saved by the datastore.

_How_

https://github.com/cloudant/CDTDatastore/commit/0c722a5eeaa3e1d171284cc5029929e922cfcd87

Adds `isFullRevision` property to `CDTDocumentRevision`. This allows partial revision objects, such as `CDTQProjectedDocumentRevision`, to be recognised when a developer attempts to write one to the database. `-updateDocumentFromRevision:error:` uses `isFullRevision` to reject updates where the property is `NO`.

- `CDTDocumentRevision` sets the property to `YES` so it can be saved.
- `CDTProjectedDocumentRevision` sets the property to `NO` so it cannot be saved.
- `-updateDocumentFromRevision:error:` first checks the property before saving.

https://github.com/cloudant/CDTDatastore/commit/f6c8250797d9dded092f5dabbf486e8d9b9f6044

Re-enables tests which needed this functionality.

### DIFF
--- a/Classes/common/CDTDocumentRevision.h
+++ b/Classes/common/CDTDocumentRevision.h
@@ -19,7 +19,6 @@
 #import "CDTChangedObserver.h"
 
 @class TD_RevisionList;
-@class CDTDocumentRevision;
 
 /**
  * Represents a single revision of a document in a datastore.
@@ -40,6 +39,15 @@
 @property (nonatomic, strong) NSMutableDictionary *attachments;
 
 @property (nonatomic, getter=isChanged) bool changed;
+
+/**
+ Indicates if this document revision contains all fields and attachments. If NO, datastore
+ will refuse to save.
+
+ This allows for document projections to conform to the CDTDocumentRevision API without
+ risking their being saved to the datastore.
+ */
+- (BOOL)isFullRevision;
 
 - (id)initWithDocId:(NSString *)docId
          revisionId:(NSString *)revId
--- a/Classes/common/CDTDatastore.m
+++ b/Classes/common/CDTDatastore.m
@@ -599,9 +599,27 @@ NSString *const CDTDatastoreChangeNotification = @"CDTDatastoreChangeNotificatio
 - (CDTDocumentRevision *)updateDocumentFromRevision:(CDTDocumentRevision *)revision
                                               error:(NSError *__autoreleasing *)error
 {
+    if (!revision.isFullRevision) {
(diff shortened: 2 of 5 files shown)


Classes/common/CDTDocumentRevision.h
```diff
@@ -19,12 +19,14 @@
...
-@class CDTDocumentRevision;
 
 /**
  * Represents a single revision of a document in a datastore.
  */
-@interface CDTDocumentRevision : NSObject <CDTChangedObserver>
+@interface CDTDocumentRevision : NSObject <CDTChangedObserver> {
+   @protected
+    bool _isFullRevision;
+}
```
mikerhodes: The `@protected` is required so sub-classes can set the property at `-init` time.

tomblench:
Two thoughts on this:
- Could you swap the logic round and call the property `isProjectedRevision`? _or_
- Could you make better use of the class hierarchy, perhaps override `updateDocumentFromRevision` in `CDTQProjectedDocumentRevision` to throw the error there instead?

I guess it partly depends on whether you can envisage any other situations where you would encounter a "non-full/partial" revision.

mikerhodes:
> - Could you swap the logic round and call the property isProjectedRevision?

From `updateDocumentFromRevision`'s perspective, it doesn't care why the revision isn't a complete revision. It's acting on the fact that the revision isn't a full one, not that it's projected. As you mentioned below, this property can be reused by other classes too, if need be.

> - Could you make better use of the class hierarchy, perhaps override updateDocumentFromRevision in CDTQProjectedDocumentRevision to throw the error there instead?

I think you're confusing `CDTDatastore` with `CDTDocumentRevision` here?

tomblench:
Right, i got a bit confused about the classes involved. How about an `isFullRevision` method which returns `YES` for the non-projected version and `NO` for the projected one?

mikerhodes:
@tomblench 
> Right, i got a bit confused about the classes involved. How about an isFullRevision method which returns YES for the non-projected version and NO for the projected one?

Isn't this basically what we have?

tomblench:
+1

rhyshort:
+1

PR #203, opened 2015-09-16
Clean up Rakefile

mikerhodes:
_What_

Firstly, adding methods to the Rakefile to avoid code duplication.

Secondly, making the dependents of the `:travis` task into true dependents rather than running a sub-process with another rake instance inside. In addition, OS X tests are run first rather than iOS. This is a fail-fast philosophy.

_Why_

Rakefile was becoming unwieldy. Many tasks had almost the same code, meaning any changes that affected many tasks were error prone (e.g., change iPhone model that tests run on).

The changes to the `travis` task are to make failures in the build cause failures on travis more quickly. Using true dependents means that the rake task will fail after the first XCTest run fails. In combination with running the OS X tests before the iOS tests, this reduces failing test runs to 6 minutes rather than 20 minutes.

My thought here was that if we fail on OS X, we're likely to fail iOS and the encryption tests. So lets run heavier tests only if the faster test runs work out.

BugzID: 52223

reviewer @tomblench

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -63,19 +63,11 @@ task :testencryptionosx do
 end
 
 desc "Run tests for all platforms"
-task :test do
-  sh "rake testios"
-  sh "rake testosx"
-  sh "rake testencryptionios"
-  sh "rake testencryptionosx"
+task :test => [:testosx, :testios, :testencryptionosx, :testencryptionios] do
 end
 
 desc "Task for travis"
-task :travis do
-  sh "rake testios"
-  sh "rake testosx"
-  sh "rake testencryptionios"
-  sh "rake testencryptionosx"
+task :travis => [:test] do
   sh "pod lib lint --allow-warnings"
 end
 
--- a/Rakefile
+++ b/Rakefile
@@ -1,46 +1,33 @@
-# Runs `build` target for workspace/scheme/destination
-def run_build(workspace, scheme, destination)
-  # build using xcpretty as otherwise it's very verbose when running tests
-  $ios_success = system("xcodebuild -workspace #{workspace} -scheme '#{scheme}' -destination '#{destination}' build | xcpretty; exit ${PIPESTATUS[0]}")
-  return $ios_success
+#
+#  Primary tasks
+#
+
+desc "Run tests for all platforms"
+task :test => [:testosx, :testios, :testencryptionosx, :testencryptionios] do
 end
 
-# Runs `test` target for workspace/scheme/destination
-def run_tests(workspace, scheme, destination)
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -1,167 +1,127 @@
...
 desc "Build docs and install to Xcode"
 task :docs do
   system("appledoc --keep-intermediate-files --project-name CDTDatastore --project-company Cloudant -o build/docs --company-id com.cloudant -i Classes/vendor -i Classes/common/touchdb Classes/")
 end
+
+#
+#  Helper methods and defines
+#
+
+# Workspaces
```
tomblench: Can we put these 'constants' at the top of the file? I think they make more sense there

Rakefile
```diff
@@ -1,167 +1,127 @@
...
+REPLICATION_ACCEPTANCE_OSX = 'RA_Tests_OSX'
+REPLICATION_ACCEPTANCE_ENCRYPTED_IOS = 'RA_EncryptionTests'
+REPLICATION_ACCEPTANCE_ENCRYPTED_OSX = 'RA_EncryptionTests_OSX'
+
+# Destinations
+IPHONE_DEST = 'platform=iOS Simulator,OS=latest,name=iPhone 4S'
+OSX_DEST = 'platform=OS X'
+
+# Runs `build` target for workspace/scheme/destination
+def run_build(workspace, scheme, destination)
```
tomblench: I can see you've removed the `"** Build failed"` and `** All tests executed successfully` messages. Was this to make the output less verbose? I think at least the test failure and build failure should be consistent with each other and have similar messages (including the red escape code which might be better defined as a constant)

Rakefile
mikerhodes: Done in 44ee7548f67a750cf90f67e06fc923ec76e152e1.

Rakefile
mikerhodes: Failure messages are here: https://github.com/cloudant/CDTDatastore/pull/203/files#diff-52c976fc38ed2b4e3b1192f8a8e24cffR124. I updated the failure methods to include the workspace and scheme for more usefulness when using grep.

I'd not thought of the red failure string there -- do you think I should add it?

tomblench: OK, I see. In that case you can remove the message here: https://github.com/cloudant/CDTDatastore/pull/203/files#diff-52c976fc38ed2b4e3b1192f8a8e24cffR119

I would be in favour of making the failure messages red. Most tools can interpret the terminal escapes correctly and they don't harm readability too much for those things which can't interpret them.

rhyshort:
+1

tomblench:
+1

PR #205, opened 2015-09-30
Set HTTP Method for `_session` requests

rhyshort:
## What
Fix issue where session authentication would fail.

## How

Set the HTTP method for `_session` requests made by the `CDTSessionCookieInterceptor` 

## Testing

The OHHTTPStub now checks for the correct HTTP method and delivers a response based on it. For `GET` requests it will return 200 with an empty body, and no headers, for `POST` the cookie header will be set and a 200 response returned.

## Reviewers

reviewer @tomblench 
reviewer @mikerhodes

### DIFF
--- a/Tests/Tests/CDTSessionCookieInterceptorTests.m
+++ b/Tests/Tests/CDTSessionCookieInterceptorTests.m
@@ -49,18 +49,29 @@ static const NSString *testCookieHeaderValue =
       return [[request.URL host] isEqualToString:@"username.cloudant.com"];
     }
         withStubResponse:^OHHTTPStubsResponse *__nonnull(NSURLRequest *__nonnull request) {
-          return [OHHTTPStubsResponse
-              responseWithJSONObject:@{
-                  @"ok" : @(YES),
-                  @"name" : @"username",
-                  @"roles" : @[ @"_admin" ]
-              }
-                          statusCode:200
-                             headers:@{
-                                 @"Set-Cookie" :
-                                     [NSString stringWithFormat:@"%@; Version=1; Path=/; HttpOnly",
-                                                                testCookieHeaderValue]
-                             }];
+
+          if ([request.HTTPMethod isEqualToString:@"POST"]) {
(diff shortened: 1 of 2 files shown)


tomblench:
+1

Tests/Tests/CDTSessionCookieInterceptorTests.m
```diff
@@ -49,18 +49,23 @@ - (void)setUp
...
+                      @"roles" : @[ @"_admin" ]
+                  }
+                              statusCode:200
+                                 headers:@{
+                                     @"Set-Cookie" : [NSString
+                                         stringWithFormat:@"%@; Version=1; Path=/; HttpOnly",
+                                                          testCookieHeaderValue]
+                                 }];
+          } else {
+              return [OHHTTPStubsResponse responseWithJSONObject:@{} statusCode:200 headers:@{}];
```
mikerhodes: I'd check whether `200` is really what Couch sends here.

mikerhodes: Or can you just somehow fail if the exact expected request isn't met -- you probably want to check for the form-encoded body too?

rhyshort: Couch returns `200` for a post to session see: http://docs.couchdb.org/docs/couchdb/en/stable/api/server/authn.html#session.

I can't check the body to make sure it is form encoded, so far my attempts to read HTTPBodyStream always fail. The only thing I can check is the content type

mikerhodes: This is in the `else` block so it's not a POST to session. It looks like it'll return some info about the user whose details are in the request with a `200` code from the docs.

rhyshort: I think I understand what you mean, and I think I implemented it in 24bb3b5

mikerhodes:
+1

PR #207, opened 2015-10-14
Avoid assertion in allDocsQuery:options: when a docId isn't found.

sdrew:
When using `allDocsQuery:options:` with an array of docIds, a missing document will cause an 'Invalid parameter not satisfying: properties' assertion in `[TD_Body initWithProperties:]`.

### DIFF
--- a/Classes/common/CDTDatastore.m
+++ b/Classes/common/CDTDatastore.m
@@ -350,6 +350,12 @@ NSString *const CDTDatastoreChangeNotification = @"CDTDatastoreChangeNotificatio
         NSString *docId = row[@"id"];
 
         NSString *revId = row[@"value"][@"rev"];
+        
+        // If a document isn't found, docId and revId will be null, and row will
+        // contain an @"error" key.
+        if (docId == nil && revId == nil && [row[@"error"] isEqualToString:@"not_found"]) {
+            continue;
+        }
 
         // deleted field only present in deleted documents, but to be safe we use
         // the fact that (BOOL)[nil -boolValue] is false
--- a/Tests/Tests/CDTDatastoreQueryTests.m
+++ b/Tests/Tests/CDTDatastoreQueryTests.m
@@ -89,6 +89,13 @@ SpecBegin(CDTDatastoreQuery) describe(@"When using datastore query", ^{
         expect(results).toNot.beNil();
         expect([results.documentIds count]).to.equal(3);
     });
+    
+    it(@"can find documents and ignore non-existent documents", ^{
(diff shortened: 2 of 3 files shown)


mikerhodes:
Could you add a failing test so I can be sure I'm properly understanding the issue at hand?

Also, you'll need to complete and send in a CLA before I can accept this. The process is reasonably painless and can be done electronically, see https://github.com/cloudant/CDTDatastore/blob/master/CONTRIBUTING.md#contributor-license-agreement.

sdrew:
Apologies. I should have checked the requirements beforehand. I've added the tests as requested and have signed the CLA, which I will be emailing shortly.

The first test is in `CDTDatastoreQueryTests.m` under `can find documents and ignore non-existent documents`. It uses `[CDTDatastore find:]` with the following query:

````objective-c
@{ @"_id" : @{ @"$in" : @[@"mike12", @"mike34", @"mike72", @"mike-not-found"] } }
````

This method works as expected, fetching 3 documents and ignoring the non-existent document. It doesn't require the change to `CDTDatastore` in this pull request.

The other test is in `DatastoreCRUD.m` under `testGetDocumentsWithIds_NonExistentDocument`. It attempts to use `[CDTDatastore getDocumentsWithIds:]` with 4 existing docIds and a non-existent one.

````objective-c
@[docIds[5], docIds[7], docIds[12], docIds[170], @"i_do_not_exist"];
````

Without the change in this pull-request, this test will fail with an assertion in the `[TD_Body initWithProperties:]` call at line **364** of `CDTDatastore.m`. This is because `[TD_Database getDocsWithIds:options:]` will return a dictionary with the following format when the docId doesn't exist:

````objective-c
@{ @"key": docID, @"error": @"not_found" }
````

To avoid any unintended side-effects, this pull request only checks for the specific case of `docId` and `revId` being nil, as well as `row` containing an `@"error"` key with the value `@"not_found"`. Once applied, `[CDTDatastore getDocumentsWithIds:]` will ignore non-existent docIds and return only the found documents, instead of throwing an assertion.

rhyshort:
@sdrew thanks for adding those tests, however we have yet to receive your signed CLA.

sdrew:
@rhyshort @mikerhodes It was sent on October 19th to github@cloudant.com, the email address listed for the organization.

Is there another email I should send it to?

rhyshort:
@sdrew it should have been sent to mike.rhodes@uk.ibm.com as per the instructions in the CLA text.

rhyshort:
@sdrew We have your signed CLA, thanks for that :)

rhyshort:
Looks good, merging

rhyshort:
@sdrew This has been released as part of 0.19.2, thanks for the hard work.

Issue #154, opened 2015-05-22
Creating a cordova plugin for CDTDatastore


eforth:
I notice that you are using cocapods for dependency management. Is it possible to create a cordova plugin for CDTDatastore without using cocapods?

mikerhodes:
I'm not quite sure what you are asking -- could you elaborate a little on your question?

Thanks!

eforth:
I would like to create a cordova plugin for Cloudant Sync but my issue regarding the iOS version is that "CDTDatastore" uses cocapods a third party tool to handle dependency management. Is there a way to use "CDTDatastore" with cocopods?

mikerhodes:
Ah, I see. I've heard that it is difficult to combine Cocoapods and Cordova together.

Right now we don't have a supported way to build a standalone framework unfortunately. It is possible, we've done one-offs of dynamic frameworks for iOS 8 but we don't have a process to reliably create them.

At some point in the medium term, I'd like to get us working with https://github.com/Carthage/Carthage to get framework building more easily, but for now I'm guessing you'd need to run `pod install` to get the library into the cordova app, then use the plugin from there :-/

surajpandey-anz:
May I ask if that 'one-offs' dynamic frameworks for iOS 8 could be shared with the 'banking' client I mentioned in slack? 

A recap: The bank is looking at a cordova plugin for Cloudant Sync, that is IBM supported.

rhyshort:
There is a newly released Cordova plug-in which uses CDTDatastore, see https://github.com/cloudant/sync-cordova-plugin

mnewmedia:
Woooow super crazy fucking cool, just freaking out! 🎉 I love you all :P

PR #215, opened 2016-01-21
Enable Building CDTDatastore as a Framework

### LINKED ISSUE #194: Cannot build as a framework

rhyshort:
## What
Enable building CDTDatastore as a Framework.

## How
- Restructure projects to a single project with multiple targets. As a result the CDTDatastore and CDTDatastoreEncryption targets can be produced for distribution.
- Simplify podspec so there isn't a duplicate code.
- Remove TD* classes from the public interface so `CommonCrypto` includes don't stop the Framework from being built.
- Some Pre-processor tricks are required however to make CDT * Constants macros work with being in a framework. `<CommonCrypto/CommonCryptor.h>` needs to be included before CDT * Constants headers which refer to encryption.
- Created `CDTDefines.h` to contain `typedef` and `enum` declarations, to separate them from TD* classes.
- Update the Sample application to use CDTDatastore as a framework
- Include building the sample in the travis process to verify the framework can still be produced using cocoapods.
- Made header includes for deps to use the <podname/header.h> format so they can be used as a framework.

## Issues

This goes some way to making it easier to create a cordova plug-in as per #154 
Fixes #194 - using the `use_frameworks!` now results in a successful build for iOS 8+

## Notes:

- Jenkins build config needs to be updated so travis is the source of truth for the time being
- Contributing will need to updated to explain the new project layout.
- Cocoapods version 1.0.0.beta.2+ is required to make sure the build settings are correct


tomblench:
It looks like `clang-format` has been run over all of the moved files which makes it difficult to see where the genuine changes are. Often I wouldn't mind but for a large PR like this it makes reviewing really hard and will also cause problems with merging any other branches.

Can we look into either undoing the changes or re-doing the moves etc from scratch, ensuring that they do not get reformatted.

rhyshort:
Looking at each commit individually, I don't see where this has happened.

tomblench:
@rhyshort see https://github.com/cloudant/CDTDatastore/commit/f131a71b8a22eaa679a052f517a81608c174c063

rhyshort:
@tomblench if that's the only commit it's happen I think I can drop it then just fix the compiler errors for the missing headers again

tomblench:
@rhyshort OK, hopefully you won't get any conflicts if you drop that commit

rhyshort:
@tomblench done commit 1ddc8c3 just adds the header files

tomblench:
What's going on with the copy-info.plist and copy2-info.plist files? Are they still supposed to be in the commit?

rhyshort:
Currently yes, it was part of the target duplication, we could remove them and point at the single info.plist

tomblench:
+1

phatblat:
:clap:

Issue #216, opened 2016-02-02
Update Contributing


rhyshort:
After the merge of PR #215 contributing will become out of date. It needs to be updated to provide the correct instructions for setting up a development environment for CDTDatastore.

rhyshort:
Closing contributing is up to date

PR #217, opened 2016-02-03
Updates to documentation now library is built as a framework

tomblench:
Bug #216 - this should now be more correct now that there is only one workspace and a load more targets.

reviewer @rhyshort

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,7 +28,7 @@ you'll need:
 
 * Xcode
 * Xcode command line tools
-* Cocoapods
+* Cocoapods (minimum version 1.0.0 beta 2)
 * Homebrew (optional, but useful)
 * xcpretty (optional)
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -135,31 +147,37 @@ encrypted databases. More exactly, `CDTDatastore` relies on [FMDB][FMDB] to
...
-SQLCipher as a pod in your `Podfiles`, replace:
+If you want CDTDatastore to cipher the databases, you do need to include
+the `Encryption` variant as a pod in your `Podfiles`, replace:
 
 ```
 pod "CDTDatastore"
 ```
+or
+```
+pod "CDTDatastoreOSX"
```
rhyshort: There isn't a pod called `CDTDatastoreOSX`, thats a target in the project, the Pod is `CDTDatastore`. I'm not sure why we have this section on the contributing, perhaps we should drop everything in this section from line 161 down

rhyshort:
+1

PR #218, opened 2016-02-03
Add iOS replication policies

brynh:
This is part 1 of a 3 part change to implement replication policies. Each of the 3 parts of the implementation of replication policies is in a separate branch.

The 3 branches include:

_What?_

_Why?_

_How?_

reviewer @rhyshort
reviewer @tomblench

### DIFF
--- a/Classes/common/CDTReplicator/CDTReplicator.h
+++ b/Classes/common/CDTReplicator/CDTReplicator.h
@@ -16,7 +16,7 @@
 #import <Foundation/Foundation.h>
 
 #import "CDTReplicatorDelegate.h"
-#import "NSURLSessionConfigurationDelegate.h"
+#import "CDTNSURLSessionConfigurationDelegate.h"
 
 @class CDTDatastore;
 @class TDReplicatorManager;
@@ -137,9 +137,9 @@ typedef NS_ENUM(NSInteger, CDTReplicatorState) {
  * achieved by setting the NSURLSessionConfiguration's allowsCellularAccess
  * attribute to 'NO'.
  *
- * @see NSURLSessionConfigurationDelegate
+ * @see CDTNSURLSessionConfigurationDelegate
  */
-@property (nonatomic, weak) NSObject<NSURLSessionConfigurationDelegate> *sessionConfigDelegate;
+@property (nonatomic, weak) NSObject<CDTNSURLSessionConfigurationDelegate> *sessionConfigDelegate;
 
 /**
  Returns true if the state is `CDTReplicatorStatePending`, `CDTReplicatorStateStarted` or
--- a/Classes/common/CDTReplicator/CDTReplicator.h
+++ b/Classes/common/CDTReplicator/CDTReplicator.h
@@ -161,6 +161,7 @@ typedef NS_ENUM(NSInteger, CDTReplicatorState) {
  */
 -(id)initWithTDReplicatorManager:(TDReplicatorManager*)replicatorManager
                      replication:(CDTAbstractReplication*)replication
+           sessionConfigDelegate:(NSObject<CDTNSURLSessionConfigurationDelegate> *)delegate
                            error:(NSError * __autoreleasing*)error;
 
 /*
--- a/Classes/common/CDTReplicator/CDTReplicator.h
+++ b/Classes/common/CDTReplicator/CDTReplicator.h
@@ -16,6 +16,7 @@
 #import <Foundation/Foundation.h>
 
 #import "CDTReplicatorDelegate.h"
+#import "NSURLSessionConfigurationDelegate.h"
 
 @class CDTDatastore;
 @class TDReplicatorManager;
@@ -80,6 +81,7 @@ typedef NS_ENUM(NSInteger, CDTReplicatorState) {
     CDTReplicatorStateError
 };
 
+
 /**
  A CDTReplicator instance represents a replication job.
 
@@ -125,6 +127,19 @@ typedef NS_ENUM(NSInteger, CDTReplicatorState) {
  * @see CDTReplicatorDelegate
  */
 @property (nonatomic, weak) NSObject<CDTReplicatorDelegate> *delegate;
+
+/**
+ * Set the delegate for handling customisation of the NSURLSession
+ * used during replication.
+ *
+ * This allows the setting of specific options on the NSURLSessionConfiguration
+ * to control the replication - e.g. replication only when on Wifi would be
+ * achieved by setting the NSURLSessionConfiguration's allowsCellularAccess
+ * attribute to 'NO'.
+ *
+ * @see NSURLSessionConfigurationDelegate
+ */
+@property (nonatomic, weak) NSObject<NSURLSessionConfigurationDelegate> *sessionConfigDelegate;
 
 /**
  Returns true if the state is `CDTReplicatorStatePending`, `CDTReplicatorStateStarted` or
@@ -178,9 +193,20 @@ typedef NS_ENUM(NSInteger, CDTReplicatorState) {
  * return NO. Only when in `CDTReplicatorStatePending` will replication.
  *
  * @param error the error describing a failure.
+ * @param taskGroup an optional dispatch_group_t allowing clients to wait for the completion
+ *                  of all replication tasks in the taskGroup before proceeding.
  * @return YES or NO depending on success.
  *
  * @see CDTReplicatorState
+ */
+- (BOOL)startWithError:(NSError *__autoreleasing *)error taskGroup:(nullable dispatch_group_t)taskGroup;
+
+/**
+ * Starts a replication.
+ *
+ * Calling this is the same as calling startWithError:taskGroup: with a nil taskGroup.
+ *
+ * @see startWithError:taskGroup:
  */
 - (BOOL)startWithError:(NSError *__autoreleasing *)error;
 
--- a/Classes/common/CDTReplicator/CDTReplicator.h
+++ b/Classes/common/CDTReplicator/CDTReplicator.h
@@ -185,28 +185,28 @@ typedef NS_ENUM(NSInteger, CDTReplicatorState) {
  * The replication will continue until the replication is caught up with the source database;
  * that is, until there are no current changes to replicate.
  *
- * -startWithError can be called from any thread and will immediately return. It queues its work
- * on a separate replication thread. The methods on the CDTReplicatorDelegate
+ * -startWithTaskGroup:error: can be called from any thread and will immediately return. It queues 
+ * its work on a separate replication thread. The methods on the CDTReplicatorDelegate
  * may be called from the background threads.
  *
  * A given CDTReplicator instance cannot be reused. Calling this method more than once will
  * return NO. Only when in `CDTReplicatorStatePending` will replication.
  *
- * @param error the error describing a failure.
  * @param taskGroup an optional dispatch_group_t allowing clients to wait for the completion
  *                  of all replication tasks in the taskGroup before proceeding.
+ * @param error the error describing a failure.
  * @return YES or NO depending on success.
  *
  * @see CDTReplicatorState
  */
-- (BOOL)startWithError:(NSError *__autoreleasing *)error taskGroup:(nullable dispatch_group_t)taskGroup;
+- (BOOL)startWithTaskGroup:(nullable dispatch_group_t)taskGroup error:(NSError *__autoreleasing *)error;
 
 /**
  * Starts a replication.
  *
- * Calling this is the same as calling startWithError:taskGroup: with a nil taskGroup.
+ * Calling this is the same as calling startWithTaskGroup:error: with a nil taskGroup.
  *
- * @see startWithError:taskGroup:
+ * @see startWithTaskGroup:error:
  */
 - (BOOL)startWithError:(NSError *__autoreleasing *)error;
 
@@ -239,7 +239,7 @@ typedef NS_ENUM(NSInteger, CDTReplicatorState) {
  * to stop the replication before it starts. If it cannot stop the replication, this method
  * will return NO and replication will start and progress as normal.
  *
- * If -startWithError was never called, this method will move the state from
+ * If -startWithTaskGroup:error: was never called, this method will move the state from
  * CDTRelicatorStatePending to CDTReplicatorStateStopped, inform the delegate and return YES.
  *
  *
--- a/Classes/common/CDTReplicator/CDTReplicatorFactory.h
+++ b/Classes/common/CDTReplicator/CDTReplicatorFactory.h
@@ -71,6 +71,12 @@
 /**
  * Create a CDTReplicator object. This is the equivalent of calling
  * oneWay:sessionConfigDelegate:error: with a nil sessionConfigDelegate.
+ *
+ * @param replication a CDTPullReplication or CDTPushReplication
+ * @param error report error information
+ *
+ * @return a CDTReplicator instance which can be used to start and
+ *  stop the replication itself.
  */
 - (CDTReplicator *)oneWay:(CDTAbstractReplication *)replication
                     error:(NSError *__autoreleasing *)error;
--- a/Classes/common/CDTReplicator/CDTReplicatorFactory.h
+++ b/Classes/common/CDTReplicator/CDTReplicatorFactory.h
@@ -14,6 +14,7 @@
 //  and limitations under the License.
 
 #import <Foundation/Foundation.h>
+#import "CDTNSURLSessionConfigurationDelegate.h"
 
 @class CDTDatastore;
 @class CDTReplicator;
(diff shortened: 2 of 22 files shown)


Classes/common/CDTReplicator/CDTReplicator.h
```diff
@@ -16,6 +16,7 @@
 #import <Foundation/Foundation.h>
 
 #import "CDTReplicatorDelegate.h"
+#import "NSURLSessionConfigurationDelegate.h"
```
rhyshort: This import needs to be in the form `#import <Foundation/NSURLSessionConfiguration.h>` it's for correctness when we build as a framework

Classes/common/HTTP/CDTURLSession.h
```diff
@@ -15,43 +15,64 @@
 #import <Foundation/Foundation.h>
 #import "CDTURLSessionTask.h"
 #import "CDTMacros.h"
+#import "NSURLSessionConfigurationDelegate.h"
```
rhyshort: Angled bracket import here too

Classes/common/HTTP/CDTURLSession.h
```diff
@@ -15,43 +15,64 @@
...
- * @param thread the thread which callbacks should be run on
- * @param reuqestInterceptors array of interceptors that should be run before each request is made.
+ * @param thread the thread which callbacks should be run on.
+ * @param requestInterceptors array of interceptors that should be run before each request is made.
+ * @param sessionConfigDelegate the delegate used to customise the NSURLSessionConfiguration.
  **/
-- (instancetype)initWithDelegate:(NSObject<NSURLSessionDelegate> *)delegate
-                  callbackThread:(NSThread *)thread
-             requestInterceptors:(NSArray *)requestInterceptors NS_DESIGNATED_INITIALIZER;
+- (nonnull instancetype)initWithCallbackThread:(nonnull NSThread *)thread
```
rhyshort: This should be nullable since all `init` methods in Objc are failable

Classes/common/HTTP/CDTURLSession.m
```diff
@@ -56,22 +70,59 @@ - (instancetype)initWithDelegate:(NSObject<NSURLSessionDelegate> *)delegate
...
+
+{
+    NSURLSessionDataTask *nsURLSessionTask = [self.session dataTaskWithRequest:request];
+    [_taskMap setObject:task forKey:[NSNumber numberWithInteger:nsURLSessionTask.taskIdentifier]];
+    return nsURLSessionTask;
+}
+
+- (void)URLSession:(NSURLSession *)session dataTask:(NSURLSessionDataTask *)dataTask didReceiveData:(NSData *)data
+{
+    __strong CDTURLSessionTask *cdtURLSessionTask = [self getSessionTaskForId:dataTask.taskIdentifier];
```
rhyshort: Isn't this reference strong by default?

Classes/common/HTTP/CDTURLSession.m
```diff
@@ -56,22 +70,59 @@ - (instancetype)initWithDelegate:(NSObject<NSURLSessionDelegate> *)delegate
...
+    if (cdtURLSessionTask) {
+        [cdtURLSessionTask processResponse:response onThread:self.thread];
+    }
+    completionHandler(NSURLSessionResponseAllow);
+}
+
+- (CDTURLSessionTask *)getSessionTaskForId:(NSUInteger)identifier
+{
+    return [_taskMap objectForKey:[NSNumber numberWithInteger:identifier]];
```
rhyshort: use `self.taskMap` for accessing the property.

Classes/common/HTTP/CDTURLSessionTask.h
```diff
@@ -17,10 +17,20 @@
...
+
+@protocol CDTURLSessionTaskDelegate
+- (void)handleData:(nullable NSData *)data;
+- (void)handleResponse:(nullable NSURLResponse *)response;
+- (void)handleError:(nullable NSError *)error;
+@end
+
 @interface CDTURLSessionTask : NSObject
+{
+    id <CDTURLSessionTaskDelegate> _delegate;
```
rhyshort: I'm confused as to why you are doing this, shouldn't it be left up to the compiler to create the variable via the `delegate` property?

Classes/common/HTTP/CDTURLSessionTask.h
```diff
@@ -17,10 +17,20 @@
...
+}
 
-@property (nullable, nonatomic, copy) void (^completionHandler)
-    (NSData *__nullable data, NSURLResponse *__nullable response, NSError *__nullable error);
+@property (nullable, nonatomic, strong) id delegate;
```
rhyshort: Shouldn't this contain the type of delegate it is?

Classes/common/HTTP/CDTURLSession.m
```diff
@@ -56,22 +70,59 @@ - (instancetype)initWithDelegate:(NSObject<NSURLSessionDelegate> *)delegate
...
+}
+
+- (void) disassociateTask:(nonnull NSURLSessionDataTask *)task
+{
+    [_taskMap removeObjectForKey:[NSNumber numberWithInteger:task.taskIdentifier]];
```
rhyshort: Since this can be called from `dealloc` in the `CDTURLSessionTask` shouldn't this also cancel the inflight task to save network traffic since the result will be discarded?

Classes/common/HTTP/CDTURLSessionTask.m
```diff
@@ -84,12 +92,15 @@ - (void)cancel
     if (t) {
         [t cancel];
     }
+    _finished = YES;
```
rhyshort: I don't think this method does what it is supposed to anymore, I can't see somewhere where the `NSURLSessionTask` get set, so calling `-cancel` results in a no op. The test cases don't take account for this because they assume the `NSURLSessionTask` is created.

Classes/common/touchdb/TDRemoteRequest.h
```diff
@@ -18,7 +18,7 @@ typedef void (^TDRemoteRequestCompletionBlock)(id result, NSError* error);
 
 /** Asynchronous HTTP request; a fairly simple wrapper around NSURLConnection that calls a
  * completion block when ready. */
```
rhyshort: While we are here we should update this doc comment

Classes/common/touchdb/TDRemoteRequest.m
```diff
@@ -267,6 +253,18 @@ - (void)connectionDidFinishLoading:(NSURLConnection *)connection
     [self respondWithResult:self error:nil];
 }
 
+- (void)handleData:(NSData *)data {
+    [self receivedData:data];
```
rhyshort: shouldn't we just convert the `-receivedData` to be `-handleData`?

Classes/common/CDTReplicator/CDTReplicator.h
```diff
@@ -178,10 +193,21 @@ typedef NS_ENUM(NSInteger, CDTReplicatorState) {
  * return NO. Only when in `CDTReplicatorStatePending` will replication.
  *
  * @param error the error describing a failure.
+ * @param taskGroup an optional dispatch_group_t allowing clients to wait for the completion
+ *                  of all replication tasks in the taskGroup before proceeding.
  * @return YES or NO depending on success.
  *
  * @see CDTReplicatorState
  */
+- (BOOL)startWithError:(NSError *__autoreleasing *)error taskGroup:(nullable dispatch_group_t)taskGroup;
```
rhyshort: To follow traditional Objc method name I think we need to rename this to `-startWithTaskGroup:Error`

Classes/common/CDTReplicator/CDTReplicator.h
brynh: Not being an expert in building as frameworks for iOS, I want to clarify if this is necessary or just a bit of confusion caused by my naming of `NSURLSessionConfigurationDelegate.h`.

`NSURLSessionConfigurationDelegate.h` is a file I created, not part of the iOS foundation framework. I can understand the confusion though as I chose to name it in such a way that it ended up sounding like it's part of the foundation framework. I chose the name because it's a delegate for setting up the `NSURLSessionConfiguration`, so it made sense (to me) from that perspective, but it is a little confusing.

rhyshort: Oh I see, that is very confusing and it will be for our library consumers and apple reserve the two letter prefixes for themselves, so we should name it something with a CDT prefix, I think perhaps, `CDTNSURLSessionConfigurationDelegate.h`

Classes/common/HTTP/CDTURLSessionTask.m
brynh: I'm not sure I understand what you mean. This still cancels the `NSURLSessionTask` stored in the `self.inProgressTask`  property, which gets set in `resume`.

Classes/common/HTTP/CDTURLSession.h
brynh: Fixed in 55a78d9.

Classes/common/HTTP/CDTURLSession.m
brynh: Fixed in a81ebcc.

Classes/common/HTTP/CDTURLSession.m
brynh: Fixed in 6d6b809

Classes/common/HTTP/CDTURLSessionTask.h
brynh: Fixed in fc3dfaa.

Classes/common/HTTP/CDTURLSessionTask.h
brynh: Fixed in fc3dfaa.

Classes/common/HTTP/CDTURLSession.m
brynh: Good spot. Fixed in d3448e6

Classes/common/touchdb/TDRemoteRequest.h
brynh: Fixed in 13cf232

Classes/common/touchdb/TDRemoteRequest.m
brynh: Fixed in e199317 by renaming the methods in the protocol and making the associated changes.

Classes/common/CDTReplicator/CDTReplicator.h
brynh: Fixed in 13cf232

Classes/common/CDTReplicator/CDTReplicator.h
brynh: Renamed as per your suggestion in aefb854a.

Classes/common/HTTP/CDTURLSession.h
brynh: Fixed in aefb854a.

Classes/common/HTTP/CDTURLSessionTask.m
rhyshort: Oh I see, in my tiredness I got confused looking through the diff, I see its fine now

Classes/common/CDTReplicator/CDTReplicatorFactory.h
```diff
@@ -68,6 +69,13 @@
  */
 
 /**
+ * Create a CDTReplicator object. This is the equivalent of calling
```
tomblench: For completeness this should have `@param`s and `@return`

Classes/common/HTTP/CDTURLSession.m
```diff
@@ -45,10 +48,21 @@ - (instancetype)initWithDelegate:(NSObject<NSURLSessionDelegate> *)delegate
         _thread = thread;
         _interceptors = [NSArray arrayWithArray:requestInterceptors];
 
-        NSURLSessionConfiguration *config = [NSURLSessionConfiguration defaultSessionConfiguration];
+        NSURLSessionConfiguration *config;
+        // Create a unique session id using the address of self.
+        NSString *sessionId = [NSString stringWithFormat:@"com.cloudant.sync.sessionid.%p", self];
+        if ([[NSURLSessionConfiguration class] respondsToSelector:@selector(backgroundSessionConfigurationWithIdentifier:)]) {
```
tomblench: Why is this if/else required? Is this something to do with iOS/OSX versions and API availability? It probably deserves a comment.

Classes/common/HTTP/CDTURLSession.m
```diff
@@ -45,10 +48,21 @@ - (instancetype)initWithDelegate:(NSObject<NSURLSessionDelegate> *)delegate
...
+            config = [NSURLSessionConfiguration backgroundSessionConfiguration:sessionId];
+        }
+        config = [sessionConfigDelegate customiseNSURLSessionConfiguration:config];
+
         _session = [NSURLSession sessionWithConfiguration:config
-                                                 delegate:delegate
+                                                 delegate:self
                                             delegateQueue:nil];
+
+        _taskMap = [NSMapTable strongToWeakObjectsMapTable];
```
tomblench: The way `_taskMap` works could probably do with some documentation. Does this type of `NSMapTable` mean that values are removed if keys are reclaimed through reference counting?

Classes/common/CDTReplicator/CDTReplicatorFactory.h
brynh: Fixed in 5264aa8.

Classes/common/HTTP/CDTURLSession.m
brynh: Fixed in 5264aa8.

Classes/common/HTTP/CDTURLSessionTask.m
```diff
@@ -42,6 +44,8 @@ @interface CDTURLSessionTask ()
 
 @property (nonatomic) int remainingRetries;
 
+@property BOOL finished;
```
tomblench: Does this need to be some sort of atomic primitive? I'm not sure but it looks like it is being read from and written to by different threads.

Classes/common/HTTP/CDTURLSession.m
brynh: Comment added in 0573bac.

Classes/common/HTTP/CDTURLSessionTask.m
brynh: Properties are atomic by default, but I guess there's no harm being explicit about it. Updated in 3f6b6ec.

Classes/common/HTTP/CDTURLSessionTask.m
```diff
@@ -169,5 +162,50 @@ - (nonnull NSArray *)filterResponseInterceptors:(nonnull NSArray *)proposedRespo
...
+        [self.inProgressTask resume];
+    } else {
+        MYOnThread(thread, ^{
+            [self.delegate receivedResponse:response];
+        });
+        self.finished = YES;
+    }
+}
+
+- (void)processError:(NSError *)error onThread:(NSThread *)thread
```
tomblench: does `processError` change the handling of response interceptors in case of errors?

I guess previously the only way that the interceptors could handle errors would be to introspect the value of the HTTP status code in the NSHTTPURLResponse object. 

Maybe contexts need to have an extra `error` field, although this is could be considered a breaking change.

tomblench: As an additional comment here, what does "error" mean if we get to this block of code? Does it mean an HTTP 4xx/5xx error or does it mean something went wrong at a lower level, like a timeout/socket close etc?

rhyshort: This means there was a networking error, so the HTTP request wasn't completed, a HTTP 4xx/5xx would result in the response handler being called along with the interceptors

Classes/common/HTTP/CDTURLSession.m
```diff
@@ -45,10 +48,28 @@ - (instancetype)initWithDelegate:(NSObject<NSURLSessionDelegate> *)delegate
...
-        NSURLSessionConfiguration *config = [NSURLSessionConfiguration defaultSessionConfiguration];
+        NSURLSessionConfiguration *config;
+        // Create a unique session id using the address of self.
+        NSString *sessionId = [NSString stringWithFormat:@"com.cloudant.sync.sessionid.%p", self];
+
+        // NSURLSessionConfiguration:backgroundSessionConfigurationWithIdentifier was introduced in iOS 8.0
+        // to replace backgroundSessionConfiguration which was deprecated in iOS 8.0, so use the new version if
+        // available.
+        if ([[NSURLSessionConfiguration class] respondsToSelector:@selector(backgroundSessionConfigurationWithIdentifier:)]) {
+            config = [NSURLSessionConfiguration backgroundSessionConfigurationWithIdentifier:sessionId];
```
rhyshort: Isn't this going to cause a compilation error on iOS 7?

Classes/common/HTTP/CDTURLSessionTask.m
brynh: Indeed, Rhys is correct that `processError:onThread` deals with lower level networking errors, so I believe the overall handling of response interceptors in case of errors is unchanged - i.e. HTTP error codes will be handled in the `processResponse:onThread:` method and `processError:onThread:` deals with lower level networking errors, but in both cases the interceptors are called if appropriate.

Classes/common/HTTP/CDTURLSession.m
brynh: Good point - I'd only taken into account running on iOS7. Fixed in ffbb92b.

tomblench:
+1

rhyshort:
+1

PR #219, opened 2016-02-03
Add documentation updates

brynh:
This is part 2 of a 3 part change to implement replication policies. Each of the 3 parts of the implementation of replication policies is in a separate branch.

The 3 branches include:

_What?_

_Why?_

_How?_

reviewer @rhyshort
reviewer @tomblench

### DIFF
--- a/doc/replication-policies.md
+++ b/doc/replication-policies.md
@@ -31,7 +31,7 @@ of other apps.
 
 ### Example
 
-Lets assume we wish to configure a replication policy as follows:
+Let's assume we wish to configure a replication policy as follows:
 
 * We only ever want replications to occur when the device is connected to a WiFi network.
 * We want to do sync replications (pull and push).
--- a/doc/replication-policies.md
+++ b/doc/replication-policies.md
@@ -152,9 +152,10 @@ Implement the `application:performFetchWithCompletionHandler:` method - e.g.:
 {
     [self.myReplicator syncInBackgroundWithCompletionHandler:^{
         // Do whatever you want when replication is complete (e.g. refresh the UI)
-    }];
 
-    completionHandler(UIBackgroundFetchResultNewData);
+        // Call the completion handler once the sync has finished.
+        completionHandler(UIBackgroundFetchResultNewData);
+    }];
 }
 ```
 
--- a/doc/replication-policies.md
+++ b/doc/replication-policies.md
@@ -51,7 +51,7 @@ On the `Capabilities` tab of your Xcode project turn on `Background Modes` and s
 
 Configure your replications following the guidance in the [Replication guide](replication.md).
 
-Note however, that in the following example we'll use the `startWithError:taskGroup:` method of
+Note however, that in the following example we'll use the `startWithTaskGroup:error:` method of
 `CDTReplicator` instead of `startWithError:` so that we can batch the replication tasks and wait
 for the completion of all of them.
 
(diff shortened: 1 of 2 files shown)


doc/replication-policies.md
```diff
@@ -0,0 +1,195 @@
...
+
+        [[UIApplication sharedApplication] endBackgroundTask:taskId];
+    });
+}
+```
+
+Setup an NSURLSessionConfigurationDelegate to customise the NSURLSession as you require:
+
+```objc
+- (NSURLSessionConfiguration*)customiseNSURLSessionConfiguration:(nonnull NSURLSessionConfiguration *)config {
```
rhyshort: I'm a bit confused, how does this method get called from the replicator? from what I can see we only pass in a target and source without a delegate

brynh: The delegate is passed to the `CDTReplicator` by setting the delegate, in the example above,there are calls to `pushReplicator.sessionConfigDelegate = self;` and `pullReplicator.sessionConfigDelegate = self;`. Maybe it would be clearer to add the delegate to the factory method creating the `CDTReplicator` making it `oneway:sessionConfigDelegate:error:` instead of `oneway:error:`. Any thoughts?

rhyshort: I think that would be better adding it to factory method, we can call it as being nullable with the annotations too. We'll need to keep `-oneway:error` but that's not a big deal

brynh: Updated documentation in 635c98e and code in e29cce0.

doc/replication-policies.md
```diff
@@ -0,0 +1,197 @@
...
+
+For example:
+```objc
+/** Start both a push and pull replication.
+ *
+ *  @param taskGroup A dispatch_group_t to allow us to wait for replications to complete.
+ */
+- (void)startReplications:(dispatch_group_t)taskGroup
+{
+    // Create and start the replicator -- -start is essential!
```
tomblench: comment doesn't match code

doc/replication-policies.md
```diff
@@ -0,0 +1,197 @@
...
+-(void)syncInBackgroundWithCompletionHandler:(void (^)())completionHandler
+{
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_BACKGROUND, 0), ^{
+        dispatch_group_t backgroundTasks = dispatch_group_create();
+
+        UIBackgroundTaskIdentifier taskId = [[UIApplication sharedApplication] beginBackgroundTaskWithExpirationHandler:NULL];
+
+        [self startReplications:backgroundTasks];
+
+        dispatch_group_wait(backgroundTasks, dispatch_time(DISPATCH_TIME_NOW, 30000000000));
```
tomblench: Maybe use NSEC_PER_SEC? The documentation for this stuff is amazingly evasive at telling us if it's measure in µs or something else, but that's what I'm guessing.

brynh: Updated documentation in ce72d8 and code in 18a8b2d

doc/replication-policies.md
brynh: Good spot, thanks. Fixed in f7355c8.

doc/replication-policies.md
```diff
@@ -0,0 +1,197 @@
+# Replication Policies
+
+## iOS replication policies
```
rhyshort: Since most of the APIs for polices can be used on Mac too, I think we need to make it clear it's a cross platform feature, not sure how we'd go about it though

doc/replication-policies.md
```diff
@@ -0,0 +1,198 @@
...
+  `UIApplication` object. The value given to `setMinimumBackgroundFetchInterval:` is advisory only and the actual
+  time between background replications may vary considerably from the value set.
+
+You should note that enabling background fetch does not guarantee that iOS will give your app any time to perform
+background fetches as iOS will attempt to balance your app's need to fetch content with the needs of the system and
+of other apps.
+
+### Example
+
+Lets assume we wish to configure a replication policy as follows:
```
tomblench: `Let's`

tomblench:
+1

doc/replication-policies.md
brynh: Fixed in 09d1ac8.

rhyshort:
+1

PR #220, opened 2016-02-03
Update sample project.

brynh:
This is part 3 of a 3 part change to implement replication policies. Each of the 3 parts of the implementation of replication policies is in a separate branch.

The 3 branches include:

_What?_

_Why?_

_How?_

reviewer @rhyshort
reviewer @tomblench

### DIFF
--- a/Project/Project.xcodeproj/project.pbxproj
+++ b/Project/Project.xcodeproj/project.pbxproj
@@ -49,7 +49,7 @@
 		27C1CDCA184E2D9C000604EF /* UIKit.framework */ = {isa = PBXFileReference; lastKnownFileType = wrapper.framework; name = UIKit.framework; path = System/Library/Frameworks/UIKit.framework; sourceTree = SDKROOT; };
 		27C1CDCE184E2D9C000604EF /* Project-Info.plist */ = {isa = PBXFileReference; lastKnownFileType = text.plist.xml; path = "Project-Info.plist"; sourceTree = "<group>"; };
 		27C1CDD0184E2D9C000604EF /* en */ = {isa = PBXFileReference; lastKnownFileType = text.plist.strings; name = en; path = en.lproj/InfoPlist.strings; sourceTree = "<group>"; };
-		27C1CDD2184E2D9C000604EF /* main.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; path = main.m; sourceTree = "<group>"; tabWidth = 5; };
+		27C1CDD2184E2D9C000604EF /* main.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; path = main.m; sourceTree = "<group>"; };
 		27C1CDD4184E2D9C000604EF /* Project-Prefix.pch */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = "Project-Prefix.pch"; sourceTree = "<group>"; };
 		27C1CDD5184E2D9C000604EF /* CDTAppDelegate.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = CDTAppDelegate.h; sourceTree = "<group>"; };
 		27C1CDD6184E2D9C000604EF /* CDTAppDelegate.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; path = CDTAppDelegate.m; sourceTree = "<group>"; };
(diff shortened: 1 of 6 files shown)


Project/Project/Project-Info.plist
```diff
@@ -36,5 +40,18 @@
 		<string>UIInterfaceOrientationLandscapeLeft</string>
 		<string>UIInterfaceOrientationLandscapeRight</string>
 	</array>
+	<key>NSAppTransportSecurity</key>
```
rhyshort: Is this still needed now that cloudant.com has a SHA2 cert?

brynh: Good spot. This must have been automatically added at some point. Fixed in 4c4bb67.

Project/Project.xcodeproj/project.pbxproj
```diff
@@ -49,7 +49,7 @@
 		27C1CDCA184E2D9C000604EF /* UIKit.framework */ = {isa = PBXFileReference; lastKnownFileType = wrapper.framework; name = UIKit.framework; path = System/Library/Frameworks/UIKit.framework; sourceTree = SDKROOT; };
 		27C1CDCE184E2D9C000604EF /* Project-Info.plist */ = {isa = PBXFileReference; lastKnownFileType = text.plist.xml; path = "Project-Info.plist"; sourceTree = "<group>"; };
 		27C1CDD0184E2D9C000604EF /* en */ = {isa = PBXFileReference; lastKnownFileType = text.plist.strings; name = en; path = en.lproj/InfoPlist.strings; sourceTree = "<group>"; };
-		27C1CDD2184E2D9C000604EF /* main.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; path = main.m; sourceTree = "<group>"; };
+		27C1CDD2184E2D9C000604EF /* main.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; path = main.m; sourceTree = "<group>"; tabWidth = 5; };
```
tomblench: How did this `tabWidth = 5;` sneak in? I guess it shouldn't be there.

brynh: Ooops. Don't know how that got there - fixed in fe04faa.

Project/Project/CDTTodoReplicator.h
```diff
@@ -18,8 +18,8 @@
 
 #import <CloudantSync.h>
 
-@interface CDTTodoReplicator : NSObject<CDTReplicatorDelegate>
+@interface CDTTodoReplicator : NSObject<CDTReplicatorDelegate, NSURLSessionConfigurationDelegate>
 
--(void)sync;
+- (void)syncInBackgroundWithCompletionHandler:(void (^)())completionHandler;
```
rhyshort: Since we've added the completion handler why don't we use it in the AppDelegate?

Project/Project/CDTViewController.m
```diff
@@ -160,13 +167,9 @@ - (void)addTodoButtonTap:(NSObject *)sender {
...
 
-    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-        [self.todoReplicator sync];
-        dispatch_async(dispatch_get_main_queue(), ^{
-            [self reloadTasks];
-            [self.tableView reloadData];
-        });
-    });
+    [self.todoReplicator syncInBackgroundWithCompletionHandler:^{
+        [self toggleCompletedShown:nil];
```
rhyshort: we shouldn't capture `self` here, we should set a good example, and use:

```objc
__weak CDTViewController *weakSelf = self;
[self.todoReplicator syncInBackgroundWithCompletionHandler:^{
    __strong CDTViewController *self = weakSelf;
    self toggleCompletedShown:nil];
}];
```

or similar.

Project/Project/CDTViewController.m
```diff
@@ -351,5 +354,30 @@ -(void)refreshTodoList
     [self.tableView endUpdates];
 }
 
+- (void)periodicReplication:(NSTimer *)timer
+{
+    [self.todoReplicator syncInBackgroundWithCompletionHandler:^{
+        [self performSelectorOnMainThread:@selector(refresh) withObject:nil waitUntilDone:false];
```
rhyshort: again we shouldn't capture self here

Project/Project/CDTTodoReplicator.h
brynh: Fixed in 8cdad4a.

Project/Project/CDTViewController.m
brynh: Fixed in 6c2ae81.

Project/Project/CDTViewController.m
brynh: Fixed in 6c2ae81.

tomblench:
+1

rhyshort:
+1

PR #221, opened 2016-02-18
Replication policies feature

brynh:
This feature branch is the merge of all the following branches following completion of their PR reviews:
1. _replication-policies-1_ (PR #218): The main code changes implementing replication policies on iOS.
2. _replication-policies-2-doc_ (PR #219): Documentation updates.
3. _replication-policies-3-sample-project_ (PR #220): Updates to the sample project to add periodic replication when on WiFi.

_What?_

_Why?_

_How?_

reviewer @rhyshort


rhyshort:
+1, However, before we make a 1.1 release to cocoapods, we need to await at least one nightly build to complete.

PR #225, opened 2016-03-07
Audit Query APIs for Nullability and add annotations as required

rhyshort:
## What
Audit the Query APIs (CDTQ prefixed classes) for nullability and add annotations as required.

## How

Go through each header and inspect underlying code to identify which parameters or properties can be 
null and which ones cannot be null.

## Reviewers
reviewer @brynh 
## Notes

Important: Due to the API impact in Swift, it needs to be made sure that the annotations are correct, this is because an incorrect annotation for an API to be `nonnull` but can in fact be null could result in a crash.

### DIFF
--- a/CDTDatastore/CDTDocumentRevision.h
+++ b/CDTDatastore/CDTDocumentRevision.h
@@ -20,6 +20,8 @@
 
 @class TD_RevisionList;
 
+NS_ASSUME_NONNULL_BEGIN
+
 /**
  * Represents a single revision of a document in a datastore.
  */
@@ -35,8 +37,8 @@
 
 @property (nonatomic, readonly) SequenceNumber sequence;
 
-@property (nonnull, nonatomic, strong) NSMutableDictionary *body;
-@property (nonnull, nonatomic, strong) NSMutableDictionary *attachments;
+@property (nonatomic, strong) NSMutableDictionary *body;
+@property (nonatomic, strong) NSMutableDictionary *attachments;
 
 @property (nonatomic, getter=isChanged) bool changed;
 
@@ -49,17 +51,17 @@
  */
 - (BOOL)isFullRevision;
 
-- (nullable instancetype)initWithDocId:(nullable NSString *)docId
-         revisionId:(nullable NSString *)revId
-               body:(nullable NSDictionary *)body
-        attachments:(nullable NSDictionary *)attachments;
+- (instancetype)initWithDocId:(nullable NSString *)docId
+                   revisionId:(nullable NSString *)revId
+                         body:(nullable NSDictionary *)body
+                  attachments:(nullable NSDictionary *)attachments;
 
-- (nullable instancetype)initWithDocId:(nullable NSString *)docId
-         revisionId:(nullable NSString *)revId
-               body:(nullable NSDictionary *)body
-            deleted:(BOOL)deleted
-        attachments:(nullable NSDictionary *)attachments
-           sequence:(SequenceNumber)sequence;
+- (instancetype)initWithDocId:(nullable NSString *)docId
+                   revisionId:(nullable NSString *)revId
+                         body:(nullable NSDictionary *)body
+                      deleted:(BOOL)deleted
+                  attachments:(nullable NSDictionary *)attachments
+                     sequence:(SequenceNumber)sequence;
 
 /**
  Creates an CDTDocumentRevision from JSON Data
@@ -75,19 +77,18 @@
 
  @return new CDTDocumentRevision instance
 */
-+ (nullable CDTDocumentRevision *)createRevisionFromJson:(nonnull NSDictionary *)jsonDict
-                                    forDocument:(nonnull NSURL *)documentURL
-                                          error:(NSError *__autoreleasing __nullable * __nullable)error __attribute__((deprecated));;
-
++ (nullable CDTDocumentRevision *)createRevisionFromJson:(NSDictionary *)jsonDict
+                                             forDocument:(NSURL *)documentURL
+                                                   error:(NSError *__autoreleasing __nullable * __nullable)error __attribute__((deprecated));
 /**
  Create a new, blank revision which will have an ID generated on saving.
  */
-+ (nullable CDTDocumentRevision *)revision;
++ (CDTDocumentRevision *)revision;
 
 /**
  Create a new, blank revision with an assigned ID.
  */
-+ (nullable CDTDocumentRevision *)revisionWithDocId:(nonnull NSString *)docId;
++ (CDTDocumentRevision *)revisionWithDocId:(NSString *)docId;
 
 /**
  Create a blank revision with a rev ID, which will be treated as an update when saving.
@@ -98,7 +99,7 @@
  Useful during testing and when it's not necessary to start with an existing revision's
  content.
  */
-+ (nullable CDTDocumentRevision *)revisionWithDocId:(nonnull NSString *)docId revId:(nonnull NSString *)revId;
++ (CDTDocumentRevision *)revisionWithDocId:(NSString *)docId revId:(NSString *)revId;
 
 /**
  Return document content as an NSData object.
@@ -120,6 +121,8 @@
 
  @return copy of this document
  */
-- (nullable CDTDocumentRevision *)copy;
+- (CDTDocumentRevision *)copy;
 
 @end
+
+NS_ASSUME_NONNULL_END
--- a/CDTDatastore/query/CDTQIndexManager.h
+++ b/CDTDatastore/query/CDTQIndexManager.h
@@ -14,6 +14,8 @@
 
 #import <Foundation/Foundation.h>
 
+NS_ASSUME_NONNULL_BEGIN
+
 extern NSString *const CDTQIndexManagerErrorDomain;
 extern NSString *const kCDTQIndexTablePrefix;
 extern NSString *const kCDTQIndexMetadataTableName;
@@ -75,42 +77,43 @@ typedef NS_ENUM(NSInteger, CDTQQueryError) {
 /**
  Constructs a new CDTQIndexManager which indexes documents in `datastore`
  */
-+ (CDTQIndexManager *)managerUsingDatastore:(CDTDatastore *)datastore
-                                      error:(NSError *__autoreleasing *)error;
++ (nullable CDTQIndexManager *)
+managerUsingDatastore:(CDTDatastore *)datastore
+                error:(NSError *__autoreleasing __nullable *__nullable)error;
 
-- (instancetype)initUsingDatastore:(CDTDatastore *)datastore
-                             error:(NSError *__autoreleasing *)error;
+- (nullable instancetype)initUsingDatastore:(CDTDatastore *)datastore
+                                      error:(NSError *__autoreleasing __nullable *__nullable)error;
 
-- (NSDictionary * /* NSString -> NSArray[NSString]*/)listIndexes;
+- (NSDictionary<NSString *, NSArray<NSString *> *> *)listIndexes;
 
 /** Internal */
-+ (NSDictionary /* NSString -> NSArray[NSString]*/ *)listIndexesInDatabaseQueue:
-        (FMDatabaseQueue *)db;
++ (NSDictionary<NSString *, NSArray<NSString *> *> *)listIndexesInDatabaseQueue:
+    (FMDatabaseQueue *)db;
 /** Internal */
-+ (NSDictionary /* NSString -> NSArray[NSString]*/ *)listIndexesInDatabase:(FMDatabase *)db;
++ (NSDictionary<NSString *, NSArray<NSString *> *> *)listIndexesInDatabase:(FMDatabase *)db;
 
-- (NSString *)ensureIndexed:(NSArray * /* NSString */)fieldNames withName:(NSString *)indexName;
+- (nullable NSString *)ensureIndexed:(NSArray<NSString *> *)fieldNames withName:(NSString *)indexName;
 
-- (NSString *)ensureIndexed:(NSArray * /* NSString */)fieldNames
-                   withName:(NSString *)indexName
-                       type:(NSString *)type;
+- (nullable NSString *)ensureIndexed:(NSArray<NSString *> *)fieldNames
+                            withName:(NSString *)indexName
+                                type:(NSString *)type;
 
-- (NSString *)ensureIndexed:(NSArray * /* NSString */)fieldNames
-                   withName:(NSString *)indexName
-                       type:(NSString *)type
-                   settings:(NSDictionary *)indexSettings;
+- (nullable NSString *)ensureIndexed:(NSArray<NSString *> *)fieldNames
+                            withName:(NSString *)indexName
+                                type:(NSString *)type
+                            settings:(nullable NSDictionary *)indexSettings;
 
 - (BOOL)deleteIndexNamed:(NSString *)indexName;
 
 - (BOOL)updateAllIndexes;
 
-- (CDTQResultSet *)find:(NSDictionary *)query;
+- (nullable CDTQResultSet *)find:(NSDictionary *)query;
 
-- (CDTQResultSet *)find:(NSDictionary *)query
-                   skip:(NSUInteger)skip
-                  limit:(NSUInteger)limit
-                 fields:(NSArray *)fields
-                   sort:(NSArray *)sortDocument;
+- (nullable CDTQResultSet *)find:(NSDictionary *)query
+                            skip:(NSUInteger)skip
+                           limit:(NSUInteger)limit
+                          fields:(nullable NSArray *)fields
+                            sort:(nullable NSArray *)sortDocument;
 
 /** Internal */
 + (NSString *)tableNameForIndex:(NSString *)indexName;
@@ -119,3 +122,4 @@ typedef NS_ENUM(NSInteger, CDTQQueryError) {
 + (BOOL)ftsAvailableInDatabase:(FMDatabaseQueue *)db;
 
 @end
+NS_ASSUME_NONNULL_END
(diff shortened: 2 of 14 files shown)


CDTDatastore/query/CDTQIndexManager.h
```diff
@@ -75,47 +77,49 @@ typedef NS_ENUM(NSInteger, CDTQQueryError) {
 /**
  Constructs a new CDTQIndexManager which indexes documents in `datastore`
  */
-+ (CDTQIndexManager *)managerUsingDatastore:(CDTDatastore *)datastore
-                                      error:(NSError *__autoreleasing *)error;
++ (nullable CDTQIndexManager *)
+managerUsingDatastore:(CDTDatastore *)datastore
+                error:(NSError *__autoreleasing __nullable *__nullable)error;
```
brynh: From the [Nullability and Objective-C blog](https://developer.apple.com/swift/blog/?id=25), I don't think it's necessary to have the nullability annotations for `NSError **` in an audited region:

> The particular type `NSError **` is so often used to return errors via method parameters that it is always assumed to be a nullable pointer to a nullable `NSError` reference.

rhyshort: although it may not be needed to tell the compiler that it this parameter can be `nil` I think its better practise to be explicit so when we are programming error handling code, it's easier for us to see that we should do the check to make sure it isn't `nil`

CDTDatastore/query/CDTQIndexManager.h
```diff
@@ -75,47 +77,49 @@ typedef NS_ENUM(NSInteger, CDTQQueryError) {
...
-+ (NSDictionary /* NSString -> NSArray[NSString]*/ *)listIndexesInDatabaseQueue:
-        (FMDatabaseQueue *)db;
++ (NSDictionary<NSString *, NSArray<NSString *> *> *)listIndexesInDatabaseQueue:
+    (FMDatabaseQueue *)db;
 /** Internal */
-+ (NSDictionary /* NSString -> NSArray[NSString]*/ *)listIndexesInDatabase:(FMDatabase *)db;
++ (NSDictionary<NSString *, NSArray<NSString *> *> *)listIndexesInDatabase:(FMDatabase *)db;
 
-- (NSString *)ensureIndexed:(NSArray * /* NSString */)fieldNames withName:(NSString *)indexName;
+- (NSString *)ensureIndexed:(NSArray<NSString *> *)fieldNames withName:(NSString *)indexName;
```
brynh: I think this method can return `nil`. For example, if `indexName` is of zero length.

CDTDatastore/query/CDTQIndexManager.h
```diff
@@ -75,47 +77,49 @@ typedef NS_ENUM(NSInteger, CDTQQueryError) {
...
 
-- (NSString *)ensureIndexed:(NSArray * /* NSString */)fieldNames
+- (NSString *)ensureIndexed:(NSArray<NSString *> *)fieldNames
```
brynh: I think this method can return `nil`. For example, if `indexName` is of zero length.

CDTDatastore/query/CDTQIndexManager.h
```diff
@@ -75,47 +77,49 @@ typedef NS_ENUM(NSInteger, CDTQQueryError) {
...
                    withName:(NSString *)indexName
                        type:(NSString *)type;
 
-- (NSString *)ensureIndexed:(NSArray * /* NSString */)fieldNames
+- (NSString *)ensureIndexed:(NSArray<NSString *> *)fieldNames
```
brynh: I think this method can return `nil`. For example, if `indexName` is of zero length.

CDTDatastore/query/CDTQIndexManager.h
rhyshort: fixed in 5710b98

CDTDatastore/query/CDTQIndexManager.h
rhyshort: fixed in 5710b98

CDTDatastore/query/CDTQIndexManager.h
rhyshort: fixed in 5710b98

CDTDatastore/query/CDTQIndexManager.h
```diff
@@ -75,47 +77,49 @@ typedef NS_ENUM(NSInteger, CDTQQueryError) {
...
-                   sort:(NSArray *)sortDocument;
+- (nullable CDTQResultSet *)find:(NSDictionary *)query
+                            skip:(NSUInteger)skip
+                           limit:(NSUInteger)limit
+                          fields:(nullable NSArray *)fields
+                            sort:(nullable NSArray *)sortDocument;
 
 /** Internal */
-+ (NSString *)tableNameForIndex:(NSString *)indexName;
++ (nullable NSString *)tableNameForIndex:(NSString *)indexName;
```
brynh: The return value for this is never null.

rhyshort: fixed in 89b770d

CDTDatastore/query/CDTQProjectedDocumentRevision.h
```diff
@@ -28,12 +30,14 @@
...
  Initialise with a datastore so mutableCopy can return a full document.
  */
-- (id)initWithDocId:(NSString *)docId
-         revisionId:(NSString *)revId
-               body:(NSDictionary *)body
-            deleted:(BOOL)deleted
-        attachments:(NSDictionary *)attachments
-           sequence:(SequenceNumber)sequence
-          datastore:(CDTDatastore *)datastore;
+- (nullable instancetype)initWithDocId:(NSString *)docId
```
brynh: The return value for this is never null.

CDTDatastore/query/CDTQQueryExecutor.h
```diff
@@ -28,7 +30,8 @@
  Constructs a new CDTQQueryExecutor using the indexes in `database` to find documents from
  `datastore`.
  */
-- (instancetype)initWithDatabase:(FMDatabaseQueue *)database datastore:(CDTDatastore *)datastore;
+- (nullable instancetype)initWithDatabase:(FMDatabaseQueue *)database
```
brynh: The return value for this is never null.

CDTDatastore/query/CDTQResultSet.h
```diff
@@ -51,13 +53,15 @@ typedef void (^CDTQResultSetBuilderBlock)(CDTQResultSetBuilder *configuration);
     NSArray *_originalDocumentIds;
 }
 
-+ (instancetype)resultSetWithBlock:(CDTQResultSetBuilderBlock)block;
++ (nullable instancetype)resultSetWithBlock:(CDTQResultSetBuilderBlock)block;
 
-- (instancetype)initWithBuilder:(CDTQResultSetBuilder *)builder;
+- (nullable instancetype)initWithBuilder:(CDTQResultSetBuilder *)builder;
```
brynh: The return value for this is never null.

CDTDatastore/query/CDTQProjectedDocumentRevision.h
rhyshort: It can, because `NSObject` `init` can return `nil`

CDTDatastore/query/CDTQQueryExecutor.h
rhyshort: It can return nil because `NSObject` can return nil from it's `-init`

CDTDatastore/query/CDTQResultSet.h
rhyshort: It can return nil because `NSObject` can return nil from it's `-init`

rhyshort: It can return nil because `NSObject` can return nil from it's `-init`

CDTDatastore/query/CDTQProjectedDocumentRevision.h
brynh: I don't believe `NSObject init` can return `nil`. Although it says in the [doc](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSObject_Class/#//apple_ref/occ/instm/NSObject/init):

> An initialized object, or nil if an object could not be created for some reason that would not result in an exception.

It goes on to say:

> The init method defined in the NSObject class does no initialization; it simply returns self.

So I think the former comment refers to `init` methods in general, whereas the latter refers to `NSObject init` specifically.

rhyshort: perhaps however when I looked at the header, it doesn't have nullability annotations or an audited region, implying that it can return null, although that might suggest that it's only to allow you to return null

brynh: Personally, I'm happy that the docs state that `NSObject`'s `init` can't return `nil`, and I can't see why they'd ever change it from just returning `self` which can surely never fail. Because of my lack of Swift knowledge I'm not sure what the impact is of having unnecessary `nullable` annotations other than them potentially propagating all round the code suggesting nullability that can't happen and presumably resulting in unnecessary checks in the code.

rhyshort: It would result in more `if let` statements although this being said, I've taken a quick look at the header of NSString, and it returns a nonnull value for init. Here's an extract that shows it:

```objc
@interface NSString : NSObject <NSCopying, NSMutableCopying, NSSecureCoding>

#pragma mark *** String funnel methods ***

/* NSString primitives. A minimal subclass of NSString just needs to implement these two, along with an init method appropriate for that subclass. We also recommend overriding getCharacters:range: for performance.
 */
@property (readonly) NSUInteger length;
- (unichar)characterAtIndex:(NSUInteger)index;

/* The initializers available to subclasses. See further below for additional init methods.
*/
- (instancetype)init NS_DESIGNATED_INITIALIZER;
- (nullable instancetype)initWithCoder:(NSCoder *)aDecoder NS_DESIGNATED_INITIALIZER;
```

Its in an audited section, so I think if its good enough for Apple it should probably be good enough for us. Any thoughts @brynh?

brynh: Good find. I think that as Apple themselves aren't annotating `init` as `nullable`, I'd definitely prefer we remove these unnecessary `nullable` annotations.

rhyshort:
@brynh I've made the changes to the init chains for the objects mentioned in the diff comments

CDTDatastore/CDTDocumentRevision.h
```diff
@@ -106,7 +108,7 @@
 
  @return document content as an NSData object.
  */
-- (nullable NSData *)documentAsDataError:(NSError *__autoreleasing __nullable * __nullable)error;
+- (NSData *)documentAsDataError:(NSError *__autoreleasing __nullable * __nullable)error;
```
brynh: I think this should still be `nullable`.

CDTDatastore/CDTDocumentRevision.h
```diff
@@ -49,12 +51,12 @@
  */
 - (BOOL)isFullRevision;
 
-- (nullable instancetype)initWithDocId:(nullable NSString *)docId
+- (instancetype)initWithDocId:(nullable NSString *)docId
          revisionId:(nullable NSString *)revId
```
brynh: Indentation

CDTDatastore/CDTDocumentRevision.h
```diff
@@ -49,12 +51,12 @@
...
                body:(nullable NSDictionary *)body
         attachments:(nullable NSDictionary *)attachments;
 
-- (nullable instancetype)initWithDocId:(nullable NSString *)docId
+- (instancetype)initWithDocId:(nullable NSString *)docId
          revisionId:(nullable NSString *)revId
```
brynh: Indentation

CDTDatastore/CDTDocumentRevision.h
```diff
@@ -72,19 +74,19 @@
 
  @return new CDTDocumentRevision instance
 */
-+ (nullable CDTDocumentRevision *)createRevisionFromJson:(nonnull NSDictionary *)jsonDict
-                                    forDocument:(nonnull NSURL *)documentURL
++ (nullable CDTDocumentRevision *)createRevisionFromJson:(NSDictionary *)jsonDict
+                                    forDocument:(NSURL *)documentURL
```
brynh: Indentation

CDTDatastore/query/CDTQQueryExecutor.h
```diff
@@ -30,7 +30,7 @@ NS_ASSUME_NONNULL_BEGIN
  Constructs a new CDTQQueryExecutor using the indexes in `database` to find documents from
  `datastore`.
  */
-- (nullable instancetype)initWithDatabase:(FMDatabaseQueue *)database
+- (instancetype)initWithDatabase:(FMDatabaseQueue *)database
                                 datastore:(CDTDatastore *)datastore;
```
brynh: Indentation

CDTDatastore/query/CDTQIndexManager.h
```diff
@@ -29,7 +31,7 @@ extern NSString *const kCDTQIndexMetadataTableName;
 @property (nonatomic, strong) NSString *sqlWithPlaceholders;
 @property (nonatomic, strong) NSArray *placeholderValues;
 
-+ (CDTQSqlParts *)partsForSql:(NSString *)sql parameters:(NSArray *)parameters;
++ (nullable CDTQSqlParts *)partsForSql:(NSString *)sql parameters:(NSArray *)parameters;
```
brynh: I don't think this can't return `nil`.

CDTDatastore/query/CDTQIndexUpdater.h
```diff
@@ -42,23 +44,24 @@
...
       fromDatastore:(CDTDatastore *)datastore
-              error:(NSError *__autoreleasing *)error;
+              error:(NSError *__nullable __autoreleasing *__nullable)error;
 
 /**
  Constructs a new CDTQQueryExecutor using the indexes in `database` to index documents from
  `datastore`.
  */
-- (instancetype)initWithDatabase:(FMDatabaseQueue *)database datastore:(CDTDatastore *)datastore;
+- (nullable instancetype)initWithDatabase:(FMDatabaseQueue *)database
```
brynh: This can't return `nil`.

CDTDatastore/CDTDocumentRevision.h
rhyshort: fixed in be679a0

CDTDatastore/CDTDocumentRevision.h
rhyshort: fixed in be679a0

CDTDatastore/CDTDocumentRevision.h
rhyshort: fixed in be679a0

CDTDatastore/query/CDTQIndexUpdater.h
```diff
@@ -66,21 +69,21 @@
...
-              error:(NSError *__autoreleasing *)error;
+         withFields:(NSArray<NSString *> *)fieldNames
+              error:(NSError *__autoreleasing __nullable *__nullable)error;
 
 /**
  Generate the DELETE statement to remove a documents entries from an index.
  */
-+ (CDTQSqlParts *)partsToDeleteIndexEntriesForDocId:(NSString *)docId
-                                          fromIndex:(NSString *)indexName;
++ (nullable CDTQSqlParts *)partsToDeleteIndexEntriesForDocId:(NSString *)docId
```
brynh: I don't think this can return `nil` now the args are nonnull as they're in an audited section.

CDTDatastore/CDTDocumentRevision.h
rhyshort: fixed in be679a0

CDTDatastore/query/CDTQQueryExecutor.h
rhyshort: fixed in be679a0

CDTDatastore/query/CDTQIndexManager.h
rhyshort: fixed in b04fa2e

CDTDatastore/query/CDTQIndexUpdater.h
rhyshort: fixed in b04fa2e

CDTDatastore/query/CDTQIndexUpdater.h
rhyshort: fixed in 61cea14

CDTDatastore/query/CDTQProjectedDocumentRevision.h
```diff
@@ -28,12 +30,14 @@
...
  */
-- (id)initWithDocId:(NSString *)docId
-         revisionId:(NSString *)revId
-               body:(NSDictionary *)body
-            deleted:(BOOL)deleted
-        attachments:(NSDictionary *)attachments
-           sequence:(SequenceNumber)sequence
-          datastore:(CDTDatastore *)datastore;
+- (instancetype)initWithDocId:(NSString *)docId
+                            revisionId:(NSString *)revId
```
brynh: Indentation

CDTDatastore/query/CDTQQueryExecutor.h
```diff
@@ -28,7 +30,8 @@
  Constructs a new CDTQQueryExecutor using the indexes in `database` to find documents from
  `datastore`.
  */
-- (instancetype)initWithDatabase:(FMDatabaseQueue *)database datastore:(CDTDatastore *)datastore;
+- (instancetype)initWithDatabase:(FMDatabaseQueue *)database
+                                datastore:(CDTDatastore *)datastore;
```
brynh: Indentation

brynh:
+1

PR #227, opened 2016-03-10
Use Cookie Authentication

rhyshort:
## What
Use cookie authentication instead of basic auth for replications
## How
Strip out user info from the url when creating a CDT*Replication class instance.
## Testing
New tests added to CDTReplicationTests.

## Reviewers
reviewer @tomblench 
reviewer @ricellis

### DIFF
--- a/CDTDatastore/CDTReplicator/CDTPullReplication.m
+++ b/CDTDatastore/CDTReplicator/CDTPullReplication.m
@@ -14,6 +14,7 @@
 //  and limitations under the License.
 
 #import "CDTPullReplication.h"
+#import "CDTSessionCookieInterceptor.h"
 #import "CDTDatastore.h"
 #import "CDTLogging.h"
 #import "TDMisc.h"
@@ -33,7 +34,16 @@
 - (instancetype)initWithSource:(NSURL *)source target:(CDTDatastore *)target
 {
     if (self = [super init]) {
-        _source = source;
+        
+        NSURLComponents * sourceComponents = [NSURLComponents componentsWithURL:source resolvingAgainstBaseURL:NO];
+        if(sourceComponents.user && sourceComponents.password){
+            CDTSessionCookieInterceptor * cookieInterceptor = [[CDTSessionCookieInterceptor alloc] initWithUsername:sourceComponents.user password:sourceComponents.password];
+            sourceComponents.user = nil;
+            sourceComponents.password = nil;
+            [self addInterceptor:cookieInterceptor];
+        }
+        
+        _source = sourceComponents.URL;
         _target = target;
     }
     return self;
--- a/CDTDatastoreTests/CDTReplicationTests.m
+++ b/CDTDatastoreTests/CDTReplicationTests.m
@@ -26,6 +26,7 @@
 #import "TD_Revision.h"
 #import "TDPuller.h"
 #import "TDPusher.h"
+#import "CDTSessionCookieInterceptor.h"
 #import <OHHTTPStubs/OHHTTPStubs.h>
 #import <OHHTTPStubs/OHHTTPStubsResponse+JSON.h>
 #import <OCMock/OCMock.h>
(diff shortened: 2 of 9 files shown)


tomblench:
@rhyshort - this may be material for a follow-on PR, but we should tidy up some of the now-obselete authentication code which is confusing and some of it is never invoked.

I was looking into PR #228 to fix a failing assertion and realised that tests like `testURLConnectionChangeTrackerWithBadCredentials` and `testURLConnectionChangeTrackerWithRealRemote` will never construct `TDAuthorizer` objects and nor will they receive challenges through `didReceiveChallenge`. This is because `NSURLSession` appears to use the username/password from the URL to re-invoke the request with basic auth credentials added.

Next steps will look something like:
- Ensure user can do basic auth (through Interceptor?) if the don't want to use cookies. For efficiency, the header should be set before making the request, instead of waiting for the challenge which is the `NSURLSession` default behaviour.
- Remove `TDAuthorizer` and all references to it.
- Consider removing `didReceiveChallenge` delegate methods - this will depend on if/how we want to support other authentication types.

rhyshort:
Yeah, I think making a following in tow remove the TD authorities stuff, then another to follow up with a basic auth interceptor. Can you raise those two things as issues, and I'll deal with them when I'm back in the office on Monday.

CDTDatastore/CDTReplicator/CDTPullReplication.m
```diff
@@ -33,7 +34,16 @@ + (instancetype)replicationWithSource:(NSURL *)source target:(CDTDatastore *)tar
 - (instancetype)initWithSource:(NSURL *)source target:(CDTDatastore *)target
 {
     if (self = [super init]) {
-        _source = source;
+        
+        NSURLComponents * sourceComponents = [NSURLComponents componentsWithURL:source resolvingAgainstBaseURL:NO];
```
tomblench: You could do this in the superclass instead of in both `CDT{Pull,Push}Replication` and avoid the awkward code duplication

CDTDatastoreTests/CDTReplicationTests.m
```diff
@@ -26,10 +26,17 @@
...
 #import "TDPusher.h"
+#import "CDTSessionCookieInterceptor.h"
 #import <OHHTTPStubs/OHHTTPStubs.h>
 #import <OHHTTPStubs/OHHTTPStubsResponse+JSON.h>
 #import <OCMock/OCMock.h>
 
+
+@interface CDTAbstractReplication (Interceptors)
+//expose the interceptor property for inspection.
+@property (nonnull, nonatomic, readwrite, strong) NSArray *httpInterceptors;
```
tomblench: If it only needs to be inspected then it doesn't need to be `readwrite`

CDTDatastore/CDTReplicator/CDTPullReplication.m
rhyshort: I don't think that is possible. The source and target are different types depending on the kind of replication. The obvious solution would be a class method, however that causes the problem of not having access to the iVars and we shouldn't pass the reference to self around until `init` has finished, the other option for this would be to have awkward return type of a dictionary or array, so I think that option is out.  Adding a parameter to the the super classes' init method wouldn't be the right thing to do since it would be stripping creds and not storing the value of the URL somewhere which would be misleading to anyone reading the subclass code in isolation.

Is there another option I may have missed?

CDTDatastoreTests/CDTReplicationTests.m
rhyshort: Fixed in c2d0cfb. It turns out that the interceptors are available on the public interface as readonly anyway

CDTDatastoreTests/CDTReplicationTests.m
```diff
@@ -67,6 +68,39 @@ @interface CDTReplicationTests : CloudantSyncTests <CDTNSURLSessionConfiguration
...
+- (void)testURLCredsReplacedWithCookieInterceptorPull
+{
+    NSError *error;
+    //Doesn't need to be real, we aren't going to actually make a replication.
+    NSURL * remoteUrl = [[NSURL alloc] initWithString:@"http://user:pass@example.com"];
+    CDTDatastore *tmp = [self.factory datastoreNamed:@"test_database" error:&error];
+    CDTPullReplication *pull =
+    [CDTPullReplication replicationWithSource:remoteUrl target:tmp];
+    
+    //check the underlying source to make sure ut doesn't contain the userinfo
```
ricellis: ut -> it

CDTDatastoreTests/CDTReplicationTests.m
```diff
@@ -67,6 +68,39 @@ @interface CDTReplicationTests : CloudantSyncTests <CDTNSURLSessionConfiguration
...
+    XCTAssertEqualObjects([pull.httpInterceptors[0] class], [CDTSessionCookieInterceptor class]);
+}
+
+- (void)testURLCredsReplacedWithCookieInterceptorPush
+{
+    NSError *error;
+    //Doesn't need to be real, we aren't going to actually make a replication.
+    NSURL * remoteUrl = [[NSURL alloc] initWithString:@"http://user:pass@example.com"];
+    CDTDatastore *tmp = [self.factory datastoreNamed:@"test_database" error:&error];
+    CDTPushReplication *pull =
```
ricellis: bit confusing that this is called `pull`

CDTDatastoreTests/CDTReplicationTests.m
```diff
@@ -67,6 +68,39 @@ @interface CDTReplicationTests : CloudantSyncTests <CDTNSURLSessionConfiguration
...
+    [CDTPushReplication replicationWithSource:tmp target:remoteUrl];
+    
+    //check the underlying source to make sure ut doesn't contain the userinfo
```
ricellis: ut -> it

CDTDatastore/CDTReplicator/CDTPullReplication.m
ricellis: I agree that this duplication is unfortunate, a couple of thoughts:

If we add a parameter to the superclass `init` could we just create/add the interceptor there and avoid the confusion (of reading the subclass in isolation) by storing the sanitized `NSURL` in the superclass and then using it as the target or source as appropriate in the subclasses?

or

Could we avoid the awkward dictionary/array return type by just having two separate class methods, one which returns a `CDTSessionCookieInterceptor` and one which returns a sanitized `NSURL`?

CDTDatastoreTests/CDTReplicationTests.m
rhyshort: done in 1244d84

CDTDatastoreTests/CDTReplicationTests.m
rhyshort: fixed in 1244d84

CDTDatastoreTests/CDTReplicationTests.m
rhyshort: done in 1244d84

CDTDatastore/CDTReplicator/CDTPullReplication.m
rhyshort: Using the init would lead to more duplication because the urls are stored as different properties, and they would still need to be stripped of creds. 

Using two sets of methods would give us similar duplication that we have no, since we'd need a method to create the interceptor, which would break the URL into components, and then extract the username and password if provided. The URL method would need to break it into components and then remove the user/password and return the url. 

If this was pre-1.0 we attempt something using the generics, but I think doing that now might result in 1.1 breaking semver, but I'm not sure on that.

rhyshort: I think I also prefer the small amount of duplication we have currently, and for such a small number of lines I think we can get away with it.

ricellis: OK, it is only a few lines so let's leave it.

ricellis:
+1

tomblench:
+1

PR #233, opened 2016-03-22
fixed several potentional null pointer dereferences

tonklon:
the NSError** argument may be nil


rhyshort:
Thanks for catching all these. :smile:  We have your company CLA, so I'm going to go a head and merge.

PR #236, opened 2016-04-08
Deprecate cloudant toolkit integration points

### LINKED ISSUE #235: Deprecate SavedHTTP attachment and -createRevisionFromJson:forDocument:error
CloudantToolkit is deprecated and as a result we should move to deprecate the integration points for toolkit and CDTDatastore.

rhyshort:
## What
Deprecate the integration points for Cloudant ToolKit.

## Why
Cloudant ToolKit is deprecated so the integration points are no longer required, and since we are following [semver](semvar.org) this has be to deprecated rather than removed.

## Reviewers

reviewer @mikerhodes

## Issues

Fixes #235

### DIFF
--- a/CDTDatastore/Attachments/CDTAttachment.h
+++ b/CDTDatastore/Attachments/CDTAttachment.h
@@ -109,6 +109,11 @@
 
 @end
 
+__attribute__((deprecated))
+/**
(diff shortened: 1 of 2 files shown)


mikerhodes:
I think the messages are fine, I'd not expect anyone external to be using them. +1

Issue #231, opened 2016-03-16
Query: use enum for `type` in public API


tomblench:
Using an untyped string for `type` in `CDTQIndexManager` is confusing and error-prone.

- Deprecate methods which take an `NSString` for `index`
- Add enum for `IndexType` with values `json` and `text`
- Replace deprecated methods with new ones taking enum

PR #238, opened 2016-04-13
Move query tests to using enum for creating indexes

rhyshort:
## What
Migrate Query tests to using the APIs which use enums instead of Strings.

## How
Replace all method calls that use strings to denote the index type to using enums where possible

## Reviewers
reviewer @mikerhodes 
## Issues
Part of #231


mikerhodes:
Should https://github.com/cloudant/CDTDatastore/blob/231-index-type-enum-tests/CDTDatastoreTests/CDTQIndexTests.m#L149 be using the CDTQIndexType enums?

rhyshort:
Yes it should, it didn't show up in my deprecated list in xcode, weird

rhyshort:
@mikerhodes done in 3d334b2

mikerhodes:
+1

PR #241, opened 2016-04-22
Correctly generate a Temp dir for TDBlobStore

rhyshort:
## What
Correctly generate a temporary directory for TDBlobStore.

## How
Use `NSTemporyDirectory()` to get a temp dir path.

## Testing
No new tests, encrypted blob store tests now pass on Jenkins with this change.

## Reviewers
reviewer @ricellis

## Notes
The main reason for the change is documented here: http://nshipster.com/nstemporarydirectory/ 
Long story: the current way we are generating tmp dirs is incorrect, the current way is designed for moving files around.


ricellis:
+1

rhyshort:
Although the travis push build job "errored" the test did run successfully. According to the log  attached, so merging.
[travis-log.txt](https://github.com/cloudant/CDTDatastore/files/232244/travis-log.txt)

PR #243, opened 2016-04-26
Fix Warnings

rhyshort:
## What 
Fix numerous compilation warnings.

## How

- Correct String formatting args by casting directly to an int.
- Remove unused TDSocketChangeTracker.h
- Remove MYBlockUtils and use `performSelector`at call sites.
- Ignore some warnings which can be safely ignored using `#pragma` directives

## Reviewers

reviewer: @brynh


CDTDatastore/Encryption/CDTEncryptionKey.m
```diff
@@ -32,7 +32,13 @@ - (NSData *)data {
 }
 
 #pragma mark - Init object
-- (instancetype)init { return [self initWithData:nil]; }
+- (instancetype)init
+{
+#pragma clang diagnostic push
```
brynh: Wouldn't it be easier and clearer to just implement this as `return nil;` as it's marked unavailable?

CDTDatastore/Encryption/CDTEncryptionKeySimpleProvider.m
```diff
@@ -26,7 +26,13 @@ @interface CDTEncryptionKeySimpleProvider ()
 
 @implementation CDTEncryptionKeySimpleProvider
 
-- (instancetype)init { return [self initWithKey:nil]; }
+- (instancetype)init
+{
+#pragma clang diagnostic push
```
brynh: Wouldn't it be easier and clearer to just implement this as `return nil;` as it's marked unavailable?

CDTDatastore/HTTP/CDTURLSessionTask.m
```diff
@@ -203,15 +202,19 @@ - (void) completedThread:(NSThread *)thread {
...
+                                  onThread:thread
+                                withObject:self.requestError
+                             waitUntilDone:NO];
         } else {
-            MYOnThread(thread, ^{
-                [self.delegate receivedResponse:self.response];
-                [self.delegate receivedData:self.requestData];
-            });
-            
+            [self.delegate performSelector:@selector(receivedResponse:)
```
brynh: The behaviour of this is now changed by waiting on the current thread for the `receivedResponse:` to complete and then calling `receivedData` whereas previously they were both called without waiting on the current thread. I think perhaps the method calls:

> [self.delegate receivedResponse:self.response];
> [self.delegate receivedData:self.requestData];

should be put in a separate wrapper method and then called with a single call to `performSelector` that calls the wrapper method on the separate thread and doesn't wait until done.

CDTDatastore/Encryption/CDTEncryptionKey.m
rhyshort: done in d7b1e3f

CDTDatastore/Encryption/CDTEncryptionKeySimpleProvider.m
rhyshort: done in d7b1e3f

CDTDatastore/HTTP/CDTURLSessionTask.m
rhyshort: done in d7b1e3f

brynh:
LGTM.  +1

PR #244, opened 2016-04-26
Implemented `ensureIndexed:` API for `CDTDatastore+query` and `CDTQIndexManager`

### LINKED ISSUE #229: Query: Implement `ensureIndexed:(NSArray *)fieldNames`
Most users don't care about index names and it should be straightforward enough to generate unique ones on the fly.

This is also relevant because it's easy enough for users to specify index names which will cause an error.

rhyshort:
## What
Implement `ensureIndexed:` API.

## How
Generate an index name when `ensureIndexed:` is called.

## Testing
One new test added to make sure the api successfully generates index names.

## Reviewers
reviewer @tomblench 

## Issues 
Fixes #229

### DIFF
--- a/CDTDatastore/query/CDTQIndexManager.m
+++ b/CDTDatastore/query/CDTQIndexManager.m
@@ -214,8 +214,10 @@ static const int VERSION = 2;
  */
 - (NSString *)ensureIndexed:(NSArray * /* NSString */)fieldNames
 {
-    CDTLogError(CDTQ_LOG_CONTEXT, @"-ensureIndexed: not implemented");
-    return nil;
+    uint8_t randBytes[20];
+    arc4random_buf(randBytes, sizeof(randBytes));
+    NSString *indexName = TDHexFromBytes(randBytes, sizeof(randBytes));
+    return [self ensureIndexed:fieldNames withName:indexName];
 }
 
 /**
--- a/CDTDatastore/query/CDTDatastore+Query.h
+++ b/CDTDatastore/query/CDTDatastore+Query.h
@@ -49,20 +49,30 @@ NS_ASSUME_NONNULL_BEGIN
 - (NSDictionary<NSString *, NSArray<NSString *> *> *)listIndexes;
 
 /**
+ Add a single, possibly compound, index for the given field names.
+
+ This function generates a name for the new index.
+
+ @param fieldNames List of fieldnames in the sort format
+ @returns name of created index
+ */
+- (nullable NSString *)ensureIndexed:(NSArray<NSString *> *)fieldNames;
+
+/**
  Create a new index over a set of fields.
- 
+
  Fields in sub-documents can be specified using dotted notation.
- 
+
  An example:
- 
+
      { "name": "mike", "address": { "street": "Any Road" } }
- 
+
  The field name "name" would index "mike", while the field name
  "address.street" would index "Any Road".
- 
+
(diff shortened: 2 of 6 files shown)


CDTDatastore/query/CDTQIndexManager.m
```diff
@@ -214,8 +214,31 @@ - (instancetype)initUsingDatastore:(CDTDatastore *)datastore error:(NSError *__a
...
+    // index name clash if the user is adding indexes from multiple threads.
+
+    NSArray<NSString *> *indexNames = [[self listIndexes] allKeys];
+
+    uint8_t randBytes[20];
+
+    NSString *indexName = nil;
+
+    for (int i = 0; !indexName && i < 200; i++) {
+        arc4random_buf(randBytes, 20);  // not sure I can do this.
```
tomblench: what does the comment `not sure I can do this` mean here?

CDTDatastore/query/CDTQIndexManager.m
```diff
@@ -214,8 +214,31 @@ - (instancetype)initUsingDatastore:(CDTDatastore *)datastore error:(NSError *__a
  */
 - (NSString *)ensureIndexed:(NSArray * /* NSString */)fieldNames
 {
-    CDTLogError(CDTQ_LOG_CONTEXT, @"-ensureIndexed: not implemented");
-    return nil;
+    // So we don't break semvar w.r.t CDTQIndex and nullablity, we need to
```
tomblench: I think it's more a case of "we said we'd get round to implementing this later"

CDTDatastore/query/CDTQIndexManager.m
rhyshort: well that comment should have been deleted, I'm not sure if I can/should just pass in 20 the `arc4random_buf` or use `sizeof` (it's been a while since I've worked with C and I'm a little rusty)

CDTDatastore/query/CDTQIndexManager.m
rhyshort: Well ideally, we generate the index name on the db queue as we do in sync-android removing the risk of a clashing name. However that would mean the semantics around CDTQIndex and indexName would change, it would become nullable where as before it wouldn't have been. So imo the comment lays out why we are doing the generation here and not there. However that being said, do you think this should be removed or improved upon?

tomblench: Well I've thought about this and there is obviously a race condition between generating the index name and then actually using it, so you could still have a collision without your code detecting it. Now of course we are only talking about a 2^160 chance of this happening. So maybe we just don't bother checking for collision?

CDTDatastore/query/CDTQIndexManager.m
tomblench: You should probably use `sizeof` to avoid repeating the `20` magic constant. If you were paranoid you'd also divide by `sizeof(uint8_t)` but you can guarantee that this type will be one byte so it's not needed.

CDTDatastore/query/CDTQIndexManager.m
rhyshort: removed collision checking in 24966d8

CDTDatastore/query/CDTQIndexManager.m
rhyshort: done in 24966d8

PR #245, opened 2016-04-28
Enable usage of special characters in index names.

### LINKED ISSUE #230: Query: special characters in index names
In the various `[CDTQIndexManager ensureIndexed]` variants, the `indexName` argument is used to dynamically build a table name via string concatenation. This is done because SQL placeholders cannot be used to form table names. 

As a consequence, characters which have special meaning to the SQL query parser (such as space or semicolon and a whole host of others) cannot be used for index names, because they cause the SQL parser to throw an error.

Potential fixes:
- At a minimum this restriction should be documented
- We should also consider mitigating strategies, like restricting the index name to a "safe" range of characters

rhyshort:
## What
Enable the usage of special characters in index names such as ` `.

## How
Quote SQL Identifiers in SQL Statements

## Tests
Added new tests to make sure indexes are escaped
Modified existing tests to match the generated SQL

## Reviewers

reviewer: @alfinkel
reviewer: @emlaver

## Issues

Fixes #230

### DIFF
--- a/CDTDatastore/query/CDTQQuerySqlTranslator.m
+++ b/CDTDatastore/query/CDTQQuerySqlTranslator.m
@@ -99,7 +99,7 @@
 
         if (allDocsIndex.length > 0) {
             NSString *tableName = [CDTQIndexManager tableNameForIndex:allDocsIndex];
-            NSString *sql = [NSString stringWithFormat:@"SELECT _id FROM %@;", tableName];
+            NSString *sql = [NSString stringWithFormat:@"SELECT _id FROM \"%@\";", tableName];
             sqlNode.sql = [CDTQSqlParts partsForSql:sql parameters:@[]];
         }
 
@@ -545,10 +545,9 @@
                                    fieldName,
                                    sqlOperator,
                                    operand];
-    NSString *subSelect = [NSString stringWithFormat:@"SELECT _id FROM %@ WHERE %@",
-                           tableName,
-                           whereForSubSelect];
-    
+    NSString *subSelect = [NSString
+        stringWithFormat:@"SELECT _id FROM \"%@\" WHERE %@", tableName, whereForSubSelect];
+
     return [NSString stringWithFormat:@"_id NOT IN (%@)", subSelect];
 }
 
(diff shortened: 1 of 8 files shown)


CDTDatastore/query/CDTQQuerySqlTranslator.m
```diff
@@ -99,7 +99,7 @@ + (CDTQQueryNode *)translateQuery:(NSDictionary *)query
 
         if (allDocsIndex.length > 0) {
             NSString *tableName = [CDTQIndexManager tableNameForIndex:allDocsIndex];
-            NSString *sql = [NSString stringWithFormat:@"SELECT _id FROM %@;", tableName];
+            NSString *sql = [NSString stringWithFormat:@"SELECT _id FROM '%@';", tableName];
```
alfinkel: Is there a reason why you're using `'%@'` here instead of `\"%@\"` as you are doing elsewhere?

CDTDatastore/query/CDTQQuerySqlTranslator.m
```diff
@@ -545,10 +545,9 @@ + (NSString *)whereClauseForNot:(NSString *)fieldName
                                    fieldName,
                                    sqlOperator,
                                    operand];
-    NSString *subSelect = [NSString stringWithFormat:@"SELECT _id FROM %@ WHERE %@",
-                           tableName,
-                           whereForSubSelect];
-    
+    NSString *subSelect =
+        [NSString stringWithFormat:@"SELECT _id FROM '%@' WHERE %@", tableName, whereForSubSelect];
```
alfinkel: Same as https://github.com/cloudant/CDTDatastore/pull/245/files#r61451170

CDTDatastore/query/CDTQQuerySqlTranslator.m
rhyshort: ah it should be `\"%@/"` fixed in 3d4a4ee

CDTDatastore/query/CDTQQuerySqlTranslator.m
rhyshort: fixed in 3d4a4ee

alfinkel:
+1

emlaver:
+1

PR #246, opened 2016-05-05
Implement bulk get

tomblench:
_What_: Use `_bulk_get` endpoint in `TDPuller` if the remote database supports it. Also clean up TDPuller to initialise various arrays in `init`.

_Testing_: Existing tests and RA tests pass against couch 1.6 and 2.0

reviewer: @rhyshort

### DIFF
--- a/CDTDatastore/touchdb/TDPuller.m
+++ b/CDTDatastore/touchdb/TDPuller.m
@@ -55,10 +55,62 @@ static NSString* joinQuotedEscaped(NSArray* strings);
 
 @implementation TDPuller
 
+
+- (instancetype)initWithDB:(TD_Database*)db
+                    remote:(NSURL*)remote
+                      push:(BOOL)push
+                continuous:(BOOL)continuous
+              interceptors:(NSArray*)interceptors
+{
+    if (self = [super initWithDB:db remote:remote push:push continuous:continuous interceptors:interceptors])
+    {
+        NSUInteger initialRevsCapacity = 100;
+        _deletedRevsToPull = [[NSMutableArray alloc] initWithCapacity:initialRevsCapacity];
+        _revsToPull = [[NSMutableArray alloc] initWithCapacity:initialRevsCapacity];
+        _bulkGetRevs = [[NSMutableArray alloc] initWithCapacity:initialRevsCapacity];
+        _bulkRevsToPull = [[NSMutableArray alloc] initWithCapacity:initialRevsCapacity];
+    }
+    return self;
+}
+
 - (void)dealloc { [_changeTracker stop]; }
+
 
 - (void)beginReplicating
 {
+    // _bulk_get check starts an async task
+    [self asyncTaskStarted];
+
+    __weak TDPuller* weakSelf = self;
+    // check to see if _bulk_get endpoint is supported and then start replication
+    [self sendAsyncRequest:@"GET"
+                      path:@"_bulk_get"
+                      body:nil
+              onCompletion:^(id result, NSError* error) {
+
+                  __strong TDPuller* strongSelf = weakSelf;
+
+                  switch(error.code) {
+                      case 404:
+                          // not found: _bulk_get not supported
+                          [strongSelf setBulkGetSupported:false];
+                          CDTLogInfo(CDTREPLICATION_LOG_CONTEXT, @"%@ Remote database does not support _bulk_get", self);
+                          break;
+                      case 405:
+                          // method not allowed: this endpoint exists, we called with the wrong method
+                          [strongSelf setBulkGetSupported:true];
+                          CDTLogInfo(CDTREPLICATION_LOG_CONTEXT, @"%@ Remote database supports _bulk_get", self);
+                          break;
+                  }
+                  
+                  // on completion...start the actual replication
+                  [strongSelf beginReplicatingInternal];
+              }];
+}
+
+- (void)beginReplicatingInternal
+{
+    
     if (!_downloadsToInsert) {
         // Note: This is a ref cycle, because the block has a (retained) reference to 'self',
         // and _downloadsToInsert retains the block, and of course I retain _downloadsToInsert.
@@ -78,7 +130,6 @@ static NSString* joinQuotedEscaped(NSArray* strings);
     }
 
     _caughtUp = NO;
-    [self asyncTaskStarted];  // task: waiting to catch up
     [self startChangeTracker];
 }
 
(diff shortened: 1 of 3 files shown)


CDTDatastore/touchdb/TDPuller.m
```diff
@@ -139,9 +190,10 @@ - (void)stop
             [self asyncTasksFinished:1];  // balances -asyncTaskStarted in -beginReplicating
     }
     _changeTracker = nil;
-    _revsToPull = nil;
-    _deletedRevsToPull = nil;
-    _bulkRevsToPull = nil;
+    [_revsToPull removeAllObjects];
+    [_deletedRevsToPull removeAllObjects];
+    [_bulkRevsToPull removeAllObjects];
+    [_bulkGetRevs removeAllObjects];
```
rhyshort: Why have these gone from being nil to reset to emptying them?

tomblench: See the new init method and the removal of lines like
```
-            if (!_bulkRevsToPull) _bulkRevsToPull = [[NSMutableArray alloc] initWithCapacity:100];
```

CDTDatastore/touchdb/TDPuller.m
```diff
@@ -398,8 +462,62 @@ - (void)pullRemoteRevision:(TD_Revision*)rev
...
+                              [remainingRevs removeObjectAtIndex:pos];
+                              [_downloadsToInsert queueObject:rev];
+                              [self asyncTaskStarted];
+                          }
+                          
+                      }
+                  }
+                  
+                  [self asyncTasksFinished:1];
+                  --_httpConnectionCount;
```
rhyshort: why prefix instead of postfix?

tomblench: There is some ancient history (http://www.idinews.com/peeves/prefixIncr.html, http://programmers.stackexchange.com/questions/59880/avoid-postfix-increment-operator) but if the result isn't being assigned to it doesn't make any difference.

More prosaically, it's a prefix decrement because I copied it from the other code in this file - you can see two other instances of it.

CDTDatastore/touchdb/TDPuller.m
rhyshort: Ah gotcha cool.

CDTDatastore/touchdb/TDPuller.m
rhyshort: Ok, i'm keen on it since its the same as the postfix, but since it does match the rest of the file, it's probably better stylisticly to go with it.

CDTDatastore/touchdb/TDPuller.m
```diff
@@ -398,8 +462,62 @@ - (void)pullRemoteRevision:(TD_Revision*)rev
...
+    NSDictionary *requestBody = @{@"docs": keys};    
+    NSMutableArray* remainingRevs = [bulkRevs mutableCopy];
+
+    [self sendAsyncRequest:@"POST"
+                      path:@"_bulk_get?revs=true&attachments=true"
+                      body:requestBody
+              onCompletion:^(id result, NSError* error) {
+
+                  // unpack revisions and queue them in _downloadsToInsert
+                  NSArray *results = result[@"results"];
```
rhyshort: Should we be crashing in the event that result isn't a dict / `nil`? ( I know it is unlikely, I'm just wondering what your thoughts on this are)

tomblench: @rhyshort see 588a47d - i think this is the only `nil` check needed as the `for ... in` constructs are inherently nil-safe and will just drop out

rhyshort: hmmm that doesn't alleviate my concern about what if `result`  isn't a `NSDictionary` or `nil` is there some code further up that ensures this?

CDTDatastore/touchdb/TDPuller.m
```diff
@@ -303,42 +354,55 @@ - (void)processInbox:(TD_RevisionList*)inbox
...
-            // Rather than pulling a single revision in 'bulk', just pull it normally:
-            [self queueRemoteRevision:_bulkRevsToPull[0]];
-            [_bulkRevsToPull removeObjectAtIndex:0];
-            nBulk = 0;
-        }
-        if (nBulk > 0) {
-            // Prefer to pull bulk revisions:
+        
+        // _bulkGetRevs count != 0 means _bulk_get endpoint supported
+        if ([_bulkGetRevs count] !=0) {
```
brynh: Can't we just check `_bulkGetSupported`? We could then do away with the comment too.

CDTDatastore/touchdb/TDPuller.m
brynh: This does appear to be `nil` safe, but I agree that it could crash if `result` isn't an `NSDictionary`. Also, we don't seem to be handling any non-nil `error` passed into this completion block.

CDTDatastore/touchdb/TDPuller.m
```diff
@@ -398,15 +460,83 @@ - (void)pullRemoteRevision:(TD_Revision*)rev
...
+                      body:requestBody
+              onCompletion:^(id result, NSError* error) {
+                  __strong TDPuller* strongSelf = weakSelf;
+                  if (error) {
+                      strongSelf.error = error;
+                      [strongSelf revisionFailed];
+                      strongSelf.changesProcessed+=bulkRevs.count;
+                  } else {
+                      // unpack revisions and queue them in _downloadsToInsert
+                      NSArray *results = $castIf(NSArray, result[@"results"]);
```
rhyshort: Do we know for sure `result` will be a `NSDictionary/nil`?

tomblench: OK good spot - I missed this one

tomblench: @rhyshort see 863e9f4 - it's ugly but at least it won't crash out.

CDTDatastore/touchdb/TDPuller.m
tomblench: @brynh thanks for spotting, see a51d730

CDTDatastore/touchdb/TDPuller.m
rhyshort: ok, so the `result` object that is passed in to the completion is always going to be an `NSDictionary` or `nil`?

tomblench: @rhyshort see 5a97dd1 which also fixes another mistake

brynh:
+1

rhyshort:
+1

PR #249, opened 2016-05-18
Fix nullability annotations

brynh:
Note: It may be easier to review this PR by commit since the first commit makes all the changes to nullability, but when combined with the second commit, which changes some files to use NS_ASSUME_NONNULL_BEGIN/END it loses some clarity over which things have changed their nullability.

## What
Correct nullability annotations.

## Why
When nullability annotations were originally added, there was a misconception that `NSObject`'s `init` method could sometime return `nil`. This lead to some methods being incorrectly annotated as `nullable` when they should be `nonnull`.

## How
The code has been reinspected and the annotations updated to reflect the nullability.

## Testing
There are no additional tests added, but existing tests that pass `nil` to method arguments annotated as `nonnull` have been updated to eliminate the warnings generated by the compiler. Since Objective-C doesn't enforce that `nonnull` arguments are only passed non-null objects, these tests are still legitimate and we may safely hide the warnings.

## Reviewers
reviewer: @rhyshort

### DIFF
--- a/CDTDatastore/Attachments/CDTAttachment.h
+++ b/CDTDatastore/Attachments/CDTAttachment.h
@@ -20,6 +20,8 @@
 
 #import "CDTBlobReader.h"
 
+NS_ASSUME_NONNULL_BEGIN
+
 /**
  Base class for attachments in the datastore.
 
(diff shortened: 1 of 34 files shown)


rhyshort:
+1

PR #250, opened 2016-05-18
Fix semantic issue warnings

brynh:
## What
Fix Xcode warnings for "Semantic Issue".

## Why
Having many "Semantic Issue" warnings means there could potentially be important semantic issues that we are warned about but go unnoticed.

## How
All the warnings of type "Semantic Issue" have been rectified so they're no longer generating warnings.

## Testing
No additional tests required.

## Reviewers
reviewer: @rhyshort


rhyshort:
+1

PR #248, opened 2016-05-16
ReplicationAcceptance fixes:

tomblench:
_What_

- Make nDocs and largeRevTreeSize configurable
- Make logging configurable
- Increase request timeout on NSURLSessionConfiguration to 300 seconds
- Limit number of concurrent NSURLRequests by semaphore counting - this was previously harming performance and causing timeouts

_Why_

Make it easier to quickly run ReplicationAcceptance (with small numbers of `nDocs`) without having to recompile. Also make diagnosis of failures easier by outputting logging messages depending on configuration.

The increase timeout value is necessary for large replications to succeed because of the threaded nature of `TDPuller` and `TDPusher` and the large number of `NSURLRequests` they start.

_Testing_

ReplicationAcceptance tests pass.

reviewer @rhyshort 
reviewer @brynh

PR #253, opened 2016-05-31
Experiment with global semaphore

tomblench:
_What_: Use a global semaphore to limit number of simultaneous `NSURLRequest`s

_Why_: #248 introduced a semaphore on a per `CDTURLSession` basis, but some `ReplicationAcceptance` tests were still failing. On closer analysis, the failing tests were those which run multiple simultaneous replications. This defeats the purpose of the semaphore because the code now makes `n*limit` simultaneous networks requests where `n` is the number of simultaneous replications.

_Testing_:
- Existing tests pass
- Added test `testSemaphoreCountsCorrectly` to ensure that the semaphore state is correctly maintained when a large number of requests are made and half of them are cancelled. (If the semaphore counted incorrectly it could cause requests to hang incorrectly or for too many simultaneous requests to be made).

reviewer: @rhyshort 
reviewer: @brynh

### DIFF
--- a/CDTDatastore/HTTP/CDTURLSession.h
+++ b/CDTDatastore/HTTP/CDTURLSession.h
@@ -26,7 +26,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@interface CDTURLSession : NSObject <NSURLSessionDataDelegate>
+@interface CDTURLSession : NSObject <NSURLSessionDataDelegate, NSURLSessionTaskDelegate>
 
 /**
  * Initalises a CDTURLSession without a delegate and an empty array of interceptors. Calling this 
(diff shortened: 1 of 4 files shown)


CDTDatastore/HTTP/CDTURLSession.h
```diff
@@ -26,7 +26,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@interface CDTURLSession : NSObject <NSURLSessionDataDelegate>
+@interface CDTURLSession : NSObject <NSURLSessionDataDelegate, NSURLSessionDelegate>
```
brynh: I think this should be ` NSURLSessionTaskDelegate` instead of `NSURLSessionDelegate`.

tomblench: Thanks for spotting, fixed in 1366aa3

brynh:
+1

rhyshort:
+1

PR #254, opened 2016-06-01
Warnings fix

brynh:
**Note:** It may be easier to review this PR commit by commit as each commit fixes a particular warning or group of related warnings.

## What
Correct compiler warnings.

## Why
Having many compiler warnings means there could potentially be important issues that we are warned about that go unnoticed because of the sheer number of things we are being warned about.

## How
Each warning has been examined to see if it represents a real issue we need to fix or whether it can be safely ignored.  In the majority of cases the code or project setup has been fixed to eliminate the warnings. 

The cases where warnings are ignored are:
- calls to deprecated operations - these have been ignored because either: the code calling the deprecated operation is itself deprecated; or the code is deliberately designed to test an operation that has been deprecated in CDTDatastore and we still want to ensure it works until we remove it.
- passing `nil` to a method that specifies a `nonnull` argument. This has been ignored only in test cases because Objective C does not enforce that a `nil` cannot be sent to an operation in an argument specified as `nonnull`, so it is a valid test to ensure that `nil` is handled sensibly.

## Testing
No additional tests required.  Existing tests still pass.

### DIFF
--- a/CDTDatastore.xcodeproj/project.pbxproj
+++ b/CDTDatastore.xcodeproj/project.pbxproj
@@ -1182,40 +1182,12 @@
 /* End PBXBuildRule section */
 
 /* Begin PBXContainerItemProxy section */
-		9873857D1C47BB5D00937212 /* PBXContainerItemProxy */ = {
-			isa = PBXContainerItemProxy;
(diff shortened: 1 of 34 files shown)


rhyshort:
it looks good to me, lets get this merged +1

PR #255, opened 2016-06-02
Fix sample project warnings

brynh:
## What
Correct compiler warnings in the sample project

## Why
Having many compiler warnings means there could potentially be important issues that we are warned about that go unnoticed because of the sheer number of things we are being warned about.

## How
Each warning has been examined and fixed.

## Testing
No additional tests required.


rhyshort:
+1

Issue #251, opened 2016-05-20
-CDTReplicator:replicatorStopped:n should set error


tomblench:
When this method sets the current state to `CDTReplicatorStateError` it should also set the error object by re-wrapping the underlying TDReplicator error.

rhyshort:
Fix merged to master closing.

PR #259, opened 2016-06-16
Copy error pointer from TDReplicator to CDTReplicator

tomblench:
See #251


CDTDatastore/CDTReplicator/CDTReplicator.m
```diff
@@ -373,6 +373,8 @@ - (void)replicatorStopped:(NSNotification *)n
                     strongSelf.state = CDTReplicatorStateError;
                 } else {
                     strongSelf.state = CDTReplicatorStateStopped;
+                    // copy underlying error pointer
+                    strongSelf.error = strongSelf.tdReplicator.error;
```
rhyshort: this should be in the block above, this is triggered if there isn't an error.

tomblench: Thanks for spotting, fixed in aa83762

brynh:
+1

PR #283, opened 2016-06-20
Fix bulk get when there are deleted revs

tomblench:


CDTDatastoreReplicationAcceptanceTests/ReplicationAcceptance.m
```diff
@@ -1104,6 +1104,40 @@ -(void) testPullDeleteAllPush
...
+ */
+-(void) testPushDeleteSomePull
+{
+    // Create docs in remote database
+    NSLog(@"Creating local documents...");
+    [self createLocalDocs:self.n_docs];
+    [self pushToRemote];
+    XCTAssertEqual(self.n_docs, self.datastore.documentCount, @"Incorrect number of documents created");
+    
+    // delete half the remote docs
```
brynh: `half` -> `some of`

CDTDatastoreReplicationAcceptanceTests/ReplicationAcceptance.m
```diff
@@ -1104,6 +1104,40 @@ -(void) testPullDeleteAllPush
...
+        NSString *docId = [NSString stringWithFormat:@"doc-%i", i];
+        [self deleteRemoteDocWithId:docId];
+    }
+    
+    NSLog(@"Pull replicate...");
+    // Replicate the changes
+    [self pullFromRemote];
+    
+    NSLog(@"Compare...");
+    BOOL same = [self compareDatastore:self.datastore
```
brynh: I think it would add a little value to also do this comparison before the `pullFromRemote`, and then assert that they're different, although I guess we have other tests that test that delete and pull do the right things, so it's not essential.

brynh:
+1

PR #292, opened 2016-06-24
Implement backoff and retry in response to 429 error code

tomblench:
_What_: Wait an exponentially increasing time on receipt of each `429 Too Many Requests` error, then set `context.shouldRetry` to `true` to re-execute the request.

_Why_: From https://tools.ietf.org/html/rfc6585#section-4 
```
   The 429 status code indicates that the user has sent too many
   requests in a given amount of time ("rate limiting").
```

_Testing_: New RA test `test429Retry`. Add small class `SimpleHttpServer` to be used for sending `404` responses for existing test `testFiltersWithChangesFeed` and for sending `429` responses for new test.

### DIFF
--- a/CDTDatastore/HTTP/CDTHTTPInterceptorContext.h
+++ b/CDTDatastore/HTTP/CDTHTTPInterceptorContext.h
@@ -25,6 +25,15 @@ NS_ASSUME_NONNULL_BEGIN
 @property (readwrite, nonatomic, strong) NSMutableURLRequest *request;
 @property (nonatomic) BOOL shouldRetry;
 @property (nullable, readwrite, nonatomic, strong) NSHTTPURLResponse *response;
+/**
+ * For storing arbitrary per-context state
+ * NOTE: Users are strongly encouranged to use unique keys by ensuring keys are prefixed, eg
+ * com.mycompany.MyInterceptor.foo, com.mycompany.MyInterceptor.bar, where:
+ * - com.company is the reversed internet domain name
+ * - MyInterceptor is the name of the interceptor class
+ * - foo and bar describe the values being stored.
+ */
+@property (readwrite, nonatomic, strong) NSMutableDictionary<NSString*, NSObject*> *state;
 
 /**
  *  Unavaiable, use -initWithRequest
@@ -37,9 +46,24 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  *  Initalizes a CDTURLSessionInterceptorContext
  *
+ *  This is equivalent to calling initWithRequest:request:[NSMutableDictionary dictionary]
+ *
  *  @param request the request this context should represent
+ *
  **/
-- (instancetype)initWithRequest:(NSMutableURLRequest *)request NS_DESIGNATED_INITIALIZER;
+- (instancetype)initWithRequest:(NSMutableURLRequest *)request;
+
+/**
+ *  Initalizes a CDTURLSessionInterceptorContext
+ *
+ *  NOTE: electing not to copy the state of a previous interceptor may cause issues (especially for 
+ *  interceptors which share state between Request and Response contexts).
+ *
+ *  @param request the request this context should represent
+ *  @param state the initial state of the interceptor pipeline, as key/value pairs
+ **/
+- (instancetype)initWithRequest:(NSMutableURLRequest *)request
+                          state:(NSMutableDictionary *)state NS_DESIGNATED_INITIALIZER;
 
 @end
 
--- a/CDTDatastore/HTTP/CDTRequestLimitInterceptor.h
+++ b/CDTDatastore/HTTP/CDTRequestLimitInterceptor.h
@@ -1,0 +1,22 @@
+//
+//  CDTRequestLimitInterceptor.h
+//  CDTDatastore
+//
+//  Created by tomblench on 23/06/2016.
+//  Copyright © 2016 IBM Corporation. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+#import "CDTSessionCookieInterceptor.h"
+#import "CDTLogging.h"
+
+
+
+@interface CDTRequestLimitInterceptor : NSObject <CDTHTTPInterceptor>
+
++ (nonnull instancetype)interceptor;
+- (nonnull instancetype)init;
+- (nonnull instancetype)initWithSleep:(NSTimeInterval)sleep
+                           maxRetries:(int)maxRetries NS_DESIGNATED_INITIALIZER;
+
+@end
--- a/CDTDatastore/HTTP/CDTRequestLimitInterceptor.m
+++ b/CDTDatastore/HTTP/CDTRequestLimitInterceptor.m
@@ -1,0 +1,85 @@
+//
+//  CDTRequestLimitInterceptor.m
+//  CDTDatastore
+//
+//  Created by tomblench on 23/06/2016.
+//  Copyright © 2016 IBM Corporation. All rights reserved.
+//
+
+#import "CDTRequestLimitInterceptor.h"
+
+static NSString *kSleepKey = @"com.cloudant.CDTRequestLimitInterceptor.sleep";
+static NSString *kRetryCountKey = @"com.cloudant.CDTRequestLimitInterceptor.retryCount";
+
+@interface CDTRequestLimitInterceptor ()
+
+// the initial time to sleep on receipt of a 429
+@property (readonly) NSTimeInterval initialSleep;
+
+// the maximum number of retries (after original request) to make on receipt of a
+// 429
+@property (readonly) int maxRetries;
+
+@end
+
+@implementation CDTRequestLimitInterceptor
+
++ (instancetype)interceptor
+{
+    return [[CDTRequestLimitInterceptor alloc] init];
+}
+
+- (instancetype)init
+{
+    self = [self initWithSleep:0.25 // 250ms
+                     maxRetries:3];
+    return self;
+}
+
+- (instancetype)initWithSleep:(NSTimeInterval)initialSleep
+                  maxRetries:(int)maxRetries;
+{
+    if (self = [super init]) {
+        _initialSleep = initialSleep;
+        _maxRetries = maxRetries;
+    }
+    return self;
+}
+
+/**
+ * Interceptor to retry after an exponential backoff if we receive a 429 error
+ */
+- (CDTHTTPInterceptorContext *)interceptResponseInContext:(CDTHTTPInterceptorContext *)context
+{
+    if (context.response.statusCode == 429) {
+
+        // if we are the first invocation in this pipeline, set some state
+        if (!context.state[kSleepKey]) {
+            [context.state setValue:@(self.initialSleep) forKey:kSleepKey];
+        }
+        if (!context.state[kRetryCountKey]) {
+            [context.state setValue:@0 forKey:kRetryCountKey];
+        }
+        
+        double sleep = [(NSNumber*)context.state[kSleepKey] doubleValue];
+        int retryCount = [(NSNumber*)context.state[kRetryCountKey] intValue];
+
+        if (retryCount < self.maxRetries) {
+            CDTLogInfo(CDTTD_REMOTE_REQUEST_CONTEXT, @"429 error code (too many requests) received. "
+                       "Will retry in %@ seconds.", context.state[@"sleep"]);
+            
+            // sleep for a short time before making next request
+            [NSThread sleepForTimeInterval:sleep];
+            [context.state setValue:@(sleep*2) forKey:kSleepKey]; // exponential back-off
+            [context.state setValue:@(retryCount+1) forKey:kRetryCountKey];
+            context.shouldRetry = true;
+        } else {
+            CDTLogWarn(CDTTD_REMOTE_REQUEST_CONTEXT, @"Maximum number of retries (%d) exceeded in "
+                       "CDTRequestLimitInterceptor.", self.maxRetries);
+        }
+
+    }
+    return context;
+}
+
+@end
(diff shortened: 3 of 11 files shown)


CDTDatastore/HTTP/CDTRequestLimitInterceptor.m
```diff
@@ -0,0 +1,46 @@
...
+}
+
+/**
+ * Interceptor to retry after an exponential backoff if we receive a 429 error
+ */
+- (CDTHTTPInterceptorContext *)interceptResponseInContext:(CDTHTTPInterceptorContext *)context
+{
+    if (context.response.statusCode == 429) {
+        
+        CDTLogInfo(CDTTD_REMOTE_REQUEST_CONTEXT, @"429 error code received. Will retry in %f seconds.", self.sleep);
```
ricellis: I think this log message will be misleading on the last attempt since I think it will get logged anyway declaring that there will be a retry in x seconds, but the outer retry control at https://github.com/cloudant/CDTDatastore/blob/master/CDTDatastore/HTTP/CDTURLSessionTask.m#L201 will prevent another request being made. I guess that was one advantage of having the number of retries available in the Java interceptor that I had forgotten about when we discussed it.

Further to that I wonder if it would be helpful to include the "too many requests" to save people needing to look it up if they are unfamiliar with the status code.

CDTDatastoreTests/CDTReplicationTests.m
```diff
@@ -138,10 +175,41 @@ - (void)testURLCredsReplacedWithCookieInterceptorPush
...
+    CDTReplicator *replicator = [replicatorFactory oneWay:pull error:&error];
+    
+    dispatch_group_t taskGroup = dispatch_group_create();
+    [replicator startWithTaskGroup:taskGroup error:&error];
+    
+    dispatch_group_wait(taskGroup, DISPATCH_TIME_FOREVER);
+
+    // after 10 retries the sleep time should equal 512:
+    // 250ms * (2^11)
+    XCTAssertEqual(512, ((CDTRequestLimitInterceptor*)(replicator.cdtReplication.httpInterceptors[0])).sleep);
```
ricellis: Doesn't this mean the test is going to take ~9 minutes? It's probably not necessary to wait for the entire time to validate the code is working. In Java we shortened the initial wait/ number of retries to make the tests faster.

tomblench: @ricellis good point, and since the test completes much faster, something must be broken somewhere

ricellis:
Do you know if we log anywhere when the max number of retries has been reached? I didn't see it in https://github.com/cloudant/CDTDatastore/blob/master/CDTDatastore/HTTP/CDTURLSessionTask.m, but I think it would be good to add.

ricellis:
I also think it would be valuable to have an additional test using the PushReplicator to serve as a regression in the event someone removes the interceptor.
There may also be value in a test that asserts that the replays stop once a  non-429 is received (whilst I think it is obvious from the code that it will work, it could prevent regressions in future). Although I guess that is a lot of extra bother with the `SimpleHttpServer`.

CDTDatastore/HTTP/CDTRequestLimitInterceptor.m
```diff
@@ -0,0 +1,46 @@
...
+        
+        // sleep for a short time before making next request
+        [[NSRunLoop currentRunLoop] runMode: NSDefaultRunLoopMode
+                                 beforeDate: [NSDate dateWithTimeIntervalSinceNow:self.sleep]];
+        self.sleep *= 2; // exponential back-off
+        context.shouldRetry = true;
```
ricellis: Should this be `YES`?

ricellis: nvm I followed it up and see it is `BOOL`

tomblench: apparently `YES` is more idiomatic so you have a point but it compiles to the same thing.

CDTDatastoreTests/CDTReplicationTests.m
```diff
@@ -63,54 +73,80 @@ - (CDTHTTPInterceptorContext *)interceptRequestInContext:(CDTHTTPInterceptorCont
...
+                port:(int)port
+{
+    if (self = [super init]) {
+        self.header = header;
+        self.port = port;
+    }
+    return self;
+}
 
 // Start a simple HTTP server on localhost that responds to any message with a "404 Not Found".
```
brynh: This comment is now incorrect.

CDTDatastoreTests/CDTReplicationTests.m
```diff
@@ -63,54 +73,80 @@ - (CDTHTTPInterceptorContext *)interceptRequestInContext:(CDTHTTPInterceptorCont
...
             if (connfd > 0) {
                 char buffer[buf_size];
                 bzero(buffer, buf_size);
-
+                
                 // Receive a message.
                 recv(connfd, buffer, buf_size, 0);
-
+                
                 // We don't care what the message was (or if we read it all), just send back a 404.
```
brynh: This comment is now incorrect.

CDTDatastore/HTTP/CDTRequestLimitInterceptor.m
```diff
@@ -50,16 +49,32 @@ - (instancetype)initWithSleep:(NSTimeInterval)sleep
  */
 - (CDTHTTPInterceptorContext *)interceptResponseInContext:(CDTHTTPInterceptorContext *)context
 {
-    if (context.response.statusCode == 429 && self.maxRequests > 0) {
+    if (context.response.statusCode == 429) {
+
+        // if we are the first invocation in this pipeline, set some state
+        if (!context.state[@"sleep"]) {
```
rhyshort: We should prefix the key for state so it looks something like: `CDTRequestLimitInterceptor.xxx`

tomblench: Agreed, or instead of exposing the dictionary I expose `getValue` and `setValue` which operate on the underlying dictionary and use some sneaky obj-c magic (if possible) to automatically prepend with the interceptor class name

CDTDatastore/HTTP/CDTHTTPInterceptorContext.h
```diff
@@ -25,6 +25,11 @@ NS_ASSUME_NONNULL_BEGIN
 @property (readwrite, nonatomic, strong) NSMutableURLRequest *request;
 @property (nonatomic) BOOL shouldRetry;
 @property (nullable, readwrite, nonatomic, strong) NSHTTPURLResponse *response;
+/**
+ * For storing arbitrary per-context state
+ * NOTE: Users are strongly encouranged to use unique keys
+ */
+@property (readwrite, nonatomic, strong) NSMutableDictionary *state;
```
rhyshort: We should also use generics here for those users who are using swift. I think `NSMutableDictionary<NSString*,NSObject*> *state` should do it.

CDTDatastore/HTTP/CDTRequestLimitInterceptor.m
rhyshort: I think that might make it bit complicated when we can just go with prefixing instead. Which is easier and something Objc users are used to. It is a bit more difficult for swift since prefixing isn't normally a thing, but I'm sure they can adapt.

CDTDatastore/HTTP/CDTRequestLimitInterceptor.m
```diff
@@ -0,0 +1,86 @@
...
+        if (!context.state[kRequestCountKey]) {
+            [context.state setValue:@0 forKey:kRequestCountKey];
+        }
+        
+        double sleep = [(NSNumber*)context.state[kSleepKey] doubleValue];
+        int requestCount = [(NSNumber*)context.state[kRequestCountKey] intValue];
+
+        if (requestCount >= self.maxRequests) {
+            // retries exceeded, exit early
+            // TODO log
```
ricellis: Add the log message.

CDTDatastore/HTTP/CDTRequestLimitInterceptor.m
```diff
@@ -0,0 +1,86 @@
...
+            [context.state setValue:@(self.initialSleep) forKey:kSleepKey];
+        }
+        if (!context.state[kRequestCountKey]) {
+            [context.state setValue:@0 forKey:kRequestCountKey];
+        }
+        
+        double sleep = [(NSNumber*)context.state[kSleepKey] doubleValue];
+        int requestCount = [(NSNumber*)context.state[kRequestCountKey] intValue];
+
+        if (requestCount >= self.maxRequests) {
```
ricellis: As discussed there is a slight gotcha here in that if the `self.maxRequests` is greater than the initial value of `_remainingRetries` in `CDTURLSessionTask` then this interceptor could log that it will retry, but the retry will never execute because `_remainingRetries` would be exhausted.
This could look a bit confusing in the logs, but I guess the alternative is also checking the number of remaining retries on the session task which means exposing that additional state too.

CDTDatastore/HTTP/CDTRequestLimitInterceptor.h
```diff
@@ -0,0 +1,22 @@
+//
+//  CDTRequestLimitInterceptor.h
+//  CDTDatastore
+//
+//  Created by tomblench on 23/06/2016.
+//  Copyright © 2016 IBM. All rights reserved.
```
ricellis: needs to say `IBM Corp.` or `IBM Corporation`.

CDTDatastore/HTTP/CDTRequestLimitInterceptor.m
```diff
@@ -0,0 +1,86 @@
+//
+//  CDTRequestLimitInterceptor.m
+//  CDTDatastore
+//
+//  Created by tomblench on 23/06/2016.
+//  Copyright © 2016 IBM. All rights reserved.
```
ricellis: needs to say `IBM Corp.` or `IBM Corporation`.

CDTDatastoreTests/CDTReplicationTests.m
```diff
@@ -123,6 +195,7 @@ - (void)testURLCredsReplacedWithCookieInterceptorPull
     // check the underlying source to make sure it doesn't contain the userinfo
     // and check that the interceptors list contains the cookie interceptor.
     XCTAssertEqualObjects(@"http://example.com", pull.source.absoluteString);
+    // 2 interceptors - because the 429 backoff interceptor is also present
     XCTAssertEqual(pull.httpInterceptors.count, 1);
```
ricellis: Comment says 2 interceptors, but the assert checks for 1.

CDTDatastoreTests/CDTReplicationTests.m
```diff
@@ -138,20 +211,86 @@ - (void)testURLCredsReplacedWithCookieInterceptorPush
     // check the underlying source to make sure it doesn't contain the userinfo
     // and check that the interceptors list contains the cookie interceptor.
     XCTAssertEqualObjects(@"http://example.com", push.target.absoluteString);
+    // 2 interceptors - because the 429 backoff interceptor is also present
```
ricellis: Again comment says 2 interceptors, but the assert checks for 1.

ricellis:
+1

CDTDatastore/HTTP/CDTHTTPInterceptorContext.h
```diff
@@ -39,7 +48,8 @@ NS_ASSUME_NONNULL_BEGIN
  *
  *  @param request the request this context should represent
  **/
-- (instancetype)initWithRequest:(NSMutableURLRequest *)request NS_DESIGNATED_INITIALIZER;
+- (instancetype)initWithRequest:(NSMutableURLRequest *)request
+                          state:(NSMutableDictionary *)state NS_DESIGNATED_INITIALIZER;
```
rhyshort: I don't think we can just add the `NSMutableDictionary` to the init method, since the original API was envisioned as a user being able to return their own context with any customisation they want, so this would be a breaking API change. Basically this comes down to whether this is deemed API or not, my stance on that from thinking back is that it is.

CDTDatastore/HTTP/CDTRequestLimitInterceptor.m
```diff
@@ -0,0 +1,86 @@
...
+    return [[CDTRequestLimitInterceptor alloc] init];
+}
+
+- (instancetype)init
+{
+    if (self = [super init]) {
+        _initialSleep = 0.25; // 250ms
+        _maxRetries = 3;
+    }
+    return self;
```
rhyshort: This should call down into the `initWithSleep:maxRequests:` method.

CDTDatastore/HTTP/CDTHTTPInterceptorContext.h
tomblench: It's an interesting point. If they can return a new context, then they can destroy the state for other interceptors. Of course they can modify the state as much as they want inside the interceptor anyway.

CDTDatastore/HTTP/CDTRequestLimitInterceptor.h
```diff
@@ -0,0 +1,22 @@
...
+#import "CDTLogging.h"
+
+
+
+@interface CDTRequestLimitInterceptor : NSObject <CDTHTTPInterceptor>
+
++ (nonnull instancetype)interceptor;
+- (nonnull instancetype)init;
+- (nonnull instancetype)initWithSleep:(NSTimeInterval)sleep
+                          maxRequests:(int)maxRetries;
```
rhyshort: Also add NS_DESIGNATED_INITIALIZER to the end of this declaration

rhyshort:
+1

PR #294, opened 2016-07-01
Delete temporary files suffixed with -wal and -shm

### LINKED ISSUE #272: SQLite temporary files are not deleted when the database is deleted
Journal mode for main SQLite database is WAL. In this case, at least one temporary file is created in the same folder adding '-wal' to the name of the database (and also another one adding '-shm'). However the method in TD_Database to delete a database: 'TD_Database:deleteDatabase:' only removes the file for the database.

rhyshort:
## What
When deleting a database from disk, delete all temp files too

## How
Delete files suffixed with `-wal` and `-shm` which are sqlite temp files for the journalling mode we use.

## Testing
No unit tests due to the difficulty in reliably making sure that the files exist before deleting.

## Issues

Fixes #272

### DIFF
--- a/CDTDatastore/touchdb/TD_Database.m
+++ b/CDTDatastore/touchdb/TD_Database.m
@@ -581,8 +581,11 @@ static BOOL removeItemIfExists(NSString* path, NSError** outError)
     if ([TD_Database existsDatabaseAtPath:path]) {
         NSString *attachmentsPath = [TD_Database attachmentStorePathWithDatabasePath:path];
 
+        // Delete the database file, and the associated temp files if they are present.
         success =
-            (removeItemIfExists(path, outError) && removeItemIfExists(attachmentsPath, outError));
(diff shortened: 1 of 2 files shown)


CDTDatastore/touchdb/TD_Database.m
```diff
@@ -582,7 +582,9 @@ + (BOOL)deleteClosedDatabaseAtPath:(NSString *)path error:(NSError **)outError
         NSString *attachmentsPath = [TD_Database attachmentStorePathWithDatabasePath:path];
 
         success =
```
emlaver: It might be worth adding a comment about what files are being removed from where.

rhyshort: done in a1fe4d7, along with a useless cocoapods xcproject diff.

emlaver:
+1

PR #298, opened 2016-07-08
Don't expose NSMutableDictionary directly

tomblench:
- _What_: Instead of exposing the `state` dictionary directly, add methods for setting/retrieving values.
- _Why_: Simplify use for API users and enable future refactoring
- _Testing_: All existing tests pass

### DIFF
--- a/CDTDatastore/HTTP/CDTHTTPInterceptorContext.m
+++ b/CDTDatastore/HTTP/CDTHTTPInterceptorContext.m
@@ -19,16 +19,18 @@
 
 @implementation CDTHTTPInterceptorContext
 
+NSMutableDictionary<NSString*, NSObject*> *_state;
+
 -(instancetype)init {
     NSAssert(NO, @"Call the designated initaliser");
     return nil;
 }
 
-- (instancetype)initWithRequest:(NSMutableURLRequest *)request {
+- (instancetype)initWithRequest:(NSMutableURLRequest*)request {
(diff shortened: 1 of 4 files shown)


CDTDatastore/HTTP/CDTHTTPInterceptorContext.m
```diff
@@ -19,16 +19,18 @@
 
 @implementation CDTHTTPInterceptorContext
 
+NSMutableDictionary<NSString*, NSObject*> *_state;
```
rhyshort: A nice new language feature we could use for this is class properties.

tomblench: Not sure what you mean @rhyshort - obviously "normal" properties can't be used for things which are meant to be private, is this something else? An example would be helpful.

rhyshort:
```objc
@property (class) NSString *someString
```

See [wwdc video](https://developer.apple.com/videos/play/wwdc2016/405/) about five minutes in.You'll need a newer LLVM compiler to be able to use it though, so it might be best to stick it as a ticket until Xcode 8 is final around september/october time.

rhyshort: Ah just watched the video a bit further on the job my memory, nothing os synthesised so we'd still need to define storage

rhyshort:
+1

PR #305, opened 2016-07-25
Fix failing replication acceptance test cases

emlaver:
## What
The test cases below are currently failing with error message `("1001") is not equal to ("4") - Wrong error code: 1001`:
`testPullErrorsWhenLocalDatabaseIsDeleted`
`testPushErrorsWhenLocalDatabaseIsDeleted`

The error code 4 is a CDTReplicator error which is supposed to reflect the 1001 TDReplicator error code.  We need to check that the CDTReplicator `error` property is null before setting the underlying error pointer.
Original issue: #251 

## How
- Fixed failing replication acceptance test cases by checking if CDTReplicator's `strongSelf.error` is null before setting underlying error pointer

## Testing
No new unit tests. 
Existing replication tests `testPullErrorsWhenLocalDatabaseIsDeleted` and 
`testPushErrorsWhenLocalDatabaseIsDeleted`are now passing.


tomblench:
+1

PR #307, opened 2016-07-26
269 datastore documentation

### LINKED ISSUE #269: Documentation issue regarding deleted datastores
The documentation for CDTDatastoreManager -deleteDatastoreNamed:error: says 

"it is the responsibility of the caller to ensure that extensions should be shutdown (and their underlying databases closed) before calling this method."

However, a CDTDatastore -close does not exist nor is there any documentation on what an extension is and how to shut it down.

rhyshort:
# What
1) Make sure databases are closed in the index manager.
2) Correct documentation of CDTDatastoreManager

## How
1) call `[database close]` in dealloc and before returning `nil` when opening the database failed because migrations couldn't be completed. 
2) Correct documentation of CDTDatastoreManager by removing incorrect doc comment.

## Issues

Fixes #269

### DIFF
--- a/CDTDatastore/CDTDatastoreManager.h
+++ b/CDTDatastore/CDTDatastoreManager.h
@@ -55,9 +55,6 @@ extern NSString *__nonnull const CDTDatastoreErrorDomain;
 
  All datastore files, including attachments and extensions, are deleted.
 
- Currently it is the responsibility of the caller to ensure that extensions should be shutdown (and
- their underlying databases closed) before calling this method.
-
  @param name datastore name
  @param error will point to an NSError object in case of error.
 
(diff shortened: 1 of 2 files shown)


CDTDatastore/CDTDatastoreManager.h
```diff
@@ -55,9 +55,6 @@ extern NSString *__nonnull const CDTDatastoreErrorDomain;
 
  All datastore files, including attachments and extensions, are deleted.
```
emlaver: Should we mention that `[database close]` is now being called? Or, would it be enough in the CHANGELOG.md?

rhyshort: I think it's something for the change log. Since there is no way to explicitly  close the index manager, you just let it deallocate and clean it's self up.

emlaver:
LGTM +1

PR #309, opened 2016-08-02
Simplify replicator creation

### LINKED ISSUE #291: Simplify replicator creation
Currently creation wends its way like this:
CDTReplicator -> TD_ReplicatorManager -> TDDatabaseManager -> TDReplicatorManager -> CDTReplicator

We should simplify with something like:
https://github.com/cloudant/CDTDatastore/compare/51918-simple-replicator-creation
i.e. removing the factory and manager use.

emlaver:
## What
Simplify the replicator creation by removing the manager classes from the current creation:
CDTReplicator -> TD_ReplicatorManager -> TDDatabaseManager -> TDReplicatorManager -> CDTReplicator

Any existing code needed in ``TDDatabaseManager`` and ``TDReplicatorManager`` was moved into CDTReplicator. 

## How
Removed TDReplicatorManager from replicator creation
- Replaced TDReplicatorManager imports and usages with TD_DatabaseManager
- Renamed initWithTDReplicatorManager to initWithTDDatabaseManager
- Handled replicatorWithConfiguration and createReplicatorWithProperties methods in CDTReplicator

Removed replication code from TD_DatabaseManager and TDReplicatorManager as it's no longer required after moving all TDReplicator create code to the CDTReplicator
- Removed TDReplicatorManager classes since they are not maintained or used
- Added remoteCheckpointDocID method from TDReplicator to CDTReplicator for testing
- Replaced TDReplicator instances with CDTReplicator in ReplicationAcceptance test class

Constructed TDReplicator entirely within CDTReplicator

Remove TDReplicatorManager from replicator creation
- Replaced TDReplicatorManager imports and usages with TD_DatabaseManager
- Renamed initWithTDReplicatorManager to initWithTDDatabaseManager
- Handled replicatorWithConfiguration and createReplicatorWithProperties methods in CDTReplicator

Removed options dictionary and replaced with heartbeat and reset variables
- Made all replication config into TDReplicator properties

Removed ``dictionaryForReplicatorDocument`` method and any instances of it as it's not used and can be replaced by ``validateOptionalHeaders``

## Testing
No new unit tests.
Removed unit tests ``testDictionaryForPullReplicationDocument`` and ``testDictionaryForPushReplicationDocument``.
Replaced instances of ``dictionaryForReplicatorDocument`` in unit tests with ``validateOptionalHeaders``.

## Issue
fixes #291


CDTDatastore/CDTReplicator/CDTAbstractReplication.m
```diff
@@ -78,25 +78,22 @@ - (void)clearInterceptors { self.httpInterceptors = @[]; }
...
         // check for strings
-        for (id key in self.optionalHeaders) {
+        for (id key in candidateHeaders) {
             if (![key isKindOfClass:[NSString class]]) {
                 CDTLogWarn(CDTREPLICATION_LOG_CONTEXT,
-                        @"CDTAbstractReplication " @"-dictionaryForReplicatorDocument Error: "
-                        @"Replication HTTP header key is invalid (%@).\n It must be NSString. "
-                        @"Found type %@",
-                        key, [key class]);
+                           @"CDTAbstractReplication " @"-dictionaryForReplicatorDocument Error: "
```
tomblench: s/dictionaryForReplicatorDocument/validateOptionalHeaders/

CDTDatastore/CDTReplicator/CDTAbstractReplication.m
```diff
@@ -106,15 +103,15 @@ - (NSDictionary *)dictionaryForReplicatorDocument:(NSError *__autoreleasing *)er
...
             }
 
-            if (![self.optionalHeaders[key] isKindOfClass:[NSString class]]) {
+            if (![candidateHeaders[key] isKindOfClass:[NSString class]]) {
                 CDTLogWarn(CDTREPLICATION_LOG_CONTEXT,
-                        @"CDTAbstractReplication " @"-dictionaryForReplicatorDocument Error: "
-                        @"Value for replication HTTP header %@ is invalid (%@).\n"
-                        @"It must be NSString. Found type %@.",
-                        key, self.optionalHeaders[key], [self.optionalHeaders[key] class]);
+                           @"CDTAbstractReplication " @"-dictionaryForReplicatorDocument Error: "
```
tomblench: s/dictionaryForReplicatorDocument/validateOptionalHeaders/

CDTDatastore/CDTReplicator/CDTPullReplication.m
```diff
@@ -65,57 +65,17 @@ - (instancetype)copyWithZone:(NSZone *)zone
...
-        return nil;
-    }
-
-    NSMutableDictionary *doc = [NSMutableDictionary dictionaryWithDictionary:superdoc];
-
-    [doc setObject:self.source.absoluteString forKey:@"source"];
-
-    if (self.target) {
-        [doc setObject:self.target.name forKey:@"target"];
+    NSMutableDictionary *dictionary = [[NSMutableDictionary alloc] init];
```
tomblench: I think you need to revisit `description` for `CDTPullReplication` and `CDTPushReplication`.

These methods are only for debugging/logging etc, but I think they're no longer correct. You just need to build up a string which describes all of the options - source, target, headers, interceptors, filters, query_params etc.

Note that these methods previously relied in large part on `dictionaryForReplicatorDocument` to do this work and then printed a string representation of the dictionary.

My advice would be to dispense with the dictionary all together and just build up a string formatted something like this:
```
@"source: %@, target: %@, headers: %@, interceptors: %@, filter: %@, query_params: %@"
```
You don't have to worry about `nil` checks as these will just come out as `nil` in the formatted output.

CDTDatastore/CDTReplicator/CDTReplicator.m
```diff
@@ -288,6 +314,74 @@ - (BOOL)startWithTaskGroup:(dispatch_group_t)taskGroup error:(NSError *__autorel
...
+    repl.requestHeaders = properties[@"headers"];
+    
+    // Pull replications can have filters assigned.
+    if ([configuration isKindOfClass:[CDTPullReplication class]]) {
+        CDTPullReplication *shadowConfig = (CDTPullReplication *)configuration;
+        repl.filterName = shadowConfig.filter;
+        repl.filterParameters = shadowConfig.filterParams;
+    }
+    
+    //    repl.authorizer = authorizer;  // we don't support authorisers
```
tomblench: remove commented code

CDTDatastore/CDTReplicator/CDTReplicator.m
```diff
@@ -288,6 +314,74 @@ - (BOOL)startWithTaskGroup:(dispatch_group_t)taskGroup error:(NSError *__autorel
...
+    }
+    
+    repl.reset = NO;
+    repl.heartbeat = nil;
+    
+    // Headers are validated before being put in properties
+    repl.requestHeaders = properties[@"headers"];
+    
+    // Pull replications can have filters assigned.
+    if ([configuration isKindOfClass:[CDTPullReplication class]]) {
```
tomblench: could replace this with `if (!push)`

CDTDatastoreReplicationAcceptanceTests/ReplicationAcceptance.m
```diff
@@ -1492,7 +1491,13 @@ -(void)testCreateReplicationWithExtraHeaders
     NSDictionary *extraHeaders = @{@"SpecialHeader": @"foo", @"user-agent":userAgent};
     pull.optionalHeaders = extraHeaders;
     
-    NSDictionary *pullDoc = [pull dictionaryForReplicatorDocument:nil];
+    NSMutableDictionary *pullDoc = [[NSMutableDictionary alloc] init];
```
tomblench: You can rewrite this test without reference to `pullDoc`, as it's no longer needed. Likewise you can remove the asserts below which reference `pullDoc`.

CDTDatastoreTests/CDTReplicationTests.m
```diff
@@ -401,20 +332,17 @@ -(void)testCreatePushReplicationWithFilter
     XCTAssertNotNil(replicator, @"%@", push);
     XCTAssertNil(error, @"%@", error);
 
-    NSDictionary *pushDoc = [push dictionaryForReplicatorDocument:nil];
+    NSMutableDictionary *pushDoc = [[NSMutableDictionary alloc] init];
```
tomblench: Rewrite this test without reference to `pushDoc`. There's actually not much to do in this test any more, but you could check that the underlying `TDReplicator` has had the correct filter and params set from the `CDTReplicator` object.

CDTDatastoreTests/CDTReplicationTests.m
```diff
@@ -569,13 +497,17 @@ -(void)testForProhibitedOptionalReplicationHeaders
 {
     CDTPullReplication *pull;
     NSError *error;
-    NSDictionary *pullDoc;
+    NSMutableDictionary *pullDoc;
```
tomblench: As above, remove `pullDoc` and see if there is anything to be gained from asserts which compare the various settings (headers, interceptors) between TD and CDT replicators.

CDTDatastore/CDTReplicator/CDTPullReplication.m
emlaver: Improved description in 9bbbc4e.

CDTDatastore/CDTReplicator/CDTReplicator.m
emlaver: Removed in 9bbbc4e.

CDTDatastore/CDTReplicator/CDTReplicator.m
emlaver: Fixed in 9bbbc4e.

CDTDatastoreReplicationAcceptanceTests/ReplicationAcceptance.m
emlaver: Fixed in 9bbbc4e.

CDTDatastoreTests/CDTReplicationTests.m
emlaver: Removed test in 9bbbc4e.

CDTDatastoreTests/CDTReplicationTests.m
emlaver: Fixed in 9bbbc4e.

CDTDatastoreTests/CDTReplicationTests.m
```diff
@@ -528,16 +480,11 @@ -(void)testForProhibitedOptionalReplicationHeaders
...
-        } else {
-            pullDoc = nil;
-        }
-        XCTAssertNil(pullDoc, @"CDTPullReplication -dictionaryForReplicatorDocument passed with "
-                       @"header: %@, pullDoc: %@", optionalHeaders, pullDoc);
+        CDTReplicatorFactory *replicatorFactory =
+        [[CDTReplicatorFactory alloc] initWithDatastoreManager:self.factory];
+        CDTReplicator *rep = [replicatorFactory oneWay:pull error:&error];
+        
+        XCTAssertNil(rep, @"Error was not set");
```
tomblench: Also check that the value of `error` is set correctly

CDTDatastoreTests/CDTReplicationTests.m
```diff
@@ -546,16 +493,11 @@ -(void)testForProhibitedOptionalReplicationHeaders
...
-        } else {
-            pullDoc = nil;
-        }
-        XCTAssertNil(pullDoc, @"CDTPullReplication -dictionaryForReplicatorDocument passed with "
-                    @"header: %@, pullDoc: %@", optionalHeaders, pullDoc);
+        CDTReplicatorFactory *replicatorFactory =
+        [[CDTReplicatorFactory alloc] initWithDatastoreManager:self.factory];
+        CDTReplicator *rep = [replicatorFactory oneWay:pull error:&error];
+        
+        XCTAssertNil(rep, @"Error was not set");
```
tomblench: Also check that the value of `error` is set correctly

CDTDatastore/CDTReplicator/CDTPullReplication.m
```diff
@@ -65,17 +65,8 @@ - (instancetype)copyWithZone:(NSZone *)zone
...
-    if ([CDTAbstractReplication validateOptionalHeaders:self.optionalHeaders error:nil]) {
-        dictionary[@"headers"] = self.optionalHeaders;
-        dictionary[@"interceptors"] = self.httpInterceptors;
-    } else {
-        return nil;
-    }
-    
-    dictionary[@"source"] = TDCleanURLtoString(self.source);
-    return [NSString stringWithFormat:@"%@: %@", [self class], dictionary];
+    return [NSString stringWithFormat:@"source: %@, target: %@, headers: %@, interceptors: %@, filter: %@, query_params: %@",
```
tomblench: You should also add `[self class]` to the output so a user can tell whether it's pull or push

CDTDatastore/CDTReplicator/CDTPushReplication.m
```diff
@@ -66,16 +66,9 @@ - (instancetype)copyWithZone:(NSZone *)zone
...
-    
-    if ([CDTAbstractReplication validateOptionalHeaders:self.optionalHeaders error:nil]) {
-        dictionary[@"headers"] = self.optionalHeaders;
-        dictionary[@"interceptors"] = self.httpInterceptors;
-    }
-    
-    dictionary[@"target"] = TDCleanURLtoString(self.target);
-    return [NSString stringWithFormat:@"%@: %@", [self class], dictionary];
+{    
+    return [NSString stringWithFormat:@"source: %@, target: %@, headers: %@, interceptors: %@, filter: %@, query_params: %@",
```
tomblench: You should also add `[self class]` to the output so a user can tell whether it's pull or push

CDTDatastore/CDTReplicator/CDTReplicator.m
```diff
@@ -288,6 +314,75 @@ - (BOOL)startWithTaskGroup:(dispatch_group_t)taskGroup error:(NSError *__autorel
...
+/**
+ Builds a TDReplicator object from a CDTPush/PullReplication configuration.
+ */
+- (TDReplicator *)buildTDReplicatorFromConfiguration:(CDTAbstractReplication *)configuration
+                                               error:(NSError *__autoreleasing *)error
+{
+    // This validates the optional headers
+    NSMutableDictionary *properties = [[NSMutableDictionary alloc] init];
+    
+    if ([CDTAbstractReplication validateOptionalHeaders:configuration.optionalHeaders error:error]) {
```
tomblench: `validateOptionalHeaders` is not needed here since it was already called by `initWithTDDatabaseManager`

CDTDatastore/CDTReplicator/CDTReplicator.m
```diff
@@ -288,6 +314,75 @@ - (BOOL)startWithTaskGroup:(dispatch_group_t)taskGroup error:(NSError *__autorel
     return YES;
 }
 
+/**
+ Builds a TDReplicator object from a CDTPush/PullReplication configuration.
+ */
+- (TDReplicator *)buildTDReplicatorFromConfiguration:(CDTAbstractReplication *)configuration
```
tomblench: It's confusing to have `configuration` as an argument to this method because it's already a member of this class, so there is no need to pass the argument to this method.

You have two choices - either make it a static method which takes the argument, or keep it as an instance method and remove the argument.

CDTDatastore/CDTReplicator/CDTAbstractReplication.h
```diff
@@ -169,15 +169,6 @@ typedef NS_ENUM(NSInteger, CDTReplicationErrors) {
  ---------------------------------------------------------------------------------------
  */
 
-/** The NSDictionary is used by CDTReplicatorFactory to properly configure CDTReplicator objects.
```
rhyshort: The comment on 161 - 165, should be removed.

CDTDatastore/CDTReplicator/CDTReplicator.m
```diff
@@ -109,26 +110,51 @@ + (NSString *)stringForReplicatorState:(CDTReplicatorState)state
     }
 }
 
++ (CDTReplicator *)replicatorWithConfiguration:(CDTAbstractReplication *)configuration
```
rhyshort: Shouldn't this be in the header so it can be used?

tomblench: @rhyshort it's internal to this class only

CDTDatastore/CDTReplicator/CDTReplicator.h
```diff
@@ -258,4 +258,7 @@ initWithTDReplicatorManager:(nonnull TDReplicatorManager *)replicatorManager
  */
 @property (nullable, nonatomic, readonly) NSError *error;
 
+/** Exposed for testing. Returns the doc ID for the checkpoint document. */
+- (nonnull NSString *)remoteCheckpointDocID;
```
rhyshort: We shouldn't have this exposed here even for testing, you can use the runtime to get the selector and then use `performSelector` to get the result.

tomblench: Even better just declare it at the top of the test class which calls it.

CDTDatastore/touchdb/TDPuller.m
```diff
@@ -155,7 +155,7 @@ - (void)startChangeTracker
     _changeTracker.filterParameters = _filterParameters;
     _changeTracker.docIDs = _docIDs;
     _changeTracker.authorizer = _authorizer;
-    unsigned heartbeat = $castIf(NSNumber, _options[@"heartbeat"]).unsignedIntValue;
+    unsigned heartbeat = $castIf(NSNumber, self.heartbeat).unsignedIntValue;
```
rhyshort: no need for the cast here, `heartbeat` is already an `NSNumber`

CDTDatastore/touchdb/TDReplicator.m
```diff
@@ -147,8 +148,8 @@ - (bool)hasSameSettingsAs:(TDReplicator*)other
 {
     return _db == other->_db && $equal(_remote, other->_remote) && self.isPush == other.isPush &&
            _continuous == other->_continuous && $equal(_filterName, other->_filterName) &&
-           $equal(_filterParameters, other->_filterParameters) &&
-           $equal(_options, other->_options) && $equal(_docIDs, other->_docIDs) &&
+           $equal(_filterParameters, other->_filterParameters) && _reset == other->_reset &&
+           $equal(_heartbeat, other->_heartbeat) && $equal(_docIDs, other->_docIDs) &&
```
rhyshort: We don't need to use `$equal` for `heartbeat` we can use `isEqualToNumber:` instead.

CDTDatastore/CDTReplicator/CDTReplicator.h
rhyshort: Won't that give a warning if the `implmentation` is missing for it in scope?

CDTDatastore/CDTReplicator/CDTPullReplication.m
emlaver: Fixed in 06cf8b0.

CDTDatastore/CDTReplicator/CDTPushReplication.m
emlaver: Fixed in 06cf8b0.

CDTDatastore/CDTReplicator/CDTReplicator.m
emlaver: Fixed in 06cf8b0.

CDTDatastore/CDTReplicator/CDTReplicator.m
emlaver: Fixed in 06cf8b0.

CDTDatastore/CDTReplicator/CDTAbstractReplication.h
emlaver: Removed in 06cf8b0.

CDTDatastore/touchdb/TDPuller.m
emlaver: Fixed in 06cf8b0.

CDTDatastore/touchdb/TDReplicator.m
emlaver: Fixed in 06cf8b0.

CDTDatastore/CDTReplicator/CDTReplicator.m
rhyshort: which is the problem. The class is the replicator and without this exposed we call `init` instead. See https://github.com/cloudant/CDTDatastore/pull/309/files#diff-4060e2491f9dfa52069155a3851436f3R84 so we should either be using this method or it should be removed.

CDTDatastore/CDTReplicator/CDTReplicator.h
emlaver: Fixed in 51b56af.

CDTDatastore/CDTReplicator/CDTReplicator.m
tomblench: OK I see what you mean, this is a convenience for calling `alloc`, `initWithTDDatabaseManager`. But I wonder if we need it because it might distract/confuse users from doing it the "proper" way which is to use the `CDTReplicatorFactory`.

rhyshort: It probably will, since it will be the first place they look, so it should be removed.

tomblench:
@emlaver once you've addressed https://github.com/cloudant/CDTDatastore/pull/309#discussion_r74215998 I think we are ready to squash, rebase and merge

CDTDatastore/CDTReplicator/CDTReplicator.m
emlaver: Removed replicatorWithConfiguration in b544832.

tomblench:
+1 squash and rebase. Once the builds look fine we can merge.

Issue #308, opened 2016-08-02
Tests test429Retry and testFiltersWithChangesFeed are failing against iOS 


emlaver:
After speaking with @tomblench, it's possible that these tests are failing because the interceptor is not being called.  It's probably best to exclude these tests from running against iOS until the issue can be fixed.

tomblench:
We can't run these tests on iOS because we can't open arbitrary ports there. These tests rely on starting a mini HTTP server so we can't run them on the iOS platform. For now, disabling them on the iOS platform is acceptable

PR #310, opened 2016-08-10
Exclude tests from iOS

tomblench:
- _What_: Exclude failing tests from building and running on iOS

- _Why_: The iOS platform doesn't allow us to listen on arbitrary ports, which we need to do in these tests

Testing: It should be sufficient for these test to only run on the macOS platform.

Issue: #308

### DIFF
--- a/CDTDatastoreTests/CDTReplicationTests.m
+++ b/CDTDatastoreTests/CDTReplicationTests.m
@@ -206,6 +206,8 @@
     XCTAssertEqualObjects([push.httpInterceptors[0] class], [CDTSessionCookieInterceptor class]);
 }
 
+// this test can only run on macOS and not iOS because it needs to start a server
+#if TARGET_OS_MAC && !TARGET_OS_IPHONE
 - (void)test429Retry
 {
     NSError *error = nil;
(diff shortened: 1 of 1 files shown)


CDTDatastoreTests/CDTReplicationTests.m
```diff
@@ -206,6 +206,8 @@ - (void)testURLCredsReplacedWithCookieInterceptorPush
     XCTAssertEqualObjects([push.httpInterceptors[0] class], [CDTSessionCookieInterceptor class]);
 }
 
+// this test can only run on macOS and not iOS because it needs to start a server
+#if TARGET_OS_MAC
```
brynh: This won't have the desired effect as `TARGET_OS_MAC` is considered a superclass of `TARGET_OS_IPHONE` and `TARGET_OS_IOS`, so will be true on iOS as iOS is considered a variant of OSX (see the comments in `TargetConditionals.h`).  Therefore, I think this should be `#if TARGET_OS_MAC && !TARGET_OS_IPHONE` to only work on Mac and not on iOS devices.

CDTDatastoreTests/CDTReplicationTests.m
```diff
@@ -254,7 +256,10 @@ - (void)test429Retry
     
     [server stop];
 }
+#endif
 
+// this test can only run on macOS and not iOS because it needs to start a server
+#if TARGET_OS_MAC
```
brynh: `#if TARGET_OS_MAC && !TARGET_OS_IPHONE`

brynh:
+1

PR #312, opened 2016-08-23
Cordova fixes

brynh:
## What
This PR:

- makes `CDTNSURLSessionConfigurationDelegate.h` a public header file
- ensures the updated request is copied to the `CDTHTTPInterceptorContext` used by the response interceptors so the updated version of the request is visible to the response interceptors.

**Note:** These changes will need to be included in a release of CDTDatastore so that we can build sync-cordova-plugin against a tagged version of the CDTDatastore framework.

## Why
These changes are required so that the sync-cordova-plugin can successfully build against the CDTDatastore framework and the automated tests can pass.

## Testing
No additional tests required.  Existing tests still pass.  The iOS tests on sync-cordova-plugin can build and pass using a framework built from this branch.


rhyshort:
+1

PR #314, opened 2016-08-25
Add concurrency controls for open, close and delete

### LINKED ISSUE #273: No concurrency control for operations in database files: 'open', 'close' & 'delete'
Once we create (and open) a FMDatabase using the methods in TD_Database, all operations in the database are executed in a FMDatabaseQueue, i.e. concurrency is managed through this serial queue. However, when the methods in TD_Database to open, close and delete one of these databases have no concurrency control at all, so it is possible to delete the database from different threads with undefined results.

This bug was first detected and commented https://github.com/cloudant/CDTDatastore/pull/112#discussion_r27128493 and https://github.com/cloudant/CDTDatastore/pull/112#discussion_r27128921.

rhyshort:
## What
Add concurrency controls for open, close and delete TD_Database operations.

##How
Use a serial dispatch queue when closing and providing isOpen via an atomic
property rather than the current direct interaction with the ivar.

## Tests

No additional tests, CI indicates all tests pass.

## Issues
Resolves #273

### DIFF
--- a/CDTDatastore/touchdb/TD_Database.m
+++ b/CDTDatastore/touchdb/TD_Database.m
@@ -69,9 +69,18 @@ NSString* const TD_DatabaseWillBeDeletedNotification = @"TD_DatabaseWillBeDelete
 //}
 //@end
 
+@interface TD_Database()
+
+@property (nonatomic) dispatch_queue_t queue;
+@property (atomic, getter=isOpen) BOOL open;
+
+
+@end
+
 @implementation TD_Database
 
 @synthesize fmdbQueue = _fmdbQueue;
+
 
 static BOOL removeItemIfExists(NSString* path, NSError** outError)
 {
@@ -99,7 +108,7 @@ static BOOL removeItemIfExists(NSString* path, NSError** outError)
     return db;
 }
 
-- (id)initWithPath:(NSString*)path
+- (instancetype)initWithPath:(NSString*)path
 {
     if (self = [super init]) {
         Assert([path hasPrefix:@"/"], @"Path must be absolute");
@@ -111,6 +120,7 @@ static BOOL removeItemIfExists(NSString* path, NSError** outError)
             _validations = nil;
             _pendingAttachmentsByDigest = nil;
         }
+        _queue = dispatch_queue_create("com.cloudant.sync.db", NULL); //Serial dispatch queue.
     }
     return self;
 }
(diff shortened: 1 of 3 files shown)


CDTDatastore/touchdb/TD_Database.m
```diff
@@ -522,48 +531,57 @@ INSERT INTO INFO (key, value) VALUES ('publicUUID',  '%@')",
...
+        self.open = YES;
         return YES;
     } else {
         return NO;
     }
 }
 
 #pragma mark Closing and deleting database
-
+//TODO: SOLOUTION- to fix the concurrency issues, dispatching the open, closed and delete of database files
```
emlaver: Remove TODO and SOLUTION from comment.

rhyshort: removed comment be52627

emlaver:
LGTM +1

brynh:
I think we need additional protection against concurrent access for `deleteDatabase:` and `deleteClosedDatabaseAtPath:error:`. Currently, only the `close` part of `deleteDatabase:` is protected, but that leaves a window of opportunity for the database to be reopened before (or during)`deleteClosedDatabaseAtPath:error:`. In `deleteDatabase:` we should really be doing the close (if necessary) and the delete as an atomic operation.

I'd suggest having internal versions of the existing `close` and `deleteClosedDatabaseAtPath:error:` that don't use the dispatch queue and then making the exported operations wrap the calls to the internal operations in a single block on the dispatch queue so we can combine them without deadlock for the `deleteDatabase:` case and call them individually from the exported `close` and `deleteClosedDatabaseAtPath:error:`.

rhyshort:
@brynh, moving to using internal methods and making sure `deleteClosedDatabaseAtPath` performs its actions on the queue is difficult. Since we can only create a queue not create a shared queue, we can't execute the delete on the same queue as the close will take place. 

Due the API in CDTDatastore, it is possible to delete a datastore that hasn't been opened. So I can't make the `deleteClosedDatabaseAtPath` a method, without making a breaking API change 😞  We might have to accept that calling the `deleteDatastoreNamed` method on the `CDTDatastoreManager` has the potential to cause problems across multiple threads.

The only thing I think we can do is have a single queue across the entire application for open, close and delete operations, but I'm not keen on the idea of having one database close waiting on another when they could be done in parallel. Any thoughts?

brynh:
Oops - I somehow missed the fact `deleteClosedDatabaseAtPath` is a class method. I agree that makes things awkward for `deleteClosedDatabaseAtPath` and that it's not nice to have a single queue across the whole application. However, we can still improve things for `deleteDatabase:` along the lines I suggested, using an internal version of `close` and then doing both the call to the internal close and the call to `deleteClosedDatabaseAtPath` in a single block on the dispatch queue.

rhyshort:
@brynh I've improved `deleteDatabase` as suggested in 50c25af

CDTDatastore/touchdb/TD_Database.m
```diff
@@ -522,18 +531,30 @@ INSERT INTO INFO (key, value) VALUES ('publicUUID',  '%@')",
...
+
+    __block BOOL result;
+    dispatch_sync(self.queue, ^{
+        result = [self closeOnQueue];
+    });
+
+    return result;
+}
+
+- (BOOL) closeOnQueue
```
brynh: I don't like this method name as the `OnQueue` suffix implies it queues it whereas this method itself doesn't use a queue but you intend that it should be called from something that does queue it.  I think I'd call it `closeInternal` or `closePrivate` (unless you can think of a better name) and maybe add a comment suggesting it should only be called when wrapped in a block on the dispatch queue.

CDTDatastore/touchdb/TD_Database.m
```diff
@@ -547,29 +568,38 @@ - (BOOL)close
...
 - (BOOL)deleteDatabase:(NSError **)outError
 {
+    __block BOOL result =  NO;
+    dispatch_sync(self.queue, ^{
+        result = [self deleteDatabaseOnQueue:outError];
+    });
+    return result;
+}
+
+- (BOOL)deleteDatabaseOnQueue:(NSError **)outError
```
brynh: As with `closeOnQueue`, I don't like the `OnQueue` suffix.

CDTDatastore/touchdb/TD_Database.m
rhyshort: done in 36b2d3c

CDTDatastore/touchdb/TD_Database.m
rhyshort: done in 36b2d3c

brynh:
I like the updated `deleteDatabase:`, but it looks like there are some tests that need fixing as a result. I think the code's right but the tests are failing because they override `close` and we're now calling `closeInternal` instead.

rhyshort:
@brynh The tests should be fixed in a1765d0

brynh:
+1

rhyshort:
@emlaver Are you still +1?

emlaver:
Yes, +1 for the new changes.

PR #315, opened 2016-08-26
Make headers for CDTDefines.h and CDTRequestLimitInterceptor.h public

brynh:
## What
Add `CDTDefines.h` and `CDTRequestLimitInterceptor.h` to public headers.

## Why
These are required so that CDTDatastore can be successfully used as a framework.

## Testing
sync-cordova-plugin can successfully build for iOS using the `CDTDatastore.framework` generated from this. Previously it had only worked because I had not deleted the contents of the `Headers` folder inside `CDTDatastore.framework`, so was picking up old versions of the header files. I've now updated the script I use to build a new version of the frameworks and copy them to sync-cordova so that it deletes the old frameworks used by cordova before replacing them with the new versions.


ricellis:
+1
although this has reminded me that we should probably look at whether CDTRequestLimitInterceptor should be renamed before release, but that is unrelated to this change.

PR #317, opened 2016-09-02
Rename CDTRequestLimitInterceptor to CDTReplay429Interceptor

### LINKED ISSUE #316: Rename CDTRequestLimitInterceptor
I think we should rename `CDTRequestLimitInterceptor` to `Replay429Interceptor` or `CDTReplay429Interceptor` to line up with the name used in Android which makes it a bit more clear what it is going to do.

rhyshort:
Rename CDTRequestLimitInterceptor to CDTReplay429Interceptor which indicates more about what it does.

Resolves #316


ricellis:
+1

PR #320, opened 2016-09-13
Recover from errors during migration.

### LINKED ISSUE #278: 'TD_Database:openWithEncryptionKeyProvider:': Rework this method so we can rescue failures during migration
Although we exit the method when there is an error while updating the version, on other cases we update the database but we do not check the result, for example:

```objc
            while ([attachmentKeys next]) {
                NSData *keyData = [attachmentKeys dataNoCopyForColumn:@"key"];
                
                [TD_Database generateAndInsertFilenameBasedOnKey:*(TDBlobKey *)keyData.bytes
                                intoBlobFilenamesTableInDatabase:db];
            }
```

rhyshort:
## What
Recover from errors during database migration by make schema updates in
a transaction and rolling back whenever an error occurs.

## How
- Move migrations to being run inside a transaction, so they can be rolled back in the event of an error.
- Check the return value from the `generateAndInsertFilenameBasedOnKey:intoBlobFilenamesTableInDatabase` method, so we can rollback if a attachment failed to be moved.

## Tests

All current tests pass, I'm not entirely sure how to force FMDB to error so we can test rollbacks, any advice on this would be appreciated.

## Issues

Fixes #278

### DIFF
--- a/CDTDatastore/touchdb/TD_Database.m
+++ b/CDTDatastore/touchdb/TD_Database.m
@@ -331,8 +331,11 @@ static BOOL removeItemIfExists(NSString* path, NSError** outError)
 
     __block BOOL result = YES;
     __weak TD_Database* weakSelf = self;
-    [_fmdbQueue inDatabase:^(FMDatabase* db) {
+    [self.fmdbQueue inTransaction:^(FMDatabase* db, BOOL* rollback) {
         TD_Database* strongSelf = weakSelf;
+
+        // Transaction will be rolled back unless, rollback is changed to NO.
+        *rollback = YES;
 
         // Check the user_version number we last stored in the database:
         int dbVersion = [db intForQuery:@"PRAGMA user_version"];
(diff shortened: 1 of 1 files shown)


CDTDatastore/touchdb/TD_Database.m
```diff
@@ -331,9 +331,11 @@ - (BOOL)openWithEncryptionKeyProvider:(id<CDTEncryptionKeyProvider>)provider
 
     __block BOOL result = YES;
     __weak TD_Database* weakSelf = self;
-    [_fmdbQueue inDatabase:^(FMDatabase* db) {
+    [self.fmdbQueue inTransaction:^(FMDatabase* db, BOOL* rollback) {
         TD_Database* strongSelf = weakSelf;
 
+        *rollback = YES;
```
ricellis: Why not add a comment here that the transaction will be rolled back unless the entire block completes successfully when rollback will be set to NO?
I think that will give us a nice compromise between having all the extra ` = YES` lines and being clear about what is going to happen with the transaction.

CDTDatastore/touchdb/TD_Database.m
```diff
@@ -528,6 +535,8 @@ INSERT INTO INFO (key, value) VALUES ('publicUUID',  '%@')",
             result = NO;
             return;
         }
+
+        *rollback = NO;
```
ricellis: Add comment something like `// Block completed successfully, no need to rollback`

rhyshort: done in 8336484

CDTDatastore/touchdb/TD_Database.m
rhyshort: done in 8336484

ricellis:
+1

Issue #287, opened 2016-06-23
Validate attachment dict key matches attachment name


ricellis:
Attachments are stored in a dict of `{name : att}`, whilst the example code demonstrates using this dict with `{att.name : att}` there is no restriction on doing `{@"otherName" : att}` which gives the impression that the attachment name is "otherName", but in fact will still be whatever `att.name` is. We should enforce the dict key being the same as the attachment name.

rhyshort:
I think there are two parts to this, we can validate and log warnings for 1.x.x and then for 2.x.x we can then error out.

rhyshort:
Merged work to print a warning, when we start working towards 2.0 we can re look at this issue and make it an error.

ricellis:
This library is now end-of-life and this issue won't be fixed.
This issue has been [tagged `EOL`](https://github.com/cloudant/CDTDatastore/issues?q=is%3Aissue+label%3AEOL) in case any forks want to import it.

PR #325, opened 2016-09-28
Validate Attachment Dictionary

rhyshort:
## What
Validate Attachment Dictionary on CDTDocumentRevision objects to ensure they
are keyed by the attachment name. 

## How
Added a new method in the CDTDatastore class which enumerates over the key value pairs in the dictionary, returning false if one of the entires is not keyed on the attachment name.
To keep API compatibility, this change only logs a message at the "Warning" level. When we are working towards a 2.0.0 release we can then update it to return an error in the event the attachments dictionary is not keyed correctly.

## Issues 

#287

### DIFF
--- a/CDTDatastore/CDTDatastore.m
+++ b/CDTDatastore/CDTDatastore.m
@@ -451,6 +451,25 @@ NSString *const CDTDatastoreChangeNotification = @"CDTDatastoreChangeNotificatio
 {
     return [_database openWithEncryptionKeyProvider:self.keyProvider];
 }
+/**
+ * Validates that attachments are keyed by attachment name.
+ * @param attachments The attachments dictionary to validate.
+ * @return True if the attachments dictionary is of the form @[ attachment.name : attachment ],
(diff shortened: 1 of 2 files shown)


CDTDatastore/CDTDatastore.m
```diff
@@ -451,6 +451,24 @@ - (BOOL)ensureDatabaseOpen
 {
     return [_database openWithEncryptionKeyProvider:self.keyProvider];
 }
+/**
+ * Validates that attachments are keyed by attachment name.
+ * @param attachments The attachments dictionary to validate.
+ * @return true if the attachments dictionary is of the form @[ attachment.name : attachment ] false
```
emlaver: Small fix to capitalize True and possibly add comma:
`True if the attachments dictionary is of the form @[ attachment.name : attachment ], and false otherwise.`

emlaver:
Other than the comment about the documentation, looks good +1

PR #326, opened 2016-09-29
Update Sample Application

### LINKED ISSUE #266: Update example app to "fire and forget" the CDTReplicator instance
Update the sample app to no longer retain the CDTReplicator instance now that it supports "fire and forget"

rhyshort:
## What

Update Sample Application to use the one line replication methods.

## How

- Remove TodoReplicator
- Update methods which called TodoReplicator to use CDTDatastore to create replications.

## Issues
Resolves #266


Project/Project/CDTTodoReplicator.m
```diff
@@ -1,203 +0,0 @@
...
-
--(void)replicatorDidChangeProgress:(CDTReplicator *)replicator
-{
-
-    NSString *state = [CDTReplicator stringForReplicatorState:replicator.state];
-    [self log:@"%@ changes total %ld. changes processed %ld. state: %@ (%d)", replicator,
-     replicator.changesTotal, replicator.changesProcessed, state, replicator.state];
-}
-
-- (void)customiseNSURLSessionConfiguration:(nonnull NSURLSessionConfiguration *)config {
```
ricellis: I'm concerned about this deletion because it is part of the [replication policies example](https://github.com/cloudant/CDTDatastore/blob/master/doc/replication-policies.md#example). I think we need to make sure that the replication policy example is maintained in the new version.

rhyshort: I'm wondering if it is wise to include both ways of replicating the sample application, it could lead to confusion. Perhaps what we should do is remove the reference to the sample project, with the aim that the sample project is a quick start, rather than show off all the advanced features such as polices.

ricellis: I'm ok with that, as long as the discrepancy is fixed and as long as replication policies still work.
@brynh wdyt?

rhyshort: Replication policies should still work, since the changes for making a replication a one liner are just friendly wrapper functions.

rhyshort: done in eedc278

ricellis: I can't claim to be familiar with the sample app's use of replication policies, but I'm concerned about the sample continuing to use the `periodicReplication` function without using the background stuff that was there previously.

Do we not need to undo more of the changes introduced by https://github.com/cloudant/CDTDatastore/commit/0f0f269f1ac8984f22856c84ec2cb036eedadfba if this sample is  supposed to represent only the simple case again and not replication policies?

rhyshort: We can unpick more of those changes, so sync only occurs on the users actions. I don't see a reason not to, do you @brynh ?

rhyshort: removed the timer code from the sample in 4728613

PR #327, opened 2016-09-30
Fix Replication Acceptance warnings

### LINKED ISSUE #313: Warnings in ReplicationAcceptance
The class `CDTRATestContext` has a few warnings present during compilation. This is due to an API change as part of #292 with the interceptor context.

rhyshort:
## What

Fix warnings in replication acceptance tests

## How
- Fix the public API for CDTHTTPInterceptorContext
- Expose state dictionary via the state property, which creates a immutable copy
of the state dictionary.

- Rename the internal state dictionary from state to internalState. Also access this
via the property rather than the iVar.

- Update Rakefile to use `iphone 5` for iOS tests now that the iPhone 4S does not support iOS 10

## Issues 

Fixes #313


emlaver:
LGTM +1

PR #328, opened 2016-10-10
Make it possible to run Replication Acceptance Tests

rhyshort:
## What
Make it possible to run Replication Acceptance tests

## How

Make it possible to run RA by adding a host application with ATS exceptions.
Also modify the ReplicationSettings.plist in the source tree rather than the bundle
because sometimes modifying it in the bundling causes build failures because the file
is not in the expected location.


CDTDatastoreReplicationAcceptanceTests/ReplicationSettings.plist
```diff
@@ -2,21 +2,21 @@
...
 <dict>
-	<key>TEST_COUCH_USERNAME</key>
-	<string></string>
-	<key>TEST_COUCH_PASSWORD</key>
-	<string></string>
 	<key>TEST_COUCH_HOST</key>
-	<string>localhost</string>
-	<key>TEST_COUCH_PORT</key>
-	<string>5984</string>
+	<string>cloudantsync002.bristol.uk.ibm.com</string>
```
ricellis: Do these plist changes really need to be checked in, given that it gets modified anyway?
I can see that some of them are perhaps more suitable defaults, but I'd be reluctant for a host name to be checked in.

rhyshort: *facedesk* they shouldn't have been checked in. I'll go and correct.

rhyshort: removed in e837bce

Issue #288, opened 2016-06-23
Replication history for checkpoint documents


ricellis:
During replication, the local and remote checkpoint documents should be compared to make sure the last sequence number contained in each matches before that sequence number is used as the starting point for reading the source database's change feed.

Both documents need to be checked, because otherwise one of the source or target databases my have been changed (e.g., deleted and recreated) in such a way that replication needs to restart before it can be carried out correctly.

This can cause problems if a checkpoint document is only written to one database or the other. Say a local checkpoint document is written, but because a device has gone offline the remote one cannot be written. In this case the local and remote checkpoint documents contain differing last sequence numbers, so the replication starts from 0 in the case of naively comparing just the last sequence numbers in the documents.

CouchDB implements a cleverer algorithm where a history of sequence numbers at replication time are maintained. This means that in the above situation it's possible for the oldest common sequence number to be found and the replication continued from there rather than having to reset completely.
1. Check if the replication code checks for matching sequence numbers in source and target. If not implement that check.
2. Start maintaining history in the checkpoint documents that we write to both ends of the replication process and implement the mechanism to find a latest common version.

rhyshort:
The [CouchDB](http://docs.couchdb.org/en/master/replication/protocol.html#find-common-ancestry) docs describe the algorithm to use, it will require a decent amount of work in the replicator. I'm starting to think that this is another argument to rewrite the replicator stack as a whole.

rhyshort:
Should be done as part of #331. Closing.

PR #331, opened 2016-10-12
Use Replication Protocol V3

rhyshort:
## What

Improve replication performance by being able to identify the last sequence at which CDTDatastore replicated to the remote.
## How
- Change the checkpoint document to match the format of V3 for the replication protocol.
- If the last seq for local and remote don't match for the checkpoint, back track over the history object to identify a seq where they do match.
## Issues

Part of #288

### DIFF
--- a/CDTDatastore/touchdb/TDReplicator.m
+++ b/CDTDatastore/touchdb/TDReplicator.m
@@ -790,13 +790,19 @@ NSString* TDReplicatorStartedNotification = @"TDReplicatorStarted";
                             NSArray<NSDictionary<NSString*, NSObject*>*>* localHistory =
                                 localCheckpoint[@"history"];
 
+                            if (!localHistory) {
+                                CDTLogInfo(CDTREPLICATION_LOG_CONTEXT,
+                                           @"%@: Local checkpoint doc does not contain history,"
+                                           @" falling back to full replication",
+                                           self);
+                            }
+
                             // This assumes the  history array is ordered (most recent -> least
                             // recent)
                             for (NSDictionary<NSString*, NSObject*>* rHistory in remoteHistory) {
                                 NSString* sessionID = (NSString*)rHistory[@"session_id"];
                                 BOOL found = NO;
-                                for (NSDictionary<NSString*, NSObject*>* lHistory in
-                                         remoteHistory) {
+                                for (NSDictionary<NSString*, NSObject*>* lHistory in localHistory) {
                                     if ([lHistory[@"session_id"] isEqual:sessionID]) {
                                         found = YES;
                                         break;
--- a/CDTDatastore/touchdb/TDReplicator.m
+++ b/CDTDatastore/touchdb/TDReplicator.m
@@ -6,6 +6,7 @@
 //  Copyright (c) 2011 Couchbase, Inc. All rights reserved.
 //
 //  Modifications for this distribution by Cloudant, Inc., Copyright (c) 2014 Cloudant, Inc.
+//  Copyright © 2016 IBM Corporation. All rights reserved.
 //
 //  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
 //  except in compliance with the License. You may obtain a copy of the License at
--- a/CDTDatastore/touchdb/TDReplicator.m
+++ b/CDTDatastore/touchdb/TDReplicator.m
@@ -16,22 +16,24 @@
 //  and limitations under the License.
 
 #import "TDReplicator.h"
-#import "TDPusher.h"
-#import "TDPuller.h"
-#import "TD_Database+Replication.h"
-#import "TDRemoteRequest.h"
-#import "TDAuthorizer.h"
-#import "TDBatcher.h"
-#import "TDInternal.h"
-#import "TDMisc.h"
-#import "TDBase64.h"
-#import "TDCanonicalJSON.h"
-#import "TDReachability.h"
-#import "MYURLUtils.h"
+#import <Foundation/Foundation.h>
 #import "CDTLogging.h"
 #import "CDTURLSession.h"
 #import "CollectionUtils.h"
+#import "MYURLUtils.h"
+#import "TDAuthorizer.h"
+#import "TDBase64.h"
+#import "TDBatcher.h"
+#import "TDCanonicalJSON.h"
+#import "TDInternal.h"
+#import "TDMisc.h"
+#import "TDPuller.h"
+#import "TDPusher.h"
+#import "TDReachability.h"
+#import "TDRemoteRequest.h"
+#import "TD_Database+Replication.h"
 #import "Test.h"
+
 #if TARGET_OS_IPHONE
 #import <UIKit/UIApplication.h>
 #endif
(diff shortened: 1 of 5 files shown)


CDTDatastore/touchdb/TDReplicator.m
```diff
@@ -819,9 +907,14 @@ - (void)saveLastSequence
                       // read-only and don't attempt to read its checkpoint next time.
                   } else if (_db) {
                       id rev = response[@"rev"];
+                      id ID = response[@"id"];
                       if (rev) body[@"_rev"] = rev;
+                      if (ID) body[@"_id"] = ID;
                       self.remoteCheckpoint = body;
-                      [_db setLastSequence:_lastSequence withCheckpointID:checkpointID];
+                      if (ID && [self.db saveCheckpointDocument:body error:nil]) {
+                          CDTLogWarn(CDTREPLICATION_LOG_CONTEXT,
```
ricellis: We have had issues before when we just log when things go wrong instead of propagating an error. I guess this is an awkward one because failing to checkpoint won't have impacted the completion of the replication, but will have implications for the next replication.
Do you think we should error here?

CDTDatastore/touchdb/TDReplicator.m
```diff
@@ -765,13 +768,52 @@ - (void)fetchRemoteCheckpointDoc
                     if (error.code == kTDStatusNotFound) [self maybeCreateRemoteDB];
                     response = $castIf(NSDictionary, response);
                     self.remoteCheckpoint = response;
-                    NSObject* remoteLastSequence = response[@"lastSequence"];
+                    NSObject* remoteLastSequence = response[@"source_last_seq"];
+                    if (!remoteLastSequence) {
+                        // the checkpoint doc is the old format, it will be converted when we
+                        // resave, for now
+                        // fall back to the `lastSequence` property
+                        remoteLastSequence = response[@"seq"];
```
ricellis: https://github.com/cloudant/CDTDatastore/pull/331/files#diff-7cda42788bca18e1e304c7a20a5c7c72R85 seems to suggest it may be called `seq`

CDTDatastore/touchdb/TD_Database+Replication.m
```diff
@@ -81,28 +83,38 @@ - (NSObject*)lastSequenceWithCheckpointID:(NSString *)checkpointID
...
 {
-    // nothing to save so return early
-    if (lastSequence == nil) {
-        return;
+    NSParameterAssert(checkpoint);
+    NSParameterAssert(checkpoint.count > 0);
+
+    NSString *remote = (NSString *)checkpoint[@"_id"];
+    // 8 is the length of _local/ +1
+    remote = [remote substringFromIndex:8];
```
ricellis: Is this not zero indexed i.e. should this be 7?

CDTDatastore/touchdb/TDReplicator.m
```diff
@@ -800,10 +863,35 @@ - (void)saveLastSequence
...
 
     CDTLogInfo(CDTREPLICATION_LOG_CONTEXT, @"%@ checkpointing sequence=%@", self, _lastSequence);
-    NSMutableDictionary* body = [_remoteCheckpoint mutableCopy];
-    if (!body) body = $mdict();
-    [body setValue:_lastSequence forKey:@"lastSequence"];
-    
+    NSMutableDictionary* body = [self.remoteCheckpoint mutableCopy];
+
+    // upgrade the replication doc to the new format.
+    if (body && !body[@"history"]) {
```
rhyshort: It can do, but I think the code will be same, since we don't need to check the number since we haven't saved one in the docs before.

CDTDatastore/touchdb/TDReplicator.m
```diff
@@ -800,10 +863,35 @@ - (void)saveLastSequence
...
+        NSMutableArray<NSDictionary<NSString*, NSObject*>*>* history = [@[ @{
+            @"session_id" : [[NSUUID UUID] UUIDString],
+            @"recorded_seq" : body[@"lastSequence"]
+        } ] mutableCopy];
+        body[@"history"] = history;
+        body[@"replication_id_version"] = @(3);
+        body[@"session_id"] = history[0][@"session_id"];
+        body[@"source_last_seq"] = body[@"lastSequence"];
+        body[@"lastSequence"] = nil;
```
rhyshort: That is removing, its removing it from the dictionary so we save it will no longer be present.

CDTDatastore/touchdb/TD_Database+Replication.m
rhyshort: fixed in 2dc3000

CDTDatastore/touchdb/TDReplicator.m
rhyshort: I'm not sure, it is a though one, I think we perhaps should error out, but with a specific error to indicate that the checkpoint wasn't written. I'm not sure if this is something we could do without a major release though.

CDTDatastore/touchdb/TDReplicator.m
ricellis: Ah I see the dictionary uses `nil` for removal I thought this was setting a `nil` value (but I see that is done differently) and then relying on not writing a JSON null to not get this in the output.

CDTDatastore/touchdb/TDReplicator.m
ricellis: ok let's leave the behaviour consistent. I guess worst case we just end up replicating some stuff again - as long as we are at least confident that this isn't always failing "silently" in our tests say.

CDTDatastore/touchdb/TD_Database+Replication.m
```diff
@@ -96,8 +96,8 @@ - (BOOL)saveCheckpointDocument:(NSDictionary<NSString *, NSObject *> *)checkpoin
     NSParameterAssert(checkpoint.count > 0);
 
     NSString *remote = (NSString *)checkpoint[@"_id"];
-    // 8 is the length of _local/ +1
-    remote = [remote substringFromIndex:8];
+    // 7 is the length of _local/ +1
```
ricellis: The +1 in the comment is confusing now since 7 is the length of `_local/`.

CDTDatastore/touchdb/TDReplicator.m
```diff
@@ -765,13 +768,52 @@ - (void)fetchRemoteCheckpointDoc
...
+                                    }
+                                }
+
+                                if (found) {
+                                    self.lastSequence = rHistory[@"recorded_seq"];
+                                    break;
+                                }
+                            }
+                        } else {
+                            CDTLogInfo(CDTREPLICATION_LOG_CONTEXT,
```
ricellis: @rhyshort did you miss this one?

CDTDatastore/touchdb/TD_Database+Replication.m
rhyshort: fixed in fc36cf4

ricellis:
The copyrights need updating

rhyshort:
@ricellis done.

ricellis:
I'm ok with this now, but it will need a second reviewer.

CDTDatastore/touchdb/TDReplicator.m
emlaver: Should we still update this message?

rhyshort: @emlaver I don't think this message needs updating.

rhyshort: @ricellis Yes I missed this one. We could put a log message in for the local checkpoint missing the history element, but since we are in control of the local doc, I think we are unlikely to encounter it missing after its been upgraded, but then again, if its unlikely to happen there is probably a customer who is going to hit it. So I'll put one in anyway.

rhyshort: done in 5ad8bc3

rhyshort:
Will merge once we have the tests running again.

rhyshort:
Rebased with latest changes which fix the build with cocoapods 1.1.1

rhyshort:
@ricellis @emlaver 

Tests are passing since the rebase, are you both still happy I merge this?

ricellis:
yes

rhyshort:
since @emlaver is on ops, going to assume its still fine.

PR #334, opened 2016-10-24
Fix attachment key dangling reference

p2k:
On line 125, a reference to the "key" column's data is created which was not properly copied when stored in the CDTSavedAttachment instance on line 145. This leads to FMDB re-using the memory if there is more than one attachment and in the "best" case leads to all attachments having the same key and in the worst case crashes the application with a segfault when working with the "key" attribute.


rhyshort:
Thanks for this,  just a little bit of admin, can you sign the CLA as mentioned in [CONTRIBUTING.MD](https://github.com/cloudant/CDTDatastore/blob/master/CONTRIBUTING.md) so I can merge it?

p2k:
I'm a member of MeetNow! GmbH and took care that the pull request runs under my company's name. We should already have signed a CLA in March this year.

rhyshort:
Confirmed you have permission as part of the MettNow! CLA. Thanks again @p2k.

PR #337, opened 2016-11-02
Compatibility with CocoaPods 1.1.1

rhyshort:
## What

 Cocoapods 1.1.1 introduced changes to pods can be installed with some targets such as app extensions and frameworks. As a result Our development environment requires app targets to embed the different dependancies for the pod depending on the configuration.

## How

Added 4 new app targets, 2 for iOS and 2 for macOS. Each app has an unencrypted and encrypted variant to allow testing with SQLCipher enabled CDTDatastore.

## Tests

All tests pass part from the the two tests which require binding to an arbitrary port. It appears by using an app for macOS it triggers the firewall prompt.  

## Jenkins

Jenkins tests do not pass, however travis does, I haven't been able to determine why one fails and the other doesn't


ricellis:
As discussed I think we need:
i) to identify if the encryption test issue is just Jenkins by fixing the fact they aren't running in Travis.
ii) get an issue open for fixing the disabled tests that need a local server if that is no longer possible in this configuration.

rhyshort:
raised issue #338  for ii). Waiting on the first build of the full matrix to run through travis, so far it is still in the queue.

CDTDatastoreCLEncrpytedTest/main.swift
```diff
@@ -0,0 +1,12 @@
+//
+//  main.swift
+//  CDTDatastoreCLEncrpytedTest
```
ricellis: This has a typo in the name `CDYDatastoreCLEncr**py**tedTest`. There appears to be a duplicate with the correct name `CDTDatastoreCLEncryptedTest/main.swift`

PodFile
```diff
@@ -1,16 +1,5 @@
...
-    pod 'CocoaLumberjack', '~> 2.0'
-end
-
-def import_encrypted_base_pods
-    pod 'FMDB/SQLCipher', '= 2.6'
-    pod 'SQLCipher/fts', '~> 3.1.0'
-    pod 'CocoaLumberjack', '~> 2.0'
-end
-
 def import_test_pods
```
ricellis: Seems like this could be simplified to:
```ruby
def import_test_pods
    import_encrypted_test_pods
    pod 'MRDatabaseContentChecker'
end
```

Although the Podfile guide seems tot suggest using `abstract_target` e.g.

```ruby
abstract_target 'test_pods' do
    pod 'Specta'
    pod 'Expecta'
    pod 'OCMock'
    pod 'OHHTTPStubs'

    target 'CDTDatastoreTests' do
        pod 'MRDatabaseContentChecker'
        inherit! :complete
    end

    target 'CDTDatastoreEncryptionTests' do
        inherit! :complete
    end
end
```

rhyshort:
Superseded by #344

PR #340, opened 2016-11-25
Restore encryption tests

rhyshort:
## What
Restore encryption tests that were used for v1.0.0.

## How
Since the files were named the same but they contain different tests for encrypted and
unencrypted tests, the files have been merged with the tests compile
time macros to ensure the tests will not get accidentally deleted in future.

To support this targets have been reworked into a single set of targets capable of both encrypted and unencrypted depending on the environment on which they are run. 

The new targets are 

- CDTDatastore
- CDTDatastoreOSX
- CDTDatastoreTests
- CDTDatastoreTestsOSX
- CDTDatastoreReplicationAcceptance
- CDTDatastoreReplicationAcceptanceOSX

## Note

Note: Without subsequent PRs the automated build will not work.


rhyshort:
The three tests that only ran for uncencryoted database but now running for encrypted too wasn't really intentional, when making sure I got the tests right I ran them to make sure they are in the correct configuration. Those tests pass in both environments so I didn't see a reason to exclude them from one of the environments, but I'm happy to enable them only for one Env if that is preferred. I'll look into the other missing test.

ricellis:
Well if they work I think we can leave them - the fewer tests we have to special case the better.

rhyshort:
@ricellis restored missing test

PR #341, opened 2016-11-25
Simplify podfile and update for new targets

rhyshort:
Simplify podfile and update for new targets

Simplify podfile so it uses abstract targets rather than methods to define
common dependencies. Also make it possible to change the variants of pods
that are used depending on the environment, for example if encryption is
being used pods which enable encryption and testing of encrypted datastores
is pulled in.

### DIFF
--- a/PodFile
+++ b/PodFile
@@ -1,100 +1,54 @@
 use_frameworks!
 
-def import_base_pods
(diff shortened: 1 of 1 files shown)


rhyshort:
Wrong base branch

PR #342, opened 2016-11-25
Simplify podfile and update for new targets

rhyshort:
Simplify podfile so it uses abstract targets rather than methods to define
common dependencies. Also make it possible to change the variants of pods
that are used depending on the environment, for example if encryption is
being used pods which enable encryption and testing of encrypted datastores
is pulled in.

Notes, automated build won't work until the PR for the rakefile is merged.

### DIFF
--- a/PodFile
+++ b/PodFile
@@ -1,68 +1,54 @@
 use_frameworks!
 
-def import_base_pods
-    pod 'FMDB', '= 2.6'
+if ENV["encrypted"]
+    MRDatabaseContentChecker = "MRDatabaseContentChecker/SQLCipher"
+    FMDB = "FMDB/SQLCipher"
+else
+    MRDatabaseContentChecker = "MRDatabaseContentChecker"
+    FMDB = "FMDB"
+end
+
+abstract_target 'base' do
+    pod FMDB, '= 2.6'
     pod 'CocoaLumberjack', '~> 2.0'
-end
+    pod 'SQLCipher/fts', '~> 3.1.0' if ENV["encrypted"]
 
-def import_encrypted_base_pods
-    pod 'FMDB/SQLCipher', '= 2.6'
-    pod 'SQLCipher/fts', '~> 3.1.0'
-    pod 'CocoaLumberjack', '~> 2.0'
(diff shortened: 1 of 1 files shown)


PodFile
```diff
@@ -1,68 +1,54 @@
...
-target :CDTDatastoreReplicationAcceptanceTestsOSX do
-    platform :osx, '10.9'
-    import_RA_pods
-    import_encrypted_base_pods
+    pod 'SQLCipher/fts', '~> 3.1.0' if ENV["encrypted"]
+
+    target :CDTDatastore do
+        platform :ios, '7.0'
+    end
+    target :CDTDatastoreOSX do
```
ricellis: shouldn't we be calling this `macOS` now, even if the podfile platform still has to be called `osx`?

PodFile
```diff
@@ -1,68 +1,54 @@
...
+        pod 'Expecta'
+        pod 'OCMock'
+        pod 'OHHTTPStubs'
+        pod MRDatabaseContentChecker, :git => 'https://github.com/rhyshort/MRDatabaseContentChecker.git'
+
+        target :CDTDatastoreTests do
+            platform :ios, '7.0'
+        end
+
+        target :CDTDatastoreTestsOSX do
```
ricellis: If the we change to `macOS` this needs to change to match.

PodFile
```diff
@@ -1,68 +1,54 @@
...
+        end
+
+    end
+
+    abstract_target 'raTests' do
+        pod "Unirest", :git => 'https://github.com/rhyshort/unirest-obj-c.git'
+        pod 'TRVSMonitor'
+        pod 'NSData+Base64'
+
+        target :CDTDatastoreReplicationAcceptanceTestsOSX do
```
ricellis: ditto to match

PodFile
rhyshort: Yes we could, but right now the target in the project is still `CDTDatastoreOSX`. I didn't think there was point in changing the name, I'm sure people will still refer to `macOS` as `OSX` for a while anyway.

PR #343, opened 2016-11-25
Update RakeFile for new targets

rhyshort:
Update RakeFile for new targets and include the
-DEncryptDatabase option for when the environment
is encrypted.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,10 @@ osx_image: xcode8
 before_install:
   - export LANG=en_US.UTF-8
 install:
-  - gem install cocoapods -v 1.0.1 --no-document
+  - gem install cocoapods --no-document
   - gem install xcpretty --no-document
   - pod install
-  - cd project && pod _1.0.1_ install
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -5,10 +5,10 @@ osx_image: xcode8
 before_install:
   - export LANG=en_US.UTF-8
 install:
-  - gem install cocoapods -v 1.0.1 --no-document
+  - gem install cocoapods --no-document
   - gem install xcpretty --no-document
   - pod install
-  - cd project && pod _1.0.1_ install
+  - cd project && pod install
```
ricellis: I think this should be changed to **upper P** `Project` to match the case of the on disk folder and the usage in the `Rakefile`. Whilst we're getting away with it now I'd rather not have to try and find/fix if Travis ever change the case sensitivity of their OSX images (since it is apparently an option).

rhyshort: done in 4f14b14

PR #344, opened 2016-11-25
Enable encryption

rhyshort:
Enable encryption tests on travis, to do this the additional environment variable "encrypted" is required to be set, as a result travis will now contain 4 builds for CDTDatastore, 1 encrypted and 1 vanilla for both iOS and OSX.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,4 +14,8 @@ script:
   - rake travis
 env:
   - PLATFORM=iOS
+    encrypted=yes
+  - PLATFORM=iOS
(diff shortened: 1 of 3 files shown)


rhyshort:
@ricellis done, thought I had done that in the last PR.

brynh:
It looks like the encryption tests aren't running on OSX. They don't seem to have target membership of CDTDatastoreTestsOSX.

rhyshort:
@brynh fixed in 154f11b

PR #350, opened 2016-12-06
Use unique paths for encrypted databases.

### LINKED ISSUE #349: EncryptionTests should use generated DB names
EncryptionTests should use generated names to stop them clashing if the tests are being run at the same time as unencrypted tests.

rhyshort:
Use unique paths for encrypted databases to stop path clashes
when encryption and unencrypted tests run on the same machine.

Fixes #349

### DIFF
--- a/CDTDatastoreTests/Encryption/TD_DatabaseEncryptionTests.m
+++ b/CDTDatastoreTests/Encryption/TD_DatabaseEncryptionTests.m
@@ -17,12 +17,13 @@
 
 #import "CDTEncryptionKeyNilProvider.h"
 #import "CDTHelperFixedKeyProvider.h"
+#import "CloudantSyncTests.h"
 #import "CloudantTests+EncryptionTests.h"
 #import "FMDatabase+SQLCipher.h"
 
 #import "TD_Database.h"
 
-@interface TD_DatabaseEncryptionTests : CloudantTests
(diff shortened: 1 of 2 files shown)


CDTDatastoreTests/Encryption/TD_DatabaseEncryptionTests.m
```diff
@@ -17,12 +17,13 @@
...
 #import "CDTEncryptionKeyNilProvider.h"
 #import "CDTHelperFixedKeyProvider.h"
+#import "CloudantSyncTests.h"
 #import "CloudantTests+EncryptionTests.h"
 #import "FMDatabase+SQLCipher.h"
 
 #import "TD_Database.h"
 
-@interface TD_DatabaseEncryptionTests : CloudantTests
+@interface TD_DatabaseEncryptionTests : CloudantSyncTests
```
tomblench: It's a shame this needs to inherit from `CloudantSyncTests` just for `createTemporaryDirectoryAndReturnPath`.

It's slightly out of scope but I wonder whether we should consider moving this method into `CloudantTests`.

rhyshort: I'll raise an issue for it, because it might be worth merging those two into one.

PR #357, opened 2017-03-15
Added per node host locking for pod file

ricellis:
*What*

Added per node host locking for pod file

*How*

Previously we used a single Jenkins lock to prevent clashing pod updates since we only had one host. Now there are multiple hosts so we should lock per host instead of globally to avoid contention.
Split the `NODE_NAME` on the `-` and prefix that part (the host) to `pod` to get the name of the lock for the node.

Note: the old `pod` lock will be deleted when this merges so other branches using the old lock won't be able to clash, but they will have to update to the new `Jenkinsfile` to pass.

*Testing*

Build change, no new tests, ran correctly obtaining new locks.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,6 +16,11 @@
 
 def podfile(podfileDir) {
     // Lock the pod repo and update the pod
+
+    // Note that each macOS host has a podfile so there is one Jenkins lock
+    // available per host. Nodes are named as nodeHost-executorEnv so we use the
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -16,7 +16,7 @@
 
 def podfile(podfileDir) {
     // Lock the pod repo and update the pod
```
emlaver: It would be good to update this comment with something that focuses on "lock the pod repo per host instead of globally to avoid contention".

PR #381, opened 2017-08-29
371 check 403 credentials expired

### LINKED ISSUE #371: Cookie interceptor should check for 403 "credentials_expired" case
The interceptor should deal with this quirk in the session code in the same manner as the java one. This means attempting to deserialise the message as well as looking at the HTTP status code.

emlaver:
## What

Check for 403 HTTP `credentials_expired` case.

## How

- When 403 `credentials_expired` error occurs, call `session` endpoint and renew cookie.

## Testing
- Test case for 403 `credentials_expired`:  get resource and response returns 403, then call session endpoint and get a new cookie.
- Test case for 403 non-expiry: get resource and response returns 403, don't call session and keep existing cookie.
- Test case with cookie flow design similar to java-cloudant's `basic403Test` method

## Issues

fixes #371

### DIFF
--- a/CDTDatastore/HTTP/CDTSessionCookieInterceptorBase.m
+++ b/CDTDatastore/HTTP/CDTSessionCookieInterceptorBase.m
@@ -74,11 +74,30 @@ static const NSInteger CDTSessionCookieRequestTimeout = 600;
  */
 - (CDTHTTPInterceptorContext *)interceptResponseInContext:(CDTHTTPInterceptorContext *)context
 {
-    if (context.response.statusCode == 401 && self.shouldMakeSessionRequest) {
+    bool retryAndAttemptNewSession = NO;
+    if (self.shouldMakeSessionRequest) {
+        if (context.response.statusCode == 401) {
+            retryAndAttemptNewSession = YES;
+        } else if (context.response.statusCode == 403) {
+            NSError *error = nil;
+            NSDictionary *statusCodeMessage = [NSJSONSerialization JSONObjectWithData:context.responseData
+                                                                              options:0
+                                                                                error:&error];
+            if(!error) {
+                NSString *http403Error = [statusCodeMessage objectForKey:@"error"];
+                if([http403Error isEqualToString:@"credentials_expired"]) {
+                   retryAndAttemptNewSession = YES;
+                }
+            }
+        }
+    }
+    
+    if (retryAndAttemptNewSession) {
         // Clear the cookies as we are no longer authorized
         _cookies = nil;
         context.shouldRetry = YES;
     }
+
     // A sliding window may send an early cookie refresh which may save us a _session round-trip
     if (context.response.allHeaderFields[@"Set-Cookie"]) {
         // Replace the cookies with any sent on the response
@@ -142,12 +161,25 @@ static const NSInteger CDTSessionCookieRequestTimeout = 600;
                                           self.shouldMakeSessionRequest = NO;
                                       } else {
                                           // Most other HTTP status codes are non-transient failures; don't retry.
-                                          CDTLogError(CDTREPLICATION_LOG_CONTEXT,
+                                          NSString *dataJsonResponse = [[NSString alloc] initWithData:data
+                                                                                             encoding:NSUTF8StringEncoding];
+                                          if(dataJsonResponse) {
+                                            CDTLogError(CDTREPLICATION_LOG_CONTEXT,
+                                                      @"Failed to get cookie from the server at %@, "
+                                                      @"response code %ld, response message: %@. Cookie "
+                                                      @"authentication will not be attempted again by this interceptor "
+                                                      @"object",
+                                                      url,
+                                                      (long)httpResp.statusCode,
+                                                      dataJsonResponse);
+                                          } else {
+                                            CDTLogError(CDTREPLICATION_LOG_CONTEXT,
                                                       @"Failed to get cookie from the server at %@, response code %ld. Cookie "
                                                       @"authentication will not be attempted again by this interceptor "
                                                       @"object",
                                                       url,
                                                       (long)httpResp.statusCode);
+                                          }
                                           self.shouldMakeSessionRequest = NO;
                                       }
                                       
--- a/CDTDatastoreTests/CDTSessionCookieInterceptorTests.m
+++ b/CDTDatastoreTests/CDTSessionCookieInterceptorTests.m
@@ -137,6 +137,28 @@ static const NSString *testCookieHeaderValue2 =
 }
 
 /**
+ * Test cookie flow, where cookie credential expires
+ * - GET a resource on the cloudant server
+ * - Cookie jar empty, so get session cookie
+ * - Verify that credentials expired with 403 HTTP response and cookie should renew
+ */
+- (void)test403CredentialsExpiredRenewal
+{
+    [self basic403Test:@"credentials_expired" withReason:@"session expired" expectedRequests:6];
+}
+
+/**
+ * Test cookie flow when 403 is not an error of "credentials_expired"
+ * - GET a resource on the cloudant server
+ * - Cookie jar empty, so get session cookie
+ * - Verify 403 HTTP response and cookie should not renew
+ */
+- (void)testNonExpiry403
+{
+    [self basic403Test:@"403_not_expired_test" withReason:@"example reason" expectedRequests:4];
+}
+
+/**
  * Test cookie flow, where a Set-Cookie header on a (non _session) response
  * pre-emptively renews the cookie.
  * - GET a resource on the cloudant server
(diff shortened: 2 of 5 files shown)


CDTDatastore/HTTP/CDTSessionCookieInterceptorBase.m
```diff
@@ -142,12 +145,24 @@ Handles retrieving a cookie ("logging in") for the credentials this interceptor
                                           self.shouldMakeSessionRequest = NO;
                                       } else {
                                           // Most other HTTP status codes are non-transient failures; don't retry.
-                                          CDTLogError(CDTREPLICATION_LOG_CONTEXT,
+                                          NSString *dataJsonResponse = [[NSString alloc] initWithData:data
+                                                                                             encoding:NSUTF8StringEncoding];
+                                          if(dataJsonResponse) {
+                                            CDTLogError(CDTREPLICATION_LOG_CONTEXT,
```
tomblench: I guess inside here still needs a check for status code 403 and "credentials_expired" in the error response. This will also affect `self.shouldMakeSessionRequest` and the `authentication will not be attempted again by this interceptor` message

CDTDatastoreTests/CDTSessionCookieInterceptorTests.m
```diff
@@ -324,4 +365,91 @@ - (void)testCookieInterceptorRenewsEarly
     XCTAssert([helper currentResponse] == 5);
 }
 
+- (void)basic403Test:(nonnull NSString *)error withReason:(NSString *)reason expectedRequests:(int)expectedRequests
```
tomblench: Will there be other tests which use this method? As it stands it seems a bit pointless with only one call site.

CDTDatastore/HTTP/CDTSessionCookieInterceptorBase.m
emlaver: @tomblench Would it be better for 403 to have it's own if block just like the 401 case?
```
...
else if (httpResp.statusCode == 403) {
  // Credentials expired; retry.
  NSString *dataJsonResponse = [[NSString alloc] initWithData:data
                                                     encoding:NSUTF8StringEncoding];
  if(!dataJsonResponse) {
      dataJsonResponse = @"{\"error\": \"credentials_expired\", \"reason\": \"Session expired\"}";
  }
  
  CDTLogError(CDTREPLICATION_LOG_CONTEXT, @"Credentials expired, response: %@", dataJsonResponse);
  self.shouldMakeSessionRequest = NO;
}
```

tomblench: Yes, makes sense

ricellis: It's worth noting that `403` with `credentials_expired` only applies for `_session` not `_iam_session`. So the `403` block might be better placed in `CDTSessionCookieInterceptor` than in the Base class. I don't think it hurts too much to have it in common, but it will mean doing unnecessary checks on `403` responses for `_iam_session`.

emlaver: Updated in f8fb921.

CDTDatastoreTests/CDTSessionCookieInterceptorTests.m
emlaver: @tomblench I've added a test in e01dc2f for verifying 403 status code that is not `credentials_expired` error.

CDTDatastore/HTTP/CDTSessionCookieInterceptorBase.m
emlaver: @tomblench @ricellis Removed this code after Slack conversation and updated the 403 if block case in `interceptResponseInContext` in ebe1749.

CDTDatastoreTests/CDTSessionCookieInterceptorTests.m
emlaver: @tomblench @ricellis Updated both 403 tests and `basic403Test` method in ebe1749.
.

CDTDatastore/HTTP/CDTSessionCookieInterceptorBase.m
```diff
@@ -78,6 +78,22 @@ - (CDTHTTPInterceptorContext *)interceptResponseInContext:(CDTHTTPInterceptorCon
...
         context.shouldRetry = YES;
+    } else if (context.response.statusCode == 403) {
+        NSError *error = nil;
+        NSDictionary *statusCodeMessage = [NSJSONSerialization JSONObjectWithData:context.responseData
+                                                                        options:0
+                                                                          error:&error];
+        if(!error) {
+            NSString *http403Error = [[NSString alloc] init];
+            http403Error = [statusCodeMessage objectForKey:@"error"];
+            if([http403Error containsString:@"credentials_expired"]) {
```
tomblench: We can probably use `isEqualTo` here as we want to match the exact string

CDTDatastore/HTTP/CDTSessionCookieInterceptorBase.m
```diff
@@ -78,6 +78,22 @@ - (CDTHTTPInterceptorContext *)interceptResponseInContext:(CDTHTTPInterceptorCon
         // Clear the cookies as we are no longer authorized
         _cookies = nil;
         context.shouldRetry = YES;
+    } else if (context.response.statusCode == 403) {
+        NSError *error = nil;
+        NSDictionary *statusCodeMessage = [NSJSONSerialization JSONObjectWithData:context.responseData
+                                                                        options:0
+                                                                          error:&error];
+        if(!error) {
+            NSString *http403Error = [[NSString alloc] init];
```
tomblench: You don't need to allocate this string object, because what you're doing is creating an object and then immediately throwing it away by assigning the pointer to the return value of `statusCodeMessage objectForKey` on the next line.

Just do `NSString *http403Error = [statusCodeMessage objectForKey:@"error"];`

CDTDatastore/HTTP/CDTSessionCookieInterceptorBase.m
```diff
@@ -78,6 +78,22 @@ - (CDTHTTPInterceptorContext *)interceptResponseInContext:(CDTHTTPInterceptorCon
...
+            http403Error = [statusCodeMessage objectForKey:@"error"];
+            if([http403Error containsString:@"credentials_expired"]) {
+                context.shouldRetry = YES;
+            } else {
+                self.shouldMakeSessionRequest = NO;
```
tomblench: This should probably have a matching `authentication will not be attempted again by this interceptor` style message. I wonder whether it would be better to have an [accessor method](https://developer.apple.com/library/content/documentation/General/Conceptual/DevPedia-CocoaCore/AccessorMethod.html) for `shouldMakeSessionRequest` which logs this message whenever it transitions from YES to NO.

CDTDatastoreTests/CDTSessionCookieInterceptorTests.m
```diff
@@ -324,4 +356,146 @@ - (void)testCookieInterceptorRenewsEarly
...
+        }];
+        
+        // get resource successfully using new cookie
+        [helper addResponse:^OHHTTPStubsResponse *__nonnull(NSURLRequest *__nonnull request) {
+            XCTAssert([testCookieHeaderValue2 isEqualToString: request.allHTTPHeaderFields[@"Cookie"]]);
+            XCTAssert([request.HTTPMethod isEqualToString:@"GET"]);
+            return [OHHTTPStubsResponse responseWithJSONObject:@{@"ok" : @(YES)} statusCode:200 headers:@{}];
+        }];
+    } else {
+        // get resource fails, cookie was not re-newed and is empty
```
ricellis: The cookie should not be empty here, it was never expired and so doesn't need renewing. It should still be on the subsequent requests.

CDTDatastoreTests/CDTSessionCookieInterceptorTests.m
```diff
@@ -324,4 +356,146 @@ - (void)testCookieInterceptorRenewsEarly
...
+    }
+    
+    // Second GET should fail with 403 error
+    CDTURLSessionTask *task2 = [session dataTaskWithRequest:request taskDelegate:nil];
+    [task2 resume];
+    while ([task2 state] != NSURLSessionTaskStateCompleted) {
+        [NSThread sleepForTimeInterval:1.0];
+    }
+    
+    if(interceptor.shouldMakeSessionRequest == NO) {
```
ricellis: As per above, if it means what I think it does as opposed to what you think it does then this check won't work once you fix it, you'll need to use something else to distinguish the test paths.

CDTDatastoreTests/CDTSessionCookieInterceptorTests.m
```diff
@@ -324,4 +356,146 @@ - (void)testCookieInterceptorRenewsEarly
...
+        // 403 non-expiry creds. case: cookie should not renew and be empty
+        XCTAssertNil(interceptor.cookies);
```
ricellis: Cookie should still be valid in non-expiry case.

CDTDatastore/HTTP/CDTSessionCookieInterceptorBase.m
emlaver: Fixed in 9b2f6bb.

CDTDatastore/HTTP/CDTSessionCookieInterceptorBase.m
emlaver: Fixed in 9b2f6bb.

CDTDatastore/HTTP/CDTSessionCookieInterceptorBase.m
emlaver: This line was removed in 9b2f6bb.

CDTDatastore/HTTP/CDTSessionCookieInterceptorBase.m
```diff
@@ -78,6 +78,22 @@ - (CDTHTTPInterceptorContext *)interceptResponseInContext:(CDTHTTPInterceptorCon
         // Clear the cookies as we are no longer authorized
         _cookies = nil;
         context.shouldRetry = YES;
+    } else if (context.response.statusCode == 403) {
```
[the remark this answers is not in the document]
emlaver: Fixed in 9b2f6bb.

CDTDatastore/HTTP/CDTSessionCookieInterceptorBase.m
```diff
@@ -78,6 +78,22 @@ - (CDTHTTPInterceptorContext *)interceptResponseInContext:(CDTHTTPInterceptorCon
...
+        NSDictionary *statusCodeMessage = [NSJSONSerialization JSONObjectWithData:context.requestData
+                                                                        options:0
+                                                                          error:&error];
+        if(!error) {
+            NSString *http403Error = [[NSString alloc] init];
+            http403Error = [statusCodeMessage objectForKey:@"error"];
+            if([http403Error containsString:@"credentials_expired"]) {
+                context.shouldRetry = YES;
+            } else {
+                self.shouldMakeSessionRequest = NO;
```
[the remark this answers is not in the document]
emlaver: Updated this whole if/else logic in 9b2f6bb.

CDTDatastore/HTTP/CDTSessionCookieInterceptorBase.m
```diff
@@ -78,6 +78,22 @@ - (CDTHTTPInterceptorContext *)interceptResponseInContext:(CDTHTTPInterceptorCon
...
+            }
+        }
+        // Clear the cookies as we are no longer authorized
+        _cookies = nil;
```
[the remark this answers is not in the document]
emlaver: Added boolean flag in 9b2f6bb.

CDTDatastoreTests/CDTSessionCookieInterceptorTests.m
emlaver: Fixed in 9b2f6bb.

CDTDatastoreTests/CDTSessionCookieInterceptorTests.m
emlaver: Fixed in 9b2f6bb.

CDTDatastoreTests/CDTSessionCookieInterceptorTests.m
emlaver: Fixed in 9b2f6bb.

CDTDatastoreTests/CDTSessionCookieInterceptorTests.m
```diff
@@ -87,6 +87,16 @@ - (void)setUp
           }
 
         }];
+    
+    [OHHTTPStubs stubRequestsPassingTest:^BOOL(NSURLRequest *__nonnull request) {
```
ricellis: I think it is confusing having this stub in the setup when all the other stubs are done in the test case itself for the 403 cases. This stub is needed for your current test, but it is returning a 403 when I think you really want a different response (probably 200 for expired path, and 403 for non-expired path) - I think it would be better to keep the stubs together rather than splitting them like this, which is already causing some unexpected results (see later comments).

CDTDatastoreTests/CDTSessionCookieInterceptorTests.m
```diff
@@ -324,4 +356,146 @@ - (void)testCookieInterceptorRenewsEarly
...
+                                               testCookieHeaderValue2]
+                              }];
+        }];
+        
+        // get resource successfully using new cookie
+        [helper addResponse:^OHHTTPStubsResponse *__nonnull(NSURLRequest *__nonnull request) {
+            XCTAssert([testCookieHeaderValue2 isEqualToString: request.allHTTPHeaderFields[@"Cookie"]]);
+            XCTAssert([request.HTTPMethod isEqualToString:@"GET"]);
+            return [OHHTTPStubsResponse responseWithJSONObject:@{@"ok" : @(YES)} statusCode:200 headers:@{}];
+        }];
```
ricellis: You should add a duplicate of this 200 stub response to respond to "task 3".

CDTDatastoreTests/CDTSessionCookieInterceptorTests.m
```diff
@@ -324,4 +356,146 @@ - (void)testCookieInterceptorRenewsEarly
...
+    } else {
+        // get resource fails, cookie was not re-newed
+        [helper addResponse:^OHHTTPStubsResponse *__nonnull(NSURLRequest *__nonnull request) {
+            XCTAssert([testCookieHeaderValue isEqualToString: request.allHTTPHeaderFields[@"Cookie"]]);
+            XCTAssert([request.HTTPMethod isEqualToString:@"GET"]);
+            return [OHHTTPStubsResponse responseWithJSONObject:@{@"error": error,
+                                                                 @"reason": reason}
+                                                    statusCode:403 headers:@{}];
+        }];
```
ricellis: You should add a duplicate of this 403 stub to respond to "task 3"

CDTDatastoreTests/CDTSessionCookieInterceptorTests.m
```diff
@@ -324,4 +356,146 @@ - (void)testCookieInterceptorRenewsEarly
...
+    CDTURLSession *session = [[CDTURLSession alloc] initWithCallbackThread:[NSThread currentThread] requestInterceptors:@[interceptor] sessionConfigDelegate: nil];
+    
+    // Initial GET will get a cookie and resource
+    CDTURLSessionTask *task = [session dataTaskWithRequest:request taskDelegate:nil];
+    [task resume];
+    while ([task state] != NSURLSessionTaskStateCompleted) {
+        [NSThread sleepForTimeInterval:1.0];
+    }
+    
+    // Second GET should fail with 403 error
```
ricellis: This is not true, the second `GET` (task2) has different results depending on the path.
expired creds -> (under the covers) 403, renew cookie -> 200
non-expired -> 403

It would probably be valuable to assert on the response code for each path.

CDTDatastoreTests/CDTSessionCookieInterceptorTests.m
```diff
@@ -324,4 +356,146 @@ - (void)testCookieInterceptorRenewsEarly
...
+    CDTURLSessionTask *task2 = [session dataTaskWithRequest:request taskDelegate:nil];
+    [task2 resume];
+    while ([task2 state] != NSURLSessionTaskStateCompleted) {
+        [NSThread sleepForTimeInterval:1.0];
+    }
+        
+    if([error isEqualToString:@"credentials_expired"]) {
+        // call to _session endpoint
```
ricellis: This task doesn't call the session endpoint, not even under the covers.

I think task 3 should move out of the `if` because it is actually the same task for both paths through this test. Only the assertions need to be different. You should probably assert the response code from task 3 as well.

CDTDatastoreTests/CDTSessionCookieInterceptorTests.m
```diff
@@ -324,4 +356,146 @@ - (void)testCookieInterceptorRenewsEarly
...
+        [task3 resume];
+        while ([task3 state] != NSURLSessionTaskStateCompleted) {
+            [NSThread sleepForTimeInterval:1.0];
+        }
+        
+        // 403 'credentials_expired' case: call to _session in task3 re-newed cookie
+        XCTAssertNotNil(interceptor.cookies);
+        
+        // GET resource with new cookie
+        CDTURLSessionTask *task4 = [session dataTaskWithRequest:request taskDelegate:nil];
```
ricellis: I'm confused as to why this 4th task even exists and I think the only reason it isn't causing a failure is because of the extra (uneeded) stub inserted in the setup. I think if you looked at the response code for this task it would be a `403` because of that extra stub. As I wrote in earlier comment I think that setup stub should be removed and this 4th task should be removed.

I think 3 tasks is sufficient to exercise the test path you want in each case. I have summarized the flow I would expect in each case which ought to help with getting the right number of stubs down each path:

 Task | Request/Stub responses (expired) | Request/Stub responses (not-expired)
--- | --- | ---
1 `GET http://username403.cloudant.com/somedb` | `GET _session` -> `200 cookie1` <br/> `GET somedb (with cookie1)` -> `200` | `GET _session` -> `200 cookie1` <br/> `GET somedb (with cookie1)` -> `200`
2 `GET http://username403.cloudant.com/somedb` | `GET somedb (with cookie1)` -> `403 {error: "credentials_expired"}` <br/> `GET _session` -> `200 cookie2` <br/> `GET somedb (with cookie2)` -> `200` | `GET somedb (with cookie1)` -> `403 {error: "403_not_expired_test"}`
3 `GET http://username403.cloudant.com/somedb` | `GET somedb (with cookie2)` -> `200` | `GET somedb (with cookie1)` -> `403 {error: "403_not_expired_test"}`
Total requests made/ stubs needed | 6 | 4

CDTDatastoreTests/CDTSessionCookieInterceptorTests.m
```diff
@@ -137,6 +147,28 @@ - (void)testCookieInterceptorHandles401
...
 
 /**
+ * Test cookie flow, where cookie credential expires
+ * - GET a resource on the cloudant server
+ * - Cookie jar empty, so get session cookie
+ * - Verify that credentials expired with 403 HTTP response and cookie should renew
+ */
+- (void)test403CredentialsExpiredRenewal
+{
+    [self basic403Test:@"credentials_expired" withReason:@"session expired" expectedRequests:5];
```
ricellis: I think this should actually be 6. See summary table in my other comment. I think the reason it asserts correctly is because the additional stub in setup isn't counted by the helper - another reason to remove that setup stub.

CDTDatastoreTests/CDTSessionCookieInterceptorTests.m
emlaver: Thanks for catching this - it was a 'left-over' stub from a removed test case.

CDTDatastoreTests/CDTSessionCookieInterceptorTests.m
ricellis: As discussed this isn't needed. Sorry.

CDTDatastoreTests/CDTSessionCookieInterceptorTests.m
emlaver: @ricellis Thanks for spending the time to write out this summary table.  I've created comments using these details in 565045f.

CDTDatastore/HTTP/CDTSessionCookieInterceptorBase.m
```diff
@@ -74,11 +74,30 @@ - (BOOL)hasValidCookieWithName:(nonnull NSString*)cookieName forRequestURL:(nonn
  */
 - (CDTHTTPInterceptorContext *)interceptResponseInContext:(CDTHTTPInterceptorContext *)context
 {
-    if (context.response.statusCode == 401 && self.shouldMakeSessionRequest) {
+    bool retryAndAttemptNewSession = NO;
+    if (self.shouldMakeSessionRequest) {
+        if (context.response.statusCode == 401) {
+            retryAndAttemptNewSession = YES;
+        } else if (context.response.statusCode == 403) {
```
ricellis: @tomblench did you think this block should move down into the `CDTSessionCookieInterceptor`?

PR #390, opened 2017-10-16
Fix build errors in Xcode 9

emlaver:
## What

Resolve build errors:
```
Implicit declaration of function 'sqlite3_rekey' is invalid in C99
Implicit declaration of function 'sqlite3_key' is invalid in C99
```

## How

From [this comment](https://discuss.zetetic.net/t/ios-11-xcode-issue-implicit-declaration-of-function-sqlite3-key-is-invalid-in-c99/2198/27):
```
This appears to be an issue with CocoaPods with use_framworks!, 
which disables the creation of header links under Pods/Headers. 
This makes it impossible for FMDB to resolve the proper sqlite3.h 
file included with SQLCipher.

In this case you can work around this by using a post_install hook
in your podfile to create a link in Pods/Headers/Private which is 
automatically included in the Header Search Path.
```

## Testing

No new tests.

### DIFF
--- a/PodFile
+++ b/PodFile
@@ -52,4 +52,9 @@ abstract_target 'base' do
           platform :ios, '7.0'
         end
     end
+
+    post_install do | installer |
+        print "SQLCipher: link Pods/Headers/sqlite3.h"
+        system "mkdir -p Pods/Headers/Private && ln -s ../../SQLCipher/sqlite3.h Pods/Headers/Private"
+    end
 end


ricellis:
It sounds from those threads like this issue might have been fixed in more recent versions of CocoaPods. I ran using CocoaPods 1.3.1/XCode 9 and observed the same issue. I then did a `pod install` and ran the test again and the problem went away.

ricellis:
Yeah we realized my run yesterday worked because I used the wrong env var `encryption=yes` vs `encrypted=yes` so although the linked threads appeared to point to resolved issues that were included in the newer CocoaPods it doesn't actually fix it :(

As @tomblench said the warning also appears for the macOS build it is just not treated as an error in that case so doesn't fail the build, but it is still incorrectly pointing to the wrong header.

I think we go with the workaround for now.

PR #394, opened 2017-10-30
Threading and Jenkins test improvements

tomblench:
## What

Threading and Jenkins test improvements.

Threading improvements increase overall robustness of replicators when they are started/stopped in a manner which exposes previous race conditions.

Jenkins improvements allow us to see that replication acceptance (RA) tests have passed before merging from master, by executing them with a smaller set of documents and excluding some large tests.

### Threading fixes

- Use a busy-wait loop in `_bulk_get` probe, because blocking on a
  semaphore blocks the run-loop which causes a deadlock when waiting
  on completion handlers for tests like
  `test{Pull,Push}ReplicationUsingOneLiner`.

- Handle error in server response in `saveLastSequence`.

- Execute the final call to `saveLastSequence` synchronously, to
  ensure the sequence number is correct when replication has
  finished. Previously this was the cause of two subtle threading
  errors: i) the `_db` field could be set to `nil` before the response
  callback fired, so the sequence wasn't persisted to the database;
  ii) the response callback never fires because the application or
  test executor exits before this can occur.

- (Related to above) post "stopped" notification after checkpoint
  save. This was a timing bug causing
  `testRemoteLastSequenceValueAfter{Pull,Push}Replication` to
  occasionally fail.

- Synchronise `TDURLConnectionChangeTracker` `start`/`stop`.
  Previously this could cause a crash at the line
  `self.inputBuffer = [NSMutableData dataWithCapacity:0];` because of
  a race condition when stopping soon after starting. The crash
  occurred because the object's state had become invalid after `stop`
  was called, but `start` was still executing.

- Only run `[TDPuller stopped]` once - if it runs more than once we
  could send messages to garbage/nil objects.

- Only run `[TDReplicator stopped]` once - fixes a race condition
  where stopped could run twice and reset `_db` to `nil` before the
  sequence can be saved.

- Add extra logging for `saveLastSequence` success and failure paths.

### Jenkins fixes

- Run 'small' RA tests (fewer documents and some tests excluded) for
  non-master branches.

- Append to logfiles (some configrations like IAM/non-IAM will result
  in the same file name).

- Fix error handling in `saveCheckpointDocument` - error pointer was
  incorrectly passed in and return codes were not being checked.

## Test fixes

- Add default values for IAM and small-RA to plist - they need to be
  present even if they are over-ridden

- `[ReplicationSettings iamApiKey]` returns nil if it's present but
  blank

## Testing

`test{Pull,Push}ReplicationUsingOneLiner` pass, were previously failing on master.

### DIFF
--- a/CDTDatastore/touchdb/TDReplicator.m
+++ b/CDTDatastore/touchdb/TDReplicator.m
@@ -299,9 +299,6 @@ NSString* TDReplicatorStartedNotification = @"TDReplicatorStarted";
                [[NSRunLoop currentRunLoop] runMode: NSDefaultRunLoopMode
                                         beforeDate: [NSDate dateWithTimeIntervalSinceNow:0.1]])
         ;
-
-        // clear the reference to the db
-        _db = nil;
         
         CDTLogInfo(CDTREPLICATION_LOG_CONTEXT, @"TDReplicator thread exiting");
 
@@ -438,22 +435,38 @@ NSString* TDReplicatorStartedNotification = @"TDReplicatorStarted";
 
 - (void)stopped
 {
-    CDTLogInfo(CDTREPLICATION_LOG_CONTEXT, @"%@ STOPPED", self);
-    CDTLogInfo(CDTREPLICATION_LOG_CONTEXT, @"Replication: %@ took %.3f sec; error=%@", self,
-            CFAbsoluteTimeGetCurrent() - _startTime, _error);
-    self.running = NO;
+    @synchronized(self) {
+        // only want stopped to run once
+        if (!_running) {
+            return;
+        }
+        CDTLogInfo(CDTREPLICATION_LOG_CONTEXT, @"%@ STOPPED", self);
+        CDTLogInfo(CDTREPLICATION_LOG_CONTEXT, @"Replication: %@ took %.3f sec; error=%@", self,
+                CFAbsoluteTimeGetCurrent() - _startTime, _error);
+        self.running = NO;
 
-    [[NSNotificationCenter defaultCenter] postNotificationName:TDReplicatorStoppedNotification
-                                                        object:self];
-    [self saveLastSequence];
+        [self saveLastSequence];
+        // wait for saveLastSequence to finish
+        // (we can't clear the reference to _db until we've saved the checkpoint to the database)
+        while (_savingCheckpoint) {
+            [[NSRunLoop currentRunLoop] runMode: NSDefaultRunLoopMode
+                                     beforeDate: [NSDate dateWithTimeIntervalSinceNow:0.1]];
+        }
 
-    _batcher = nil;
-    [_host stop];
-    _host = nil;
-    
-    CDTLogInfo(CDTREPLICATION_LOG_CONTEXT, @"STOP %@", self);
-    [[NSNotificationCenter defaultCenter] removeObserver: self];
-    _replicatorStopped = YES;
+        // post "stopped" notification after saving last sequence number so it's guaranteed to be
+        // up-to-date for anyone waiting on the replicator to stop
+        [[NSNotificationCenter defaultCenter] postNotificationName:TDReplicatorStoppedNotification
+                                                            object:self];
+        // clear the reference to the db
+        _db = nil;
+        _batcher = nil;
+        [_host stop];
+        _host = nil;
+
+        CDTLogInfo(CDTREPLICATION_LOG_CONTEXT, @"STOP %@", self);
+        [[NSNotificationCenter defaultCenter] removeObserver: self];
+        _replicatorStopped = YES;
+    }
 }
 
 -(BOOL) threadExecuting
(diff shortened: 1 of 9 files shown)


CDTDatastoreReplicationAcceptanceTests/ReplicationAcceptance.m
```diff
@@ -1659,6 +1664,10 @@ -(void) doTestMultithreadedPullCancelHalf:(int)nPulls
 // this test is disabled because it causes too many build falures
```
emlaver: Remove this comment unless you are going to re-disable the test.

CDTDatastore/touchdb/TDReplicator.m
```diff
@@ -438,22 +435,38 @@ - (void)stop
 
 - (void)stopped
 {
-    CDTLogInfo(CDTREPLICATION_LOG_CONTEXT, @"%@ STOPPED", self);
-    CDTLogInfo(CDTREPLICATION_LOG_CONTEXT, @"Replication: %@ took %.3f sec; error=%@", self,
-            CFAbsoluteTimeGetCurrent() - _startTime, _error);
-    self.running = NO;
+    @synchronized(self) {
+        // only want stopped to run once
+        if (!_running) {
```
ricellis: I'm a bit concerned that `_running` has such a mix of synch (S) and unsynch (U) access in different places:

Method | `stop` | `stopped` | `goOnline` | `readyIfRetry` | `addToInbox` | `addRevsToInbox`
--- | --- | --- | --- | --- | --- | ---
TDPuller | U | **S** | U | - | - | - 
TDReplicator | **S** | **S** | U | U | U | U 

Let me know if you're comfortable with this or maybe we should do a walkthrough of this particular flag?

Jenkinsfile
```diff
@@ -40,19 +40,23 @@ def buildAndTest(nodeLabel, target, rakeEnv, encrypted, testIam='no') {
 
         // Build and test
         try {
-            def envVariables = []
+            def envVariables = ["TEST_COUCH_LOGGING_LEVEL=31"]
```
ricellis: Is this extra logging left on intentionally, or is the plan to enable it as needed?

Rakefile
```diff
@@ -109,7 +109,7 @@ end
 def run_tests(workspace, scheme, destination)
   settings = "GCC_PREPROCESSOR_DEFINITIONS='${inherited} ENCRYPT_DATABASE=1'" unless !ENV["encrypted"]
   logName = (ENV["encrypted"] ? "Encrypted" : "") + "#{scheme}.log"
-  return system("xcodebuild -verbose -workspace #{workspace} -scheme '#{scheme}' -destination '#{destination}' #{settings} test | tee #{logName} | xcpretty -r junit; exit ${PIPESTATUS[0]}")
+  return system("xcodebuild -verbose -workspace #{workspace} -scheme '#{scheme}' -destination '#{destination}' #{settings} test | tee -a #{logName} | xcpretty -r junit; exit ${PIPESTATUS[0]}")
```
ricellis: I'd prefer to see an independent log name rather than this append. The combinations of which branch runs on which node will potentially mean different orderings or an overwrite when the artifact is attached to Jenkins regardless of whether we appended on the box or not, and that's not accounting for the case where we may not start from a clean checkout.
Whilst this was fine for getting what you needed I think it is likely to come back and bite us later if we don't do something more robust (I'm willing for it to happen under another PR as it is somewhat unrelated having been introduced by the IAM changes rather than this work, and only being exposed by the need to use the logs here).

Jenkinsfile
tomblench: It was to help debug some test failures, but I wonder whether we should leave it in? It's frustrating to have a test failure and not have detailed test logs, especially if it's a transient failure.

The only downside I can is the storage requirements but the log files are only a few megabytes.

ricellis: I'm fine with that, just checking it wasn't an oversight

CDTDatastore/touchdb/TDReplicator.m
tomblench: Good analysis (everyone likes a table), `[TDPuller stop]` should be sync because it also depends on some boolean status flags which may cause race conditions. It seems awkward for a sub-class and a super-class to both have sync for the same methods, but sometimes the super-class method is invoked directly so it's unavoidable.

The `stopped` methods are called by `stop` or `updateActive`, or via the sub-class. Because of the `updateActive` case, they need their own synchronisation. I think `updateActive` itself doesn't need any synchronisation because its main job is just to fire `stopped` when the task count gets to zero.

 `goOnline` and `goOffline` are set through `TDReachability` which I'm pretty sure is only single threaded - there appears to be a run-loop updating the network reachability status.

`retryIfReady` doesn't manipulate any of the boolean status flags but can kick off `beginReplicating` to start the whole thing over. It gets called from `updateActive` but only in the `continuous` state and if there were failed revisions. From the CDT layer we don't support continuous replications (this field is hardcoded to `NO` at construction time), so I think we can safely ignore it.

`addToInbox` is driven from the `_changes` feed so I guess there is only one thread modifying the underlying `TDBatchers`'s `inBox` which is an `NSMutableArray`. (see https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/Multithreading/ThreadSafetySummary/ThreadSafetySummary.html for details of which Cocoa classes are thread safe/unsafe).

`addRevsToInbox` only seems to get called from the `TDPusher` sub-class. Again it seems to only be called in a single-threaded fashion from `processInbox`.

tomblench: Oh I meant to add, in summary I think that `[TDPuller stop]` should be sync but I think we are OK with the others.

Rakefile
tomblench: I think we can find a better way of naming these logs, maybe jenkins should pass in a unique name fore this matrix combination.

CDTDatastore/touchdb/TDReplicator.m
ricellis: Sounds good, thanks for digging through it.

Rakefile
tomblench: As discussed I'll back this out and cover it in another PR.

CDTDatastore/touchdb/TDReplicator.m
emlaver: @tomblench Thanks for explaining and writing this up.

Issue #396, opened 2017-11-06
Jenkinsfile should declare Rakefile log output


tomblench:
The log file name should depend on the axes used - the Jenkinsfile should concatenate these into a name suitable for the log file and pass it to the Rakefile. Currently the Rakefile doesn't take the IAM axis into account, so one set of tests will over-write another.

tomblench:
Closing, this was done in the referenced PR

PR #404, opened 2017-11-16
Fix jenkins logging

tomblench:
## What

Use a more logical log naming scheme and fix a bug where the docs step tries to archive a log which doesn't exist (and silently fails the whole pipeline).

## How

- Define RA/build/test log names in jenkinsfile and pass to rakefile
- buildDocs step uses "CDTDatastore_buildDocs.log" as log name

## Testing

Jenkins runs, logs are there

## Issues

#396

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -109,7 +109,7 @@ end
 def run_tests(workspace, scheme, destination)
   settings = "GCC_PREPROCESSOR_DEFINITIONS='${inherited} ENCRYPT_DATABASE=1'" unless !ENV["encrypted"]
   # if jenkins passed us a log name, use that, otherwise default to "test.log"
-  logName = (ENV["LOG_NAME"] == nil ? "test.log" : ENV["LOG_NAME"])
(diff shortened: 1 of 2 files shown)


Rakefile
```diff
@@ -108,7 +108,8 @@ end
 # Runs `test` target for workspace/scheme/destination
 def run_tests(workspace, scheme, destination)
   settings = "GCC_PREPROCESSOR_DEFINITIONS='${inherited} ENCRYPT_DATABASE=1'" unless !ENV["encrypted"]
-  logName = (ENV["encrypted"] ? "Encrypted" : "") + "#{scheme}.log"
+  # if jenkins passed us a log name, use that, otherwise default to "test.log"
+  logName = (ENV["LOG_NAME"] == nil ? "test.log" : ENV["LOG_NAME"])
```
ricellis: I think for local testing using `rake` it would be easier if the default was `#{scheme}.log"` since the default test runs two schemes for macOS and iOS it would be nice to still get two differently named log files out.

tomblench: Good point, see 3fd8515 and afb2cc4

Issue #400, opened 2017-11-08
CDTReplicationcomplete is received before replication is actually complete.


AjayHegdeK:
#CDTDatastore:  1.2.2
iOS - 9 and above
There are 2 applications both are accessing CDTDatastore in a single device. When I switch between apps I am getting database download complete even before it is actually completed. 
And same issue also occurs when you switch on and off internet frequently.

Detailed log:

#CDTDatastore:  1.2.2
iOS - 9 and above
There are 2 applications both are accessing CDTDatastore in a single device. When I switch between apps I am getting database download complete even before it is actually completed. 
And same issue also clientIDcurs when you switch on and off internet frequently.

Detailed log:

2017-10-27 18:05:57:782 clientID Production[733:188615]

emlaver:
We’ve been unable to reproduce this, but are continuing to investigate.
We recommend that you configure [replication policies](https://github.com/cloudant/CDTDatastore/blob/master/doc/replication-policies.md) for your applications.

AjayHegdeK:
Hi,  I have gone through replication policies and followed the document as is, but still issue exists. Our sync time is around 3 mins we have more 10k documents, can you please try out with minimum 10k documents. 
Step1:  start downloading 
we have used sync mechanism in the following way

func sync(direction: SyncDirection, from: Database, to: Database, completionHandler: ((_ replicationState: ReplicationState) -> Void)?) {
        // Start measuring replication time
        replicationDuration = Duration().startMeasuring("\(direction.rawValue) replication for \(to.rawValue)")
        // Sync local data to or from the central cloud.
        replication = SyncManager.replications[direction]
        guard replication == nil else {
            Log.debug("Ignore \(direction) replication; already running")
            return
        }
        
        semaphore = DispatchSemaphore(value: 0)
        askForAssertionWithSemaphore(semaphore: semaphore)
        
        // Store completion handler for callback purpose
        self.completionHandler = completionHandler
        // Intitialize datastore based on replication direction
        let syncDatastore = (direction == SyncDirection.pull)
            ? SyncDataStore.initDatastore(name: to.rawValue)
            : SyncDataStore.initDatastore(name: from.rawValue)
        
        if (syncDatastore.datastoreManager != nil && syncDatastore.datastore != nil) {
            datastore = syncDatastore.datastore
            let factory = CDTReplicatorFactory(datastoreManager: syncDatastore.datastoreManager!)
            let replication = (direction == .push)
[...]

PR #408, opened 2017-11-22
Set error state when network offline

tomblench:
## What

When the network is offline as determined by `TDReachability`, set an appropriate error condition before stopping.

## How

Add symbolic constant `TDReplicatorErrorNetworkOffline`, and set an `NSError` object with this code before calling `[self postProgressChanged]` in `goOffline`.

## Testing

Added `testPullLotsOfOneRevDocumentsGoesOffline`.

## Issues

See #400 - currently when the network goes offline it looks like the replication has completely successfully.

### DIFF
--- a/CDTDatastore/touchdb/TDReplicator.m
+++ b/CDTDatastore/touchdb/TDReplicator.m
@@ -493,6 +493,11 @@ NSString* TDReplicatorStartedNotification = @"TDReplicatorStarted";
     if (!_online) return NO;
     CDTLogInfo(CDTREPLICATION_LOG_CONTEXT, @"%@: Going offline", self);
     _online = NO;
+    NSString *msg = @"Replication stopped because the reachability tracker determined the device was offline.";
+    NSDictionary *userInfo = @{NSLocalizedDescriptionKey: NSLocalizedString(msg, nil)};
+    self.error = [NSError errorWithDomain:TDInternalErrorDomain
+                                                      code:TDReplicatorErrorNetworkOffline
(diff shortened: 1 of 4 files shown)


CDTDatastore/touchdb/TDReplicator.m
```diff
@@ -507,6 +507,11 @@ - (BOOL)goOffline
     if (!_online) return NO;
     CDTLogInfo(CDTREPLICATION_LOG_CONTEXT, @"%@: Going offline", self);
     _online = NO;
+    NSString *msg = @"Reachability tracker has determined we have gone offline. Stopping replication.";
```
ricellis: I think this message would benefit from being a little less colloquial. Also I think given that it isn't logged immediately but IIUC arrives in the completion handler the tense of "stopping" may be incorrect.
Maybe:
>Replication stopped because the reachability tracker determined the device was offline.

WDYT?

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@
 - [FIXED] Threading issues in replicators.
 - [REMOVED] Removed deprecated class `CDTSavedHTTPAttachment` and method
   `createRevisionFromJson` on `CDTDocumentRevision`.
+- [FIXED] Added status code `TDReplicatorErrorNetworkOffline` -
+  replicators will go into an error state with this error code if the
+  network goes offline
```
ricellis: This probably needs an "instead of appearing to complete normally" so that people understand the change of the fix.
And a full stop.

PR #409, opened 2017-11-23
Remove support for SQLCipher encryption

tomblench:
## What

Remove support for SQLCipher encryption in build, CocoaPods, CI.

Note that the core functionality and associated tests have not been removed from the code base.

## Issues

See https://github.com/CocoaPods/CocoaPods/issues/7232 , 
https://github.com/ccgus/fmdb/issues/612 ,
https://discuss.zetetic.net/t/ios-11-xcode-issue-implicit-declaration-of-function-sqlite3-key-is-invalid-in-c99/2198/27 ,


ricellis:
@tomblench I think this can close now right, since we have a `pod` workaround?

tomblench:
yup

PR #411, opened 2018-01-25
Try push and pull replication up to 10 times

tomblench:
This allows long RA test runs with thousands of documents to be more tolerant of transient network issues.


CDTDatastoreReplicationAcceptanceTests/ReplicationAcceptance.m
```diff
@@ -1879,7 +1879,7 @@ - (void) testRemoteLastSequenceValueAfterPushReplication
     TD_Database *tdb = self.datastore.database;
 
     //see https://github.com/cloudant/CDTDatastore/issues/395
-    [NSThread sleepForTimeInterval:1.0f];
+    [NSThread sleepForTimeInterval:10.0f];
```
ricellis: Given that this is unrelated to the retries and didn't help with the other failure anyway I suggest we drop this commit from the PR and handle it under another issue.

Issue #417, opened 2018-04-02
Getting crash on calling listIndexes


navaneet:
Hi, we are using the latest version of CDTDatastore version 2.0.0 with swift. However when we call

`self.dataStore?.listIndexes()`

It results in the following crash

Execution was interrupted, reason: EXC_BAD_INSTRUCTION (code=EXC_I386_INVOP, subcode=0x0)

We have been able to call all other CDTDatastore methods needed for our purpose except this one. Please help. Also this exception has been persistent in the earlier versions of the library also.

We have tried  accessing the indexes the following way also, still no luck
 ```
     do {
            let i =  try CDTQIndexManager(using: self.dataStore!)
            i.listIndexes()
            } catch {}

```

**Comment by tomblench** [comment]
This works for us in objective-c although we haven't tried in swift yet. See [here](https://gist.github.com/tomblench/e80581182d72deb148b60488bb47de73) for an example gist which works.

Can you provide any more information like backtrace? `Execution was interrupted, reason: EXC_BAD_INSTRUCTION (code=EXC_I386_INVOP, subcode=0x0)` is extremely generic and doesn't help us track down the issue.

**Comment by tomblench** [comment]
OK I've reproduce on swift, digging into what's going wrong.

**Comment by tomblench** [comment]
@navaneet the problem occurs in objective-c/swift bridging, because the return type of `listIndexes` is incorrectly declared [here](https://github.com/cloudant/CDTDatastore/blob/master/CDTDatastore/query/CDTDatastore+Query.h#L49).

For reference it should be declared as
```objc
- (NSDictionary<NSString *, NSDictionary *> *)listIndexes;
```
and you can patch it locally yourself if you need an urgent fix.

Hopefully we should be able to get a release done soon which fixes this.

navaneet:
I will wait for the official patch, thanks @tomblench

tomblench:
@navaneet we have just release 2.0.1 which contains a fix for this bug.

navaneet:
Thanks let me check.

Issue #406, opened 2017-11-22
Update sample project


tomblench:
See https://github.com/cloudant/CDTDatastore/pull/405#issuecomment-346039612 - the sample should show users how to stop replications when the app is backgrounded.

It's also worth considering reverting to the previous "replication-policy" style sample code which existed before [this change](https://github.com/cloudant/CDTDatastore/commit/3a4429130ce08df72f7a1a94d30e0a6fd30ae7ae). There's a tension between a simple example and something which shows all the bells and whistles (as was discussed on #266), but the original sample would be useful to show to app developers who want code which automatically syncs on a schedule and adjusts its behaviour depending on whether it's in the foreground or background etc.

navaneet:
Please provide a swift 4 working sample as well, it's very difficult to find any swift example for CDTDatastore.

tomblench:
@navaneet sorry but a swift example isn't a priority right now

navaneet:
Ok I understand, please give us an Objective C working sample so that we can test against it and bridge the code as required.

tomblench:
@navaneet there's a sample project at https://github.com/cloudant/CDTDatastore/tree/master/Project plus extensive documentation in the https://github.com/cloudant/CDTDatastore/tree/master/doc, the README.MD file and appledocs comments. Plus the project is open source so you can read the source or follow it in a debugger.

Unfortunately we have other bugfixes and feature requests across this library and the other sync/client libraries that we provide and we have to prioritise resources. I'll look into the other bugs you raised.

navaneet:
@tomblench We have extensively gone through all the extensive documentation, problem is that the sample is not detailed enough and doesn't cover the required error handling scenarios. 

E.g. The sample shows the following

```
[self.datastore pullReplicationWithSource:[self replicatorURL]
                            completionHandler:^(NSError *_Nullable error) {
                              __strong CDTViewController *strongSelf = weakSelf;
                              [strongSelf.datastore
                                  pushReplicationWithTarget:[strongSelf replicatorURL]
                                          completionHandler:^(NSError *_Nullable error) {
                                            __strong CDTViewController *strongSelf = weakSelf;
                                            [strongSelf toggleCompletedShown:nil];
                                          }];
                            }];
```

Whereas in the documentation 

https://github.com/cloudant/CDTDatastore/blob/master/doc/replication.md 

it mentions only the usage of delegates instead of callbacks.

```
/**
* Called when the replicator changes state.
*/
[...]

PR #418, opened 2018-04-06
Fix listIndexes declaration

tomblench:
## What

Fix crash in `listIndexes` on swift.

## How

Fix listIndexes declaration.

This method returns a dictionary keyed by name, and each entry is itself a dictionary. The previous declaration was incorrect but only became apparent when invoking this on swift, which crashed whilst trying to bridge the dictionary from objective-c to swift.

## Testing

Reproduced original issue and tested fix with a custom test harness. Note that any automated testing is out of scope for this as there are currently no swift tests which run in jenkins/travis. This partly comes under #406.

## Issues

#417

### DIFF
--- a/CDTDatastoreTests/SwiftTests.swift
+++ b/CDTDatastoreTests/SwiftTests.swift
@@ -1,0 +1,37 @@
+//
+//  SwiftTests.swift
+//  CDTDatastore
+//
+//  Created by tomblench on 09/04/2018.
+//  Copyright © 2018 IBM Corporation. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+//  except in compliance with the License. You may obtain a copy of the License at
+//    http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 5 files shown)


CDTDatastoreTests/SwiftTests.swift
```diff
@@ -0,0 +1,37 @@
...
+    // test that the result of listIndexes() is correctly bridged from obj-c to swift
+    public func testListIndexes() {
+        do {
+            let store = try factory.datastoreNamed("my_ds")
+            let rev = CDTDocumentRevision()
+            rev.body = NSMutableDictionary(dictionary: ["hello":"world"])
+            try store.createDocument(from: rev)
+            store.ensureIndexed(["hello"])
+            let indexes = store.listIndexes()
+            print ("Indexes \(indexes)")
```
emlaver: Should we instead use some type of assertion here?

tomblench: Good point, the main reason for this test is that previously this code would have crashed, but we can do some asserts too.

PR #428, opened 2018-05-15
Fix retain cycle between CDTQIndexManager and CDTDatastore

tomblench:
## What

Fix retain cycle between `CDTQIndexManager` and `CDTDatastore`

## How

Make a weak reference to the associated object from the `CDTDatastore` to the `CDTQIndexManager`. Because it isn't reference counted, the `CDTQIndexManager` instance `dealloc` is correctly called (it didn't before because its reference count never reached 0), at which point the reference is cleared.

## Testing

See added test class `CDTQLifecycleTests`


ricellis:
@tomblench before you merge this also needs a CHANGELOG entry

Issue #427, opened 2018-05-11
Getting exception with large number of document insertions


navaneet:
Hi,

I am getting the following exception when I try to insert a 1000 documents into the datastore

```
doc 964 inserted
2018-05-11 13:45:28.844689+0530 Sample[12770:198297] Unknown error calling sqlite3_step (14: unable to open database file) eu
2018-05-11 13:45:28.844861+0530 Sample[12770:198297] DB Query: pragma user_version = 2
```
The error message starts appearing towards the later half of the insertions of a 1000 documents. It doesn't happen if the number of documents is small like 100 or so. I am using a macbook pro 2014 model to run the tests with the latest xcode.

This I believe is caused due to the datastoremanager not properly maintaining the datastore in it's internal map which I have been told should return the same instance of the datastore if it's for the same db which it currently doesn't. The following sample code  in swift can demonstrate the issue:

 ```
            let dbName = "flight"
            let DATABASE_PATH = "databases"
            // Check if datastore manager doen't exist
            let path = FileManager.default.documentDirectoryWith(lastPath: DATABASE_PATH)
            var manager: CDTDatastoreManager?
            do {
                manager = try CDTDatastoreManager(directory: path)
            } catch {
                print("\(error)")
            }
            
            //delete any previous datastore
            do {
                try manager?.deleteDatastoreNamed(dbName)
            } catch {
                print("\(error)")
            }
            
            var datastore: CDTDatastore?
            
            //insert documents
            for number in 1...1000 {
                let rev: CDTDocumentRevision = CDTDocumentRevision(docId: String(number))
                rev.body = ["type":"temp", "number": number, "name": "abc\(number)"]
                do {
                    //datastore created for every run of the loop
                    datastore = try manager?.datastoreNamed(dbName)
                    try datastore?.createDocument(from: rev)
                } catch {
                    print("Error while creating doc \(error)")
                }
                datastore?.ensureIndexed(["type", "number", "name"], withName: dbName)
                print("doc \(number) inserted")
            }
            
            //update indexes
            print("updating all indexes")
            datastore?.updateAllIndexes()
```

Here if the datastore is created outside the loop it doesn't show that error message which revalidates my point that the datastoremanager for some reason is unable to properly handle the insertions with large quantity of document insertions when the datastores are recreated at runtime for each new document insertion.

```
datastore = try manager?.datastoreNamed(dbName)
for number in 1...1000 {
```

ricellis:
[...]

PR #429, opened 2018-05-16
Add test to ensure CDTDatastore filehandles aren't leaked

tomblench:
Note that the aim of this PR has diverged from the original purpose, hence the branch name not reflecting the PR title.

## What

Add a test to ensure that we don't excessively leak filehandles, which is an indication that the `CDTDatastore` and `CDTQIndexManager` objects are not being dealloc'd correctly.

## How

Repeatedly add documents and call `ensureIndexed`. In each loop iteration use `proc_pidinfo` to check number of open filehandles, and assert on this.

## Testing

See added test `testDatastoreClosesFilehandles`.

_Note_ 

This test is currently expected to fail until #428 merges.

_Note_

 This test will fail without the `@autoreleasepool` present. See https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/MemoryMgmt/Articles/mmAutoreleasePools.html for further details.

In particular: "Cocoa always expects code to be executed within an autorelease pool block, otherwise autoreleased objects do not get released and your application leaks memory. (If you send an autorelease message outside of an autorelease pool block, Cocoa logs a suitable error message.) The AppKit and UIKit frameworks process each event-loop iteration (such as a mouse down event or a tap) within an autorelease pool block". 

Because we "are writing a program that is not based on a UI framework, such as a command-line tool", without the `@autoreleasepool`, the `dealloc` methods would only be called all at once at the end of the test (which would also leak file handles).

## Issues

#427

### DIFF
--- a/CDTDatastoreTests/DatastoreManagerTests.m
+++ b/CDTDatastoreTests/DatastoreManagerTests.m
@@ -20,7 +20,8 @@
 #import "TDInternal.h"
 #import "FMDatabase.h"
 #import "FMDatabaseQueue.h"
-
+#import <libproc.h>
+#import <sys/proc_info.h>
 @interface DatastoreManagerTests : CloudantSyncTests
 
 @end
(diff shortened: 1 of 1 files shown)


CDTDatastoreTests/DatastoreManagerTests.m
```diff
@@ -96,6 +97,40 @@ -(void) testSchema6ToSchema100Upgrade {
     }
 }
 
+- (void) testDatastoreClosesFilehandles {
+    // repeatedly obtain the same datastore in a loop, adding documents to it and calling ensureIndexed
+    // on each iteration, check that we have not excessively leaked filehandles, which would  indicate that the datastore and indexmanager are not being dealloc'd correctly
```
ricellis: extra space between `would` and `indicate`

CDTDatastoreTests/DatastoreManagerTests.m
```diff
@@ -96,6 +97,40 @@ -(void) testSchema6ToSchema100Upgrade {
...
+            int bufferSize = proc_pidinfo(pid, PROC_PIDLISTFDS, 0, 0, 0);
+            // buffer size is bytes, we want number of processes
+            struct proc_fdinfo *fdInfo = (struct proc_fdinfo *)malloc(bufferSize);
+            int fdCount = bufferSize / PROC_PIDLISTFD_SIZE;
+            // now call with the buffer to get the actual number of processes (may be lower than buffer size)
+            bufferSize = proc_pidinfo(pid, PROC_PIDLISTFDS, 0, fdInfo, fdCount);
+            // buffer size is bytes, we want number of processes
+            fdCount = bufferSize / PROC_PIDLISTFD_SIZE;
+            // as observed in testing, this should never go above 8, but we'll set a conservative limit of 100 to allow some breathing room
+            XCTAssertTrue(fdCount < 100);
```
ricellis: Can we break the loop if this fails? Currently the test result repeats the failure many times, but I think one is enough for us to know it has gone wrong.

CDTDatastoreTests/DatastoreManagerTests.m
```diff
@@ -20,7 +20,8 @@
 #import "TDInternal.h"
 #import "FMDatabase.h"
 #import "FMDatabaseQueue.h"
-
```
ricellis: Needs a copyright update ^^

PR #430, opened 2018-05-16
Synchronise access to _databases dictionary

tomblench:
## What

Synchronise access to _databases dictionary, to handle cases where multiple threads interact with the same `CDTDatastoreManager`.

## How

Add `@synchronized` blocks to stop potential race conditions.

## Testing

TODO


tomblench:
@ricellis I synchronised `close` in dd2b6ea because I think there's a race condition where you could get an "orphaned" database. 

`allOpenDatabases` is a read-only view into the map and it doesn't appear to be used except in test code so I'm not worried about that.

PR #436, opened 2018-06-25
CDTQIndexManager has weak reference to CDTDatastore

### LINKED ISSUE #435: Crash in release mode on Xcode 9.4.1 with CDTDatastore version 2.0.3
## Bug Description
`datastore?.find([:])` crashes consistently in release mode but works in debug mode.

### 1. Steps to reproduce and the issue
Run the sample app provided in release mode, it will crash on the following lines of code

```
                //print results
                resultSet = (datastore?.find([:]))!
```
Release mode can be enabled by going to edit scheme and selecting release in build configuration

edit scheme

release mode

### 2. What you expected to happen
Not to crash!

### 3. What actually happened
It crashed!

crash

## Environment details
- Version affected by this issue.
    CDTDatastore 2.0.3
- iOS version tested on
    iOS  11.4
- Swift Version
    4.1

## Sample attached below

[DataStore Sample.zip](https://github.com/cloudant/CDTDatastore/files/2131957/DataStore.Sample.zip)

tomblench:
## Checklist

- [x] Tick to sign-off your agreement to the [Developer Certificate of Origin (DCO) 1.1](../blob/master/DCO1.1.txt)
- [x] Added tests for code changes _or_ test/build only changes
- [x] Updated the change log file (`CHANGES.md`|`CHANGELOG.md`) _or_ test/build only changes
- [x] Completed the PR template below:

## Description

Fix lifecycle issues regarding proper opening and closing of resources across `CDTDatastore`, `CDTQIndexManager`, `CDTDatastoreManager`.

Note that the primary motivator here is that the associated object relationship between `CDTDatastore` and `CDTQIndexManager` is currently incorrect when facading through index/query methods, and that it is currently possible to obtain multiple `CDTQIndexManager` instances for the same underlying SQLite database, which leads to incorrect behaviour.

For more details see "Approach" which gives an overview of the design issues.

Fixes #435

#### `CDTDatastore` -> `CDTQIndexManager`

1-to-1 relationship, strong reference via `objc_{get,set}AssociatedObject` in
`CDTDatastore+Query.m`.

Each index manager is constructed on demand.

#### `CDTQIndexManager` -> `CDTDatastore`

1-to-1 relationship, weak reference via property `datastore`, therefore no retain cycles. Because
reference to `CDTQIndexManager` can't 'leak' (all methods facade through `CDTDatastore+Query`), this
reference will always be valid.

Note that if `CDTQIndexManager` is constructed using `+managerUsingDatastore` or
`-initUsingDatastore`, the above doesn't hold and it is possible for the reference to `datastore` to
be invalid. Even worse, a user could then use a mixture of API calls through the constructed index
manager and facaded through the datastore. Therefore this usage is discouraged (all of our current
documentation shows index/query usage facaded through datastore).

For the above reason, add a comment in `CDTQIndexManager` to clarify that it's not public API.

#### `CDTDatastoreManager` -> `CDTDatastore`

1-to-many relationship, strong references via the `openDatastores` property, which is a map whose
keys are the datastore name, and values the datastore itself.

Note that the only way of obtaining a new or existing datastore instance is through the manager.

This is a change to existing functionality and guarantees users will obtain the same reference when
multiple calls to `-datastoreNamed` are made. This is essential for the `CDTDatastore` ->
`CDTQIndexManager` to work properly. As of now, this is not the case, and multiple calls to
`-datastoreNamed` will result in a new and different `CDTDatastore` to be constructed each time
(although they all point to the same underlying `TD_Database`). This causes errors as the different
index managers will all try to simultaneously open the same underlying sqlite database.

Note that a user could still have a 'dangling' reference to a `CDTDatastore`, that is they can
assign the pointer of a previously constructed datastore, then `dealloc` its manager. Now they have
a datastore they can't delete (they need the manager for this), and worse, they can now obtain a
different pointer for the same underlying datastore by constructing a new manager and calling
`-datastoreNamed`. I think we have to be pragmatic and decide we can't protect users from this API
abuse.

#### `CDTDatastore` -> `CDTDatastoreManager`

1-to-1 relationship. This is currently held as a strong reference via the property `_manager`, but
will cause retain cycles when combined with the relationship described in the above para. Is not
currently used so can be downgraded to a weak reference or removed.

## Schema & API Changes

Add new method `closeDatastoreNamed` on `CDTDatastoreManager` to allow users with many datastores open to avoid resource (file handle) starvation.

## Security and Privacy

No change.

## Testing

See added test class `CDTLifecycleTests`. Test approach as follows:

### New instances

- Calls to `CDTDatastore+Query.m` methods should always obtain the same `CDTQIndexManager` instance
  for a given `CDTDatastore` instance.
- Calls to `CDTDatastoreManager` `-datastoreNamed` should always obtain the same `CDTDatastore`
  instance for a given datastore name.

### Retain count reaching zero

- When `dealloc` is called on `CDTDatastore`, `dealloc` should be called on the `CDTQIndexManager`
  held as an associated object.
- When `dealloc` is called on `CDTDatastoreManager`, `dealloc` should be called on all
  `CDTDatastores` in `openDatastores`.
- Should we have tests for `TD_Database` retain count?

### Explicit close

- Explicit close supported and tested via new method call on  `CDTDatastoreManager`

### Explicit delete

- When deleting extensions, `dealloc` should be called on `CDTQIndexManager` to allow it to close
  its underlying database for the extensions directory is deleted. (In `deleteDatastoreNamed` call
  `[_openDatastores removeObjectForKey:name]` first). Note that not doing this is the cause of the
  following message, as reported by customer:

### Other test changes

Updated some existing tests which now need an explicit close before re-opening database to reproduce expected behaviour.

Tests run via `Rake` now run in `release` mode, as the original bug report shows certain failure modes are only exhibited when run in `release` (as opposed to `debug`) mode.

Removed some `#ifdef(DEBUG)` guards to allow test code to compile in `release` mode.

## Monitoring and Logging

TODO - some debug `NSLog`s have been left in, these need to be evaluated and turned into "proper" log statements if necessary.

### DIFF
--- a/CDTDatastore/CDTDatastore.m
+++ b/CDTDatastore/CDTDatastore.m
@@ -4,6 +4,7 @@
 //
 //  Created by Michael Rhodes on 02/07/2013.
 //  Copyright (c) 2013 Cloudant. All rights reserved.
+//  Copyright © 2018 IBM Corporation. All rights reserved.
 //
 //  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
 //  except in compliance with the License. You may obtain a copy of the License at
@@ -42,7 +43,7 @@ NSString *const CDTDatastoreChangeNotification = @"CDTDatastoreChangeNotificatio
 
 @property (nonatomic, strong, readonly) id<CDTEncryptionKeyProvider> keyProvider;
 
-@property (readonly) CDTDatastoreManager *manager;
+@property (readonly, weak) CDTDatastoreManager *manager;
 - (void)TDdbChanged:(NSNotification *)n;
 - (BOOL)validateBodyDictionary:(NSDictionary *)body error:(NSError *__autoreleasing *)error;
 
@@ -93,7 +94,11 @@ NSString *const CDTDatastoreChangeNotification = @"CDTDatastoreChangeNotificatio
     return self;
 }
 
-- (void)dealloc { [[NSNotificationCenter defaultCenter] removeObserver:self]; }
+- (void)dealloc {
+    _database = nil;
+    NSLog(@"dealloc CDTDatastore");
+    [[NSNotificationCenter defaultCenter] removeObserver:self];
+}
 
 #pragma mark Properties
 
--- a/CDTDatastore/CDTDatastore.m
+++ b/CDTDatastore/CDTDatastore.m
@@ -96,7 +96,7 @@ NSString *const CDTDatastoreChangeNotification = @"CDTDatastoreChangeNotificatio
 
 - (void)dealloc {
     _database = nil;
-    NSLog(@"dealloc CDTDatastore");
+    CDTLogDebug(CDTDATASTORE_LOG_CONTEXT, @"-dealloc CDTDatastore %@", self);
     [[NSNotificationCenter defaultCenter] removeObserver:self];
 }
 
--- a/CDTDatastore/CDTDatastoreManager.m
+++ b/CDTDatastore/CDTDatastoreManager.m
@@ -138,7 +138,7 @@ NSString *const CDTExtensionsDirName = @"_extensions";
             // delete any cloudant extensions
             // remove the open datastore: this ensures that any associated index manager has -dealloc
             // called on it _before_ we attempt to delete its underlying database
-            NSLog(@"calling close from delete %@", name);
+            CDTLogDebug(CDTDATASTORE_LOG_CONTEXT, @"calling close from delete %@", name);
             [_openDatastores removeObjectForKey:name];
             NSString *dbPath = [self.manager pathForName:name];;
             NSString *extPath = [dbPath stringByDeletingLastPathComponent];
--- a/CDTDatastore/CDTDatastoreManager.m
+++ b/CDTDatastore/CDTDatastoreManager.m
@@ -113,10 +113,6 @@ NSString *const CDTExtensionsDirName = @"_extensions";
 
 - (void)dealloc {
     CDTLogDebug(CDTDATASTORE_LOG_CONTEXT, @"-dealloc CDTDatastoreManager %@", self);
-    /*
-    for (CDTDatastore *ds in _openDatastores) {
-        [[ds database] close];
-    }*/
     [_openDatastores removeAllObjects];
     _manager = nil;
 }
--- a/CDTDatastore/CDTDatastoreManager.m
+++ b/CDTDatastore/CDTDatastoreManager.m
@@ -4,6 +4,7 @@
 //
 //  Created by Michael Rhodes on 04/07/2013.
 //  Copyright (c) 2013 Cloudant. All rights reserved.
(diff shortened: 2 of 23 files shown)


CDTDatastore/CDTDatastoreManager.m
```diff
@@ -26,6 +27,8 @@
 
 @interface CDTDatastoreManager ()
 
+@property NSMutableDictionary<NSString*, CDTDatastore*> *openDatastores;
```
smithsz: could this be made nonatomic given all reads/writes are inside a synchronized block?

CDTDatastore/CDTDatastoreManager.m
```diff
@@ -55,72 +59,110 @@ - (CDTDatastore *)datastoreNamed:(NSString *)name
...
+        }
+        [[ds database] close];
+        [_openDatastores removeObjectForKey:name];
     }
+}
 
-    return datastore;
+- (void)dealloc {
+    NSLog(@"dealloc CDTDatastoreManager");
+    /*
```
smithsz: forget to delete this? I assume the dealloc is called on databases as they fall out of scope.

CDTDatastoreTests/CDTLifecycleTests.m
```diff
@@ -0,0 +1,278 @@
...
+    NSError *error;
+    CDTDatastoreManager *factory = [[CDTDatastoreManager alloc] initWithDirectory:factoryPath error:&error];
+    // assign to ds1 inside an autoreleasepool so our pointer doesn't hang on to it
+    @autoreleasepool {
+        CDTDatastore *ds1 = [factory datastoreNamed:@"test" error:&error];
+        XCTAssertNil(error);
+        [ds1 updateAllIndexes];
+    }
+    // currently there is no way to assert on this - watch the log and check that the following message is not output:
+    // "BUG IN CLIENT OF libsqlite3.dylib: database integrity compromised by API violation: vnode unlinked while in use"
```
ricellis: Can we read in the log file and assert the message is not found?

CDTDatastore/CDTDatastore.m
```diff
@@ -4,6 +4,7 @@
 //
 //  Created by Michael Rhodes on 02/07/2013.
 //  Copyright (c) 2013 Cloudant. All rights reserved.
+//  Copyright © 2018 IBM Corporation. All rights reserved.
```
emlaver: Should the latest copyright date be on a separate line or should it be similar to the other libraries e.g. `Copyright (C) 2013, 2018 IBM Corp. All rights reserved.`

tomblench:
@ricellis suggested log-grepping strategy in e8472c404fc337ad1a3c201f803c74373a1d19f4 - `freopen` `stderr` to a temp file, then `mmap` the file to check contents.

ricellis:
@tomblench looks good, as discussed I think it is worth updating the sample app with the close call if it is relevant there.
Other query tests likely need some updates too, or at least an issue to track those updates.

Issue #454, opened 2019-03-11
Intermittent 404 received on pull replication 


paulJordaan:
Please [read these guidelines](http://ibm.biz/cdt-issue-guide) before opening an issue.

## Bug Description

Intermittently during a pull replication a device would get stuck in a state where it is not able to complete replication to a database because it is receiving a 404 response for a certain sequence. After a lot of digging I came across this [issue](https://github.com/couchbase/couchbase-lite-ios/issues/377) on the [couchbase-lite-ios](https://github.com/couchbase/couchbase-lite-ios) repo which described a similar issue.  

### 1. Steps to reproduce and the simplest code sample possible to demonstrate the issue
This issue is difficult to reproduce as it only occurs intermittently on some devices. I have not been able to recreate the issue in a reliable way.

### 2. What you expected to happen
Expect a replication to complete.

### 3. What actually happened
A replication can get stuck in loop where it continuously errors with "404 Not Found" for a specific sequence when setting `CDTChangeLogLevel(CDTREPLICATION_LOG_CONTEXT, DDLogLevelWarning);`

When searching for the requested sequence id via the web UI the sequence can also not be found.

According to the issue from the above mentioned couchbase repo this can happen "as a race condition if a revision on the server gets replaced by a newer one and then compacted away, before the client receives a notification of the newer revision."

By manually applying the changes in this [pull request](https://github.com/couchbase/couchbase-lite-ios/commit/87f32616d405d00da798a45450c15fa97aa3485c) I was able to get one of my devices experiencing the issue out of the loop. Would it be possible for one of the maintainers to review the changes from the mentioned pull request to see if it might be relevant to this library. I'm not familiar enough with the code and the replication logic to be sure that the mentioned changes is not breaking anything else

## Environment details
- Version(s) that are affected by this issue.
    > 2.1.1
- iOS/macOS version
    > iOS 10.13.1

ricellis:
Interestingly the CouchDB replication protocol doesn't explicitly specify what to do in the case of receiving a `404` for a revision that has been calculated as needing to be fetched.

The [Replicate Changes](http://docs.couchdb.org/en/stable/replication/protocol.html#replicate-changes) flow shows only a `200` response for `Fetch Next Changed Document`.

From [protocol robustness](http://docs.couchdb.org/en/stable/replication/protocol.html#protocol-robustness):
[...]

PR #458, opened 2019-07-22
[WIP] Added handling for server side missing revs

ricellis:
## Checklist

- [x] Tick to sign-off your agreement to the [Developer Certificate of Origin (DCO) 1.1](../blob/master/DCO1.1.txt)
- [ ] Added tests for code changes _or_ test/build only changes
- [ ] Updated the change log file (`CHANGES.md`|`CHANGELOG.md`) _or_ test/build only changes
- [x] Completed the PR template below:

## Description

If server side compaction removes revs between the iOS client fetching the changes feed and fetching the changes there are potential `404 not found` conditions.

See #454 

## Approach

Since the `404` alone does not provide enough information to determine if it is transient (i.e. the document is yet to be consistent across all nodes) or permanent (i.e. the rev has been permanently removed) then we must assume the transient case in the first instance.

To deal with this any rev that gets a 404 from the bulk cases will be requeued for individual fetching. If the document is subsequently not found when an individual fetch attempt is made then it is considered a permanent failure.

This is approximately inline with the behaviour of the [CouchDB replicator](https://github.com/apache/couchdb/blob/103a0624f309ea0d796176a55eb5faea68f26047/src/couch_replicator/src/couch_replicator_worker.erl#L334-L341) which performs a single retry after a configurable delay.

This change will not solve the problem for cases where only an individual fetch attempt is made, but all Couch 2.x support the `_bulk_get` endpoint and Couch 1.x is EOL so this fix should be suitable.

## Schema & API Changes

- No API changes

## Security and Privacy

- No change

## Testing

TODO

## Monitoring and Logging

- Added new log line to warn of permanent failures fetching a revision.

### DIFF
--- a/CDTDatastore/touchdb/TDPuller.m
+++ b/CDTDatastore/touchdb/TDPuller.m
@@ -452,9 +452,21 @@ static NSString* joinQuotedEscaped(NSArray* strings);
                                                __strong TDPuller* strongSelf = weakSelf;
                                                // OK, now we've got the response revision:
                                                if (error) {
-                                                   strongSelf.error = error;
-                                                   [strongSelf revisionFailed];
-                                                   strongSelf.changesProcessed++;
+                                                   if ([error.domain isEqualToString: TDHTTPErrorDomain] &&
+                                                       error.code == kTDStatusNotFound) {
+                                                       // The revision could not be found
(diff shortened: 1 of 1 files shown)


ricellis:
This library is now end-of-life.

